PR #71, opened 2020-09-25
feat: set up automatic labelling of pull requests

wincent:
As per:

https://github.com/actions/labeler

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,36 @@
+---
+# Configures labels to be automatically applied to PRs based on paths.
+#
+# See: https://github.com/actions/labeler
+
+changelog-generator:
+  - projects/npm-tools/packages/changelog-generator/*
+  - projects/npm-tools/packages/changelog-generator/**/*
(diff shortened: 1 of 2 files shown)


wincent:
I don't know of any way to test this short of actually merging it, so I figure that we may a well just push it and see, #amirite?

wincent:
See this in action over at: https://github.com/liferay/liferay-frontend-projects/pull/72

Screenshot 2020-09-25 -105002-ULMtKxla@2x

jbalsas:
Pretty cool!!

We could also explore additional labelling, like `team-*` so we can prioritize, or `bug`, `story` based on the LPS status... (maybe this is getting out of hand, though :D )

wincent:
Yep, if you look through [the docs](https://github.com/actions/labeler) you can do somewhat sophisticated glob-based labelling - more than what I am doing here. It doesn't quite go as far as what you are saying, but it might in the future, seeing as the labeler is an "offical" GitHub action.

PR #72, opened 2020-09-25
docs: move CI badge up to top-level README

wincent:
As I've noted in other commits, for now we are just going to run CI over everything in the monorepo. Once that gets too big/slow we will look at making the runs "smarter" (eg. only run on projects which have changes in them).

But for now, given that the run is global, the badge belongs somewhere global too.


wincent:
This works:

Screenshot 2020-09-25 -105049-RyZGJbEl@2x

and it is trivial, so I am just going to merge it as soon as I see CI green.

PR #73, opened 2020-09-25
docs: add a table with links to per-project and per-package issue listings

wincent:
With more projects in the monorepo we need good tools for dealing with the increased load on the issue tracker, so that people can focus on what they need to on any given moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,27 @@ This is an experimental exploration of the ideas proposed in [liferay-frontend-g
 
 ## Projects
 
--   [@liferay/eslint-config](./projects/eslint-config)
--   [Liferay NPM Tools](./projects/npm-tools):
-    -   [@liferay/changelog-generator](./projects/npm-tools/packages/changelog-generator)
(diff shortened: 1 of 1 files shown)


wincent:
Whoops: link text is wrong — going to force push to update.

jbalsas:
> Whoops: link text is wrong — going to force push to update.
 
dissapointment

jbalsas:
Would be nice to also have issue templates that can target the different projects?

Not sure if we can have hierarchies, so we can have:

- Project A
    - Bug
    - Feature Request
- Project B
    - Conversation
    - Question

Main benefit of at least the root level would be to automatically provide project labels which I guess are out of the scope of our amazin labeller action ;)

wincent:
Yes, I think issue templates are a good idea. I was going to look at those when I bring in a project that actually already has them, but we can do it for these two projects that we already have (I'll make that a separate PR).

wincent:
Initial sketch of templates here: https://github.com/liferay/liferay-frontend-projects/pull/76

wincent:
Gonna merge this. Too lazy to find a gif.

lazy

PR #74, opened 2020-09-25
feat: format YAML files with Prettier

wincent:
Seeing as [Prettier has supported YAML for a long time now](https://prettier.io/blog/2018/07/29/1.14.0.html), we may as well use it.


wincent:
Merging this one too. Can't be bothered finding a gif to justify.

irresponsible

PR #78, opened 2020-09-25
docs: make additional notes about how to keep repo "meta" stuff up-to-date

wincent:
Seeing as we have added some new config files lately, let's add a reminder here about keeping them up-to-date.


jbalsas:
sealofapproval

PR #76, opened 2020-09-25
feat: set up issues templates

wincent:
As discussed [here](https://github.com/liferay/liferay-frontend-projects/pull/73#issuecomment-698824867).

Note: this won't scale super well once this project has many more packages in it (which will happen as soon as we pull in the js-toolkit), so we may have to revisit this in the future.

We can't have a hierarchical issue template (ie. projects, with issue types as sub-templates), so I have just stuck the type in here as a generic checklist. We can flesh these out more later. Maybe something super generic like:

---

**Desired behavior:**

**Current behavior:**

**Other information (environment, versions etc):**

---

That might cover feature requests, bug reports and questions well enough, even though it is a bit imprecise.

But note that even without hierarchical templates, doing what I've done here (one template per package) is going to get out of hand soon. At that point we might have to distill this down a bit more.

PR #79, opened 2020-09-25
fix: don't use a mailto URL in the issue template configuration

wincent:
As described [here (see screenshots)](https://github.com/liferay/liferay-frontend-projects/pull/76#issuecomment-698896555), it doesn't seem to be allowed.

Example project using a non-mailto URL:

- https://github.com/node-fetch/node-fetch/issues/new/choose
- https://github.com/node-fetch/node-fetch/blob/master/.github/ISSUE_TEMPLATE/config.yml

Note it is also using `blank_issues_enabled`, but it doesn't seem to make any difference. Clay for example, has the same behavior even without the property. Perhaps it defaults to `true`.

- https://github.com/liferay/clay/issues/new/choose
- https://github.com/liferay/clay/tree/master/.github/ISSUE_TEMPLATE

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -2,5 +2,5 @@
 blank_issues_enabled: true
 contact_links:
     - name: 👮 Report a security issue
-      url: 'mailto:pt-frontend-infrastructure@liferay.com'
+      url: https://github.com/node-fetch/node-fetch/issues/new/choose
       about: Report security problems privately here


wincent:
Once again, sadly, the only way to know for sure that this works is to merge it (and I am too lazy to set up a separate repo to test it in isolation).

slack

wincent:
Sigh. So the next undocumented(?) wonderful thing is that issue template filenames can't have hyphens in them...

Screenshot 2020-09-25 -142725-PtFsxlKH@2x

PR #80, opened 2020-09-25
fix: remove hyphens from issue template names

wincent:
Because [apparently](https://github.com/liferay/liferay-frontend-projects/pull/79#issuecomment-698900089), GitHub won't use them if they have hyphens...


wincent:
Maybe can get this right on the third try...

cowboy

wincent:
Hm, it still doesn't work, and I don't know why. Maybe it doesn't like something in the metadata, like the leading `@` etc.

- Working example from Clay, which uses Unicode directly (no quoting): https://raw.githubusercontent.com/liferay/clay/master/.github/ISSUE_TEMPLATE/Bug.md
- Working example from node-fetch, which uses a Unicode escape sequence in a quoted string: https://raw.githubusercontent.com/node-fetch/node-fetch/master/.github/ISSUE_TEMPLATE/support-or-usage.md
- Our broken config: https://raw.githubusercontent.com/liferay/liferay-frontend-projects/88c323c6210332eeaa3a587964b6d3bc40cbd8b1/.github/ISSUE_TEMPLATE/ChangelogGenerator.md

My best guess is that the `@` is causing the frontmatter parser to choke, as seen in these renderings:

Screenshot 2020-09-25 -144233-P3yPy1zP@2x
Screenshot 2020-09-25 -144158-jWY71djW@2x
Screenshot 2020-09-25 -144213-wC6NS4MW@2x

Yep, looks like that's it — look at what happens when I quote that string:

Screenshot 2020-09-25 -144343-bfr3BrRJ@2x

PR #82, opened 2020-09-25
docs: add direct links for per-project/package issue creation

wincent:
Does what it says on the tin.


wincent:
Just docs, so seems safe enough to go in.

safe

PR #83, opened 2020-09-25
feat: teach the changelog-generator to work with new monorepo structure

wincent:
Previously, the changelog-generator would work with single-project repos:

    CHANGELOG.md
    package.json
    ...

or monorepos with this fixed structure:

    package.json
    packages/
        a/
            package.json
            ...
        b/
            package.json
            ...
        ...

When working with monorepos, there would typically be a single `.yarnrc` at the top and a shared changelog:

    .yarnrc
    CHANGELOG.md
    package.json
    packages/
        a/
            package.json
            ...
        b/
            package.json
            ...
        ...

or per-package changelogs with per-package `.yarnrc` files:

    package.json
    packages/
        a/
            .yarnrc
            CHANGELOG.md
            package.json
            ...
        b/
            .yarnrc
            CHANGELOG.md
            package.json
            ...
        ...

We also supported working in repos without `.yarnrc` files, but that isn't the common case for us, especially in repos that use the `js-publish` package to do their releases.

With the new monorepo, we have a more complicated structure, with things being divided into projects, and potentially nested packages below that:

    package.json
    projects/
        project-a/
            .yarnrc
            CHANGELOG.md
            package.json
            ...
        project-b/
            package.json
            packages/
                a/
                    .yarnrc
                    CHANGELOG.md
                    ...
                b/
                    .yarnrc
                    CHANGELOG.md
                    ...
                ...
        ...

So, let's loosen our assumptions about project structure and stop hardcoding paths (eg. `packages/*`). The point of those requirements was to catch when people accidentally ran the changelog generator from the wrong directory; eg.

- from "packages/"
- or from "packages/some-project/some-subdirectory/"
- instead of "packages/some-project/"

(Running from the right directory is important because we rely on that in order to correctly query Git for the changes that affect the package and which should be included in the CHANGELOG.)

Now, we achieve the same ends by just checking to make sure that we get run from a directory with a "package.json" in it. It's true that you could `cd` into some `node_modules` subdirectory or a test fixture directory with a "package.json" in it, but if we ignore that kind of edge case we can make the code a lot simpler, so that's what we do.

**Test plan:** Make a fake merge in this repo and run the generator and see it work; then `cd` too deep into a subdirectory and see an error:

    # From this PR's branch, go back to master.
    git checkout master

    # Make a merge that looks like a GitHub one.
    git merge --no-ff wincent/changelog-directories -m 'Merge pull request #9000 from blah\n\nfix: dat shit'

    # Run the generator.
    node bin/liferay-changelog-generator.js --version=100.0.0

    # Inspect the changes to the changelog.
    git diff

    # Undo the changes.
    git checkout .

    # Go too deep.
    cd src

    # Run the generator again, expecting an error.
    node ../bin/liferay-changelog-generator.js --version=100.0.0

    # Put everything back the way it was.
    git reset --hard upstream/master
    git checkout -
    cd ..

### DIFF
--- a/projects/npm-tools/packages/changelog-generator/src/index.js
+++ b/projects/npm-tools/packages/changelog-generator/src/index.js
@@ -628,14 +628,14 @@ async function generate({date, from, remote, to, version}) {
 /**
  * Returns a tag prefix that can be used to construct or find a version tag.
  *
- * -   If we're being run from a "packages/$PACKAGE" subdirectory in a
- *     monorepo and a ".yarnrc" file exists, we return its "version-tag-prefix".
+ * -   If we're being run from a subdirectory in a monorepo and a ".yarnrc" file
+ *     exists, we return its "version-tag-prefix".
  *
  *     For example, given a prefix of "my-package/v", then we can find matching
  *     tags using `git describe --match='my-package/v*'.
  *
  *     Likewise, if we're being run from the repo root and a ".yarnrc" file
- *     exists.
+ *     exists there.
  *
  *     For example, given a prefix of "v", then we can find matching tags using
  *     `git describe --match='v*'`.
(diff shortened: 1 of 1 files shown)


wincent:
Also tested this by producing the updates in https://github.com/liferay/liferay-frontend-projects/pull/84, so I think this one is probably... safe...

safe-2

PR #88, opened 2020-09-29
feat!(js-publish): don't bother updating stable branch any more

wincent:
After using this "auto-update-`stable`" behavior for a year now, we've never once needed to go back to `stable` and cut a release from it.

So, let's just jettison it. Now that we've moved to the monorepo it seems like a good opportunity to avoid adding the clutter.


wincent:
It's green and @izaera approves (via Slack) of getting rid of `stable`, so I'm going to merge this.

PR #90, opened 2020-09-29
fix(scripts): remove stale reference to old bundler preset

wincent:
But note that, [as described here](https://github.com/liferay/liferay-js-toolkit/pull/654), we are going to need a fix to the bundler because it doesn't currently work with named-scope presets:


wincent:
This is a trivial, tested config change, and I need to cut a release, so going to merge it.

PR #94, opened 2020-09-30
docs(changelog-generator): update project links

wincent:
The old repos have been archived (marked read-only) as of today, so we may as well update these links to point at the new locations.


wincent:
Just link changes, and I double-checked them, so going to merge this.

#safe

safe

PR #97, opened 2020-09-30
feat: add support for formatting and linting TypeScript

wincent:
Note that the goal here is to format and lint TypeScript in liferay-portal, but we're "dog-fooding" in this monorepo already in this PR, even though there isn't any TypeScript in here (yet... when we bring across the js-toolkit packages TypeScript will start coming in here too).

Fun fact, we're so forward-thinking 😂 that we already have ".ts" and ".tsx" in [our list of lintable globs](https://github.com/liferay/liferay-frontend-projects/blob/ab17631602f56b768dd1bebf7d8a7a096ac8dbbc/projects/npm-tools/packages/npm-scripts/src/scripts/lint.js#L32) and it's been that way [for 16 months now](https://github.com/liferay/liferay-frontend-projects/commit/9388c34309ac260e3a5780afa78f44ab6570c2a8) — it's just that we never configured our globs in liferay-portal to actually target TS files, because there weren't any yet (and that's also why we never added the necessary deps: `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin`).

Note how we can lint JS just fine with the TS parser.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/96

PR #98, opened 2020-10-01
feat: add script to install packages locally in a DXP checkout

wincent:
This is a crude script that will almost certainly have to be evolved as the shape and structure of the monorepo evolves, but it works for now with the set of packages that we currently have.

It addresses a bunch of "gotchas" that come with trying to test out packages with local changes in a liferay-portal checkout. Depending on which packages you want to test and the dependency graph between them, you may find that `yarn link` (or even `yarn add`) doesn't do what you want. Even if you manually copy files (`cp`), things still might not work because Gradle may decide to re-run a task that ends up blowing away any changes you make locally under `node_modules`.

At least for the simple case, this script was good enough to allow me to test [#97](https://github.com/liferay/liferay-frontend-projects/pull/97), even though it involved changes to both `eslint-config` and `npm-scripts`.

As you can see, it does three things:

- Blows away nested `node_modules` folders inside the monorepo, because `yarn add` can end up copying these over in ways that cause confusion (broken `bin` links and so on).
- Cleans out the existing versions of the dependencies (in my testing, removing and then re-adding seemed to produce a cleaner result than just overwriting).
- Adds the modules to be tested.

Like I said at the start, this is a hack (and not even trying to make it work on Windows), but it's a useful hack.

### DIFF
--- a/support/install-packages.sh
+++ b/support/install-packages.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+#
+# SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+# SPDX-License-Identifier: MIT
+
+##
+#
+# A hacky script to overwrite liferay-frontend-projects packages in a
+# liferay-repo checkout with the ones from this repo, for testing purposes.
+#
+# Usage:
+#
+#     bash install-packages.sh $PATH_TO_PORTAL_REPO
+
+die() {
+  echo "error: $*"
+  exit 1
+}
+
+remove() {
+  while (( "$#" )); do
+    local DEP=$1
+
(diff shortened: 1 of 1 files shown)


wincent:
cc: resident `#BashGuy`, @julien

hacker

---

Fun fact: the high-quality version of that GIF on Giphy is 33 MB.

jbalsas:
ducttape

wincent:
> Need any more duct tape?

Not a real review.

judgment

jbalsas:
> > Need any more duct tape?
> 
> Not a real review.

Review this

reviewthis

wincent:
over-the-line

So, I guess I'm just going to merge this wonderful code.

jbalsas:
LGTM! 😂

PR #104, opened 2020-10-02
chore: ignore yarn-error.log files

wincent:
Keep the clutter out of the `git status` output for these files that are all too easy to produce.


wincent:
Trivial, so going to merge.

PR #105, opened 2020-10-02
chore: update dependencies

wincent:
There is some risk in these updates but testing them is hard, so I am batching as many of these as possible to "rip the band-aid off" and get it all over and done with.

Having said that, I did this very carefully, looking at the output of `yarn outdated` and consulting the changelogs (or the "compare" views when there was no proper changelog) to detect likely problems. I think all of the updates are going to be ok.

The other thing I did was manipulate the ranges a bit to make things:

- **Easier:** For packages which follow semver (presumably everything), use "^" ranges to make updates easier and also make it less likely that we'll bring even more duplicates into liferay-portal. For many of these packages, I ran `yarn why` in liferay-portal to assess the current state of duplication there. I believe that when we ship this to liferay-portal it won't have any significant impact.
- **Safer:** For packages in the "< 1.0.0" range (where breaking changes are allowed by semver, I tightened some of those deps by removing "^" and locking to an exact version.

I did all of this with `yarn add`; no manual editing of package.json files.

This leaves the following output of `yarn outdated`:

    @storybook/addon-a11y       5.3.21  5.3.21 6.0.22 @liferay/npm-scripts dependencies https://github.com/storybookjs/storybook#readme
    @storybook/addon-actions    5.3.21  5.3.21 6.0.22 @liferay/npm-scripts dependencies https://github.com/storybookjs/storybook/tree/master/addons/actions
    @storybook/addon-knobs      5.3.21  5.3.21 6.0.22 @liferay/npm-scripts dependencies https://github.com/storybookjs/storybook/tree/master/addons/knobs
    @storybook/addon-viewport   5.3.21  5.3.21 6.0.22 @liferay/npm-scripts dependencies https://github.com/storybookjs/storybook/tree/master/addons/viewport
    @storybook/core             5.3.3   5.3.3  6.0.22 @liferay/npm-scripts dependencies https://github.com/storybookjs/storybook/tree/master/lib/core
    @storybook/react            5.3.21  5.3.21 6.0.22 @liferay/npm-scripts dependencies https://github.com/storybookjs/storybook/tree/master/app/react
    @testing-library/dom        5.6.1   5.6.1  7.24.3 @liferay/npm-scripts dependencies https://github.com/testing-library/dom-testing-library#readme
    @testing-library/jest-dom   4.2.4   4.2.4  5.11.4 @liferay/npm-scripts dependencies https://github.com/testing-library/jest-dom#readme
    @testing-library/react      8.0.9   8.0.9  11.0.4 @liferay/npm-scripts dependencies https://github.com/testing-library/react-testing-library#readme
    @testing-library/user-event 4.2.4   4.2.4  12.1.6 @liferay/npm-scripts dependencies https://github.com/testing-library/user-event#readme
    babel-jest                  25.5.1  25.5.1 26.3.0 @liferay/npm-scripts dependencies https://github.com/facebook/jest#readme
    css-loader                  3.6.0   3.6.0  4.3.0  @liferay/npm-scripts dependencies https://github.com/webpack-contrib/css-loader
    jest                        25.5.4  25.5.4 26.4.2 @liferay/npm-scripts dependencies https://jestjs.io/
    sass-loader                 8.0.2   8.0.2  10.0.2 @liferay/npm-scripts dependencies https://github.com/webpack-contrib/sass-loader

- Can't update the "loader" packages because the new versions all require a newer version of NodeJS than we have in liferay-portal.
- Not updating `@testing-library` stuff because they are lots of breaking changes and that deserves a PR and some extensive testing of its own.
- Likewise, not updating `@storybook` because that one has caused us problems with duplication and other headaches in the past, so is also worth of a dedicated PR.
- Not updating `jest` yet due to [an open performance issue](https://github.com/facebook/jest/issues/9457).

`yarn audit` is pretty clean after this:

    8 vulnerabilities found - Packages audited: 2798
    Severity: 6 Low | 2 High

Nothing to worry about there, but stuff coming in from `liferay-theme-tasks` and the ancient `metal-tools-soy`. We'll address at least the theme-tasks ones at some point, but not before moving that project into the monorepo.

Only other thing to note in here is a small Prettier update, which may create a little bit of churn when we run it in liferay-portal, but I don't think it will be too much.


projects/npm-tools/packages/js-insights/package.json
```diff
@@ -3,11 +3,11 @@
 	"dependencies": {
 		"@babel/parser": "^7.6.0",
 		"@babel/traverse": "^7.6.2",
-		"airtable": "0.8.1",
+		"airtable": "0.10.0",
```
jbalsas, line 6: Maybe just kill this project... just flat out delete `js-insights`... my poor baby 😂

projects/npm-tools/packages/npm-scripts/package.json
```diff
@@ -22,41 +22,41 @@
...
+		"html-webpack-plugin": "^4.3.0",
+		"http-proxy-middleware": "^1.0.5",
 		"jest": "^25.1.0",
 		"jest-environment-jsdom-thirteen": "1.0.1",
-		"jest-fetch-mock": "3.0.1",
+		"jest-fetch-mock": "^3.0.3",
 		"liferay-lang-key-dev-loader": "^1.0.3",
 		"liferay-npm-bridge-generator": "2.19.3",
 		"liferay-npm-bundler": "2.19.3",
 		"liferay-theme-tasks": "10.0.2",
```
jbalsas, line 51: Interesting that these ones are pinned... what's our plan moving onto the monorepo?

jbalsas (approved):
Fire at will

fireatwill

projects/npm-tools/packages/js-insights/package.json
wincent: Never!

over-my-dead-body

In all seriousness though, it's not hurting us. If it ends up causing problems in the dependency graph we can archive it — ie. move it outside of the yarn workspace:

```
mkdir archived
git mv projects/npm-tools/packages/js-insights archived/
```

and twiddle the links that point at the old location. We can also skip linting etc of `archived/`. One of the strengths of the monorepo is that we never have to worry "losing" stuff, because we can keep it all together.

projects/npm-tools/packages/npm-scripts/package.json
wincent: I will document this somewhere, but my thinking is:

- Projects with a good track record of following semver and which are >= 1.0.0: Use `^` ranges to make updating easier.
- Projects which are reckless with semver or < 1.0.0: Use `--exact` versions.
- Projects which are exactly 1.0.0: behavior and policy is undefined 😉 
- Our own internal projects: We can exactly lock these like we're already doing because we control the releases (ie. don't have to worry about duplicates), and because it allows us to do a controlled roll-out to liferay-portal (ie. if I `rm yarn.lock` in liferay-portal, I shouldn't suddenly get a new version of the bundler when I run `yarn install` — the only have to update our dependencies should be by updating `@liferay/npm-scripts`.

So Prettier is an interesting example here. We have traditionally locked it because we didn't want the `rm yarn.lock && yarn` scenario to pull in a new version and potentially churn a bunch of files; but at this point in Prettier's lifecycle, it is very stable and moving slowly, and they've shown that they respect semver, so there's no real reason to treat it differently than any other dependency.

wincent: Created https://github.com/liferay/liferay-frontend-projects/issues/106 to track documenting this, and to give us a place to discuss the details further if we need to.

PR #109, opened 2020-10-02
docs(npm-tools): update a stale link

wincent:
This one was still pointing at the old repo (which is now archived/read-only) instead of the monorepo.


wincent:
Trivial doc update — merging.

PR #101, opened 2020-10-01
fix(scripts): apply Liferay-specific Prettier overrides to TS also

wincent:
Noticed while testing a PR for [#92](https://github.com/liferay/liferay-frontend-projects/issues/92) that we aren't applying our Liferay-specific Prettier overrides to TS files. ie. we're not enforcing this:

    }
    else {

**Test plan:**

Add a ".ts" file, "x.ts" with these contents:

    if (true) {
    alert('t');
    } else {
        alert('f');
    }

Add an "npmscripts.config.js" file (you don't need the ".md" and ".yml" in there, but this is the file I had to create to test the other PR anyway, so I'm just using the same one):

    module.exports = {
            check: ['**/*.{md,ts,yml}'],
            fix: ['**/*.{md,ts,yml}'],
    }

Run `projects/npm-tools/packages/npm-scripts/bin/liferay-npm-scripts.js check` and see the error reported.

Run `projects/npm-tools/packages/npm-scripts/bin/liferay-npm-scripts.js fix` and see the code fixed to:

    if (true) {
            alert('t');
    }
    else {
            alert('f');
    }

PR #110, opened 2020-10-02
fix(npm-tools): make Prettier overrides work with TS

wincent:
So, I thought I had this working in [#101](https://github.com/liferay/liferay-frontend-projects/pull/101) (see the test plan), but when I got this into liferay-portal in [liferay-frontend/liferay-portal#415](https://github.com/liferay-frontend/liferay-portal/pull/415), I found it wasn't working (the `else` formatting wasn't applying). So I made the edits in this PR directly in the node_modules folder to realllllly see it working.

First step: switch in the TS parser when dealing with a TS file.

Second step: observe the same problem we saw (and fixed) in d8b03088be5c76fed — namely, that the TS parser doesn't put `start` and `end` properties on its tokens. We can switch to `range[0]` and `range[1]` respectively, making the rule work with both parsers.

Test plan: manually tested in-place, plus added automated tests.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/prettier/index.js
+++ b/projects/npm-tools/packages/npm-scripts/src/prettier/index.js
@@ -16,6 +16,7 @@
  * to version).
  */
 
+const typescriptEslint = require('@typescript-eslint/parser');
 const babelEslint = require('babel-eslint');
 const {CLIEngine, Linter} = require('eslint');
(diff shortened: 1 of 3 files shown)


wincent:
Who's going to review this on a Friday, _amirite_?
am-i-right

It's pretty well tested, so I'm going to merge it.

Issue #91, opened 2020-09-29
Integrate liferay-npm-scripts into Liferay Learn


rdai10:
The [Liferay Learn project](https://github.com/liferay/liferay-learn/) would like to use `@liferay/npm-scripts` to format its source code (`scss`, `js`, `md`, `yml`, and html in the form of `jinja 2` templates).

Preliminary work has been done to incorporate the package into the Learn project, but it's only being used to format `scss` so far. There are a few bumps on the road:
- The project uses a tooling called `Sphinx` that packages jquery out of the box. In the project's javascript files, there are references to the jquery `$` that throws errors when using npm-scripts. Are there any tips for configurations that we can add to get around that? A way to avoid inlining `<!-- prettier-ignore -->` comments would be preferred.
- There is a conscious effort to organize the project into `/docs` and `/site` folders at the root to create a separation between documentation material and its [website](https://learn.liferay.com/) source. However, we do want to be able to format all the md files in the `/docs` folder from inside the `/site/homepage` folder. I ran into issues defining an appropriate glob pattern to go up the directories for my `npmscripts.config.js` file. I came across https://github.com/liferay/liferay-npm-tools/pull/141 where a custom glob library was implemented for `liferay-npm-scripts` and hope to receive some help understanding what it expects.

Let me know if I can provide any clarification that will be helpful! Thanks!

rdai10:
@jrhoun, feel free to chime in!

wincent:
Thanks @rdai10.

I added [one comment here](https://github.com/brianchandotcom/liferay-learn/pull/69/commits/9a208f247fd9bc25673ed726ab339c31b6cf4673#r497271170):

> You don't need the `yarn` here.
>
> You also have a couple of options if you want to roll this out more "gently".
>
> - One is to use `fix` and `check` like you are here. Those are going to run _all_ of ESLint, Prettier, stylelint, plus any additional "pre-flight" checks that `@liferay/npm-scripts` implements. This is obviously a higher barrier which may add some friction in return for more consistency/quality (at least, that's the theory).
> - The other is to use `format` and `format:check` (ie. `liferay-npm-scripts format` and `liferay-npm-scripts format:check`). This _only_ runs Prettier, so the barrier to entry is lower and it is easier to turn on this lighter-weight check, and later on tighten things up by activating the broader check.

For example, I don't think it's Prettier which is throwing the error you're talking about here:

> there are references to the jquery `$` that throws errors when using npm-scripts.

[...]

PR #113, opened 2020-10-05
chore(eslint-config): mark typescript as a peer dependency

wincent:
This isn't _strictly_ necessary, but it is an accurate description of the requirements of the project. We go with the somewhat unsatisfying range of "*" to to minimize the risk of introducing duplicates in a consuming project. At the moment, we're using v4+ in this monorepo, 3.9.7 in liferay-portal, and the lowest range that we're currently using in any other project is "^3.6.3" (in liferay-js-toolkit). In the future, we should be able to converge on v4+.

If you have a ".ts" file in your project, you are going to have `typescript` in your `devDependencies`, and our `overrides` configuration is going to use `@typescript-eslint/parser`, which in turn is going to require `typescript`, transitively, in `@typescript-eslint/typescript-estree` [here](https://github.com/typescript-eslint/typescript-eslint/blob/c5ffe8833da7875e82953de117b0abe6fa8a60b0/packages/typescript-estree/src/parser.ts#L2).

And all will work, even though `@typescript-eslint/typescript-estree` doesn't explicitly declare its hard dependency on `typescript` in [its `package.json`](https://github.com/typescript-eslint/typescript-eslint/blob/c5ffe8833da7875e82953de117b0abe6fa8a60b0/packages/typescript-estree/package.json#L41-L49).

But if you don't have any ".ts", then you shouldn't need to have `typescript` in your dependency graph, and we won't use `@typescript/eslint-parser`. So, it makes sense for us to call this an `peerDependency` and mark it as `optional` as per these posts:

- https://medium.com/@noamgjacobsonknzi/what-is-peerdependenciesmeta-acea887c32dd
- https://github.com/npm/cli/issues/1247

(Note that the `peerDependenciesMeta` field isn't documented yet.)

You could also argue that this should be under [`optionalDependencies`](https://docs.npmjs.com/files/package.json#optionaldependencies), but I think "optional peer" comes closer to the reality here. `optionalDependencies` is for things that you can add, but the package should work if they're not there. That's not the case here. The package won't work if you have ".ts" files but no `typescript`.

Related: https://github.com/liferay/liferay-frontend-projects/issues/91#issuecomment-702881845

### DIFF
--- a/projects/eslint-config/package.json
+++ b/projects/eslint-config/package.json
@@ -33,7 +33,13 @@
 	"main": "index.js",
 	"name": "@liferay/eslint-config",
 	"peerDependencies": {
-		"eslint": ">=4.1.1"
+		"eslint": ">=4.1.1",
+		"typescript": "*"
+	},
+	"peerDependenciesMeta": {
+		"typescript": {
+			"optional": true
+		}
 	},
 	"repository": {
 		"directory": "projects/eslint-config",
--- a/yarn.lock
+++ b/yarn.lock
@@ -16039,7 +16039,7 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=
 
-typescript@4.0.3:
+typescript@*, typescript@4.0.3:
   version "4.0.3"
   resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.0.3.tgz#153bbd468ef07725c1df9c77e8b453f8d36abba5"
   integrity sha512-tEu6DGxGgRJPb/mVPIZ48e69xCn2yRmCgYmDugAVwmJ6o+0u1RI18eO7E7WBTLYLaEVVOhwQmcdhQHweux/WPg==


jbalsas:
I'm assuming you know what you're doing, so go ahead 😂 

If I understand correctly, this means that you'll always need to have both `@liferay/eslint-config` and `@typescript` as devDeps (unless you don't have `.ts` as you point out). 

What does this mean for those simply adding `@liferay/npm-scripts` as a `devDependency`? I would expect that perhaps we wouldn't need to add `@typescript` independently.

wincent:
> If I understand correctly, this means that you'll always need to have both `@liferay/eslint-config` and `@typescript` as devDeps (unless you don't have `.ts` as you point out).

Not exactly. If you're a project like liferay-amd-loader, you won't need `typescript` because it is only a peer dep, and an optional one at that (which means you shouldn't see a warning printed when you don't have it). Because liferay-amd-loader contains no ".ts" files, we'll never try to `require('@typescript-eslint/parser')` which means that everything should just work fine.

> What does this mean for those simply adding `@liferay/npm-scripts` as a `devDependency`? I would expect that perhaps we wouldn't need to add `@typescript` independently.

If you look in the next PR (https://github.com/liferay/liferay-frontend-projects/pull/114) you'll see what it means for projects that just add `@liferay/npm-scripts`: they'll get `typescript` as well, because our code requires `@typescript-eslint/parser` and therefore needs `typescript`. Basically, as soon as you pull in `@liferay/npm-scripts`, even if only for formatting (eg. liferay-learn's case), you're going to get the whole "batteries included" set of everything we use (that includes things like webpack, storybook, liferay-npm-bundler). It's overkill, but it means that we can guarantee that everything works together and is there for you "automatically".

jbalsas:
>[...] Basically, as soon as you pull in @liferay/npm-scripts, even if only for formatting (eg. liferay-learn's case), you're going to get the whole "batteries included" set of everything we use (that includes things like webpack, storybook, liferay-npm-bundler). It's overkill, but it means that we can guarantee that everything works together and is there for you "automatically".

Music to my ears ❤️ 

Fire at will!!

PR #114, opened 2020-10-05
fix(npm-scripts): bundle TypeScript dependencies

wincent:
As reported [here](https://github.com/liferay/liferay-frontend-projects/issues/91#issuecomment-702881845), our use of the `@typescript-eslint/parser` (since 50a4d99417b2461) means that we have a runtime dependency on `typescript` package. This worked fine in liferay-portal, where `typescript` is already present, but not in liferay-learn, where it is not. And I didn't know that `@typescript-eslint/parser` doesn't actually declare all of its transitive dependencies; as noted in [#113](https://github.com/liferay/liferay-frontend-projects/pull/113), `@typescript-eslint/typescript-estree` does an unguarded `import` of `typescript` even though it is not declared in its dependencies (they declare it as an optional peer dependency).

So, let's not only fix this, by explicitly declaring our dependency on TypeScript, but also declare all of our TS-related dependencies. This will allow us to delete the devDependencies currently declared [here](https://github.com/liferay/liferay-portal/blob/f1a2b2ce06fa050d5e1ab193728ab9ec06df4315/modules/apps/remote-app/remote-app-client-js/package.json#L3-L6).


wincent:
Marking this one as a draft because it depends on #113 going in first.

wincent:
I'm amazed that GitHub actually did something useful for once. As soon as I:

1. Merged #113
2. Deleted the corresponding branch
3. Took this out of draft

...it did this:

Screenshot 2020-10-05 -114138-TBSxlmZC@2x

Which is exactly what I wanted it to do... (not sure if it did it it after (1) or (2) or (3), but whatever).

blown-away

wincent:
As this one is intimately related to the (merged/approved #113) I'm going to merge this one too.

PR #118, opened 2020-10-05
docs: encourage people to dog-food local packages

wincent:
A few places where we're telling people to use the last-published version of the changelog-generator (via `npx`) tell them to use `yarn run` instead; this means they'll get the current version in the workspace instead.

About the only time this won't work is if you haven't run `yarn` yet, which I'm not particularly worried about, or you have broken the local copy of the generator (and that's one of the main reasons why we like dog-fooding anyway; to find out when we've broken things).

We also get rid of one `npx` usage in eslint-config's `package.json file. That runs the last-published version of js-publish, but we should instead  directly run the one that we have in the repo (we're already doing this
in the npm-tools packages).

### DIFF
--- a/projects/eslint-config/CONTRIBUTING.md
+++ b/projects/eslint-config/CONTRIBUTING.md
@@ -14,11 +14,11 @@
 
     Run [`@liferay/changelog-generator`](https://www.npmjs.com/package/@liferay/changelog-generator):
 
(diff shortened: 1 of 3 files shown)


wincent:
This is trivial so I am going to `#CowboyIt`

cowboy-it

PR #119, opened 2020-10-05
style: fix formatting on imported projects

wincent:
This shows one of the nice things about the monorepo: even though we have two "products" in here (the 2.x series living under "maintenance/projects/" and the 3.x series living under "projects/"), we can apply uniform formatting to both of them that matches the formatting used everywhere else in the monorepo. And this is true even though 2.x is isolated from top-level Yarn workspaces.

Note that we also delegate the imported project "format"/"format:check" scripts back to the top level (we always want to do a top-down format to enforce uniformity across the entire repo) and remove the unnecessary formatting config files nested inside the projects.

CI won't be green on this (still other stuff to unbreak, but at least the formatting check should pass now).


wincent:
As this is a machine-generated commit, and CI likes the formatting:

Screenshot 2020-10-05 -185733-IcDb6vHN@2x

I'm going to merge this so that I can build on top of it.

PR #120, opened 2020-10-05
docs: advise people to remove the temporary remote after importing

wincent:
To avoid any unpleasantness, like accidentally pushing unwanted tags the next time you run `git push` with `--tags` or `--follow-tags`, or the simple annoyance of having to delete dozens or hundreds of tags.


wincent:
Going to merge this to reduce clutter. It's just a doc update.

PR #121, opened 2020-10-05
docs: hoist js-tookit licensing info up to top-level

wincent:
Just like we did for the other projects in e6786ef0300ac3607202fb5f092f966f6971ebbb and 4793d6e8405fe703dba9517fbc61ba8f55799a5f.


wincent:
Again, don't expect this to be green, because the only CI task I've fixed so far is the formatting one.

wincent:
Just moving docs around so I am going to merge this to reduce clutter.

PR #122, opened 2020-10-05
chore: configure labeler for js-toolkit

wincent:
Anything touching either 2.x or 3.x gets a "js-toolkit" label.

Additionally, we add "2.x" and "3.x" to distinguish the two projects.

(Note that there is no problem with having a "2.x" label eventually being shared by another project, like, say, AlloyEditor, because these projects can always be distinguished by looking at the combination of the version number label and the project name label.)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -3,6 +3,14 @@
 #
 # See: https://github.com/actions/labeler
 
+2.x:
+    - maintenance/projects/js-toolkit/*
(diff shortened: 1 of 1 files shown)


wincent:
Again, don't expect this to be green yet. Still working through the issues.

wincent:
Just a config change so I think it is pretty low-risk to merge; going to do so in order to reduce clutter.

PR #124, opened 2020-10-06
docs(js-toolkit): add links from top-level README and CONTRIBUTING

wincent:
I took the liberty of backtick-izing some project names to make them stand out from the version numbers, seeing as js-toolkit is the first project we have here that obliges us to deal with versions.


wincent:
Again, don't expect this to be green yet.

wincent:
Innocuous documentation updates, so not worried about risk. Going to merge.

PR #126, opened 2020-10-06
fix(js-toolkit): adjust linting to work in monorepo

wincent:
-   We delegate the task of running linting to the top-level.
-   We hoist the ignore rules up to the top (off the top of my head, I don't recall whether ESLint requires this, but Prettier does, so we may as well be consistent and keep everything together; additionally, if we every dog-food liferay-npm-scripts as a formatting tool in this repo, we'll need the config at the top).
-   Switch to `@liferay` preset and adjust rule names to match.
-   Run lint autofixes.
-   Do remaining lint fixes by hand.


wincent:
This won't be green (still other stuff to fix), but at least the lint and format checks should be green.

wincent:
I see linting failed in CI, even though it passed locally. Will investigate.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/babel-util.ts
```diff
@@ -161,14 +163,6 @@ export function getModuleName(moduleFilePath: string): string {
...
 }
 
-function getPackageName(modulePath: string): string {
-	const modulePathParts = modulePath.split('/');
-
-	return modulePath.startsWith('@')
-		? `${modulePathParts[0]}/${modulePathParts[1]}`
-		: modulePathParts[0];
-}
-
```
wincent, line 171: This function was unreferenced.

wincent:
[Linting is green in CI](https://github.com/liferay/liferay-frontend-projects/pull/126/checks?check_run_id=1214370932) (the red test failures are expected because I haven't started on getting those working yet).

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/jar.ts
```diff
@@ -1,15 +1,13 @@
 /**
- * © 2017 Liferay, Inc. <https://liferay.com>
- *
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
  * SPDX-License-Identifier: LGPL-3.0-or-later
  */
 
+import {Project} from '.';
```
wincent, line 6: This automatic reordering is hopefully benign. Importing from "." is a bit weird, so it's not a case that we planned for in our import auto-sorter rule.

This is getting treated as an "absolute" source because it doesn't start with "./" or "..", as per [the  `isRelative()` check here](https://github.com/liferay/liferay-frontend-projects/blob/d396bc98df0cb69ff29917647929c4ff5d03044b/projects/eslint-config/plugins/liferay/lib/common/imports.js#L154-L156). I'll make a separate issue to handle this special case.

wincent: Here's the issue: https://github.com/liferay/liferay-frontend-projects/issues/129

maintenance/projects/js-toolkit/packages/liferay-npm-bundler-plugin-replace-browser-modules/src/index.ts
```diff
@@ -124,6 +124,9 @@ function processAliases(
 					).linkToCode(2);
 					break;
 				}
+
+				default:
+					break;
```
wincent, line 129: We could throw an error here for an unhandled case, but I don't know all the cases and prefer not to get side-tracked looking for them at this point. We can always loop back to this and do an audit, and in the meantime, we are probably not making things worse by doing this... I think... the alternative would be to add a suppression, which at least would give us something to grep for later on. 🤷

maintenance/projects/js-toolkit/packages/liferay-npm-bundler-plugin-replace-browser-modules/src/index.ts
```diff
@@ -177,22 +182,26 @@ export function reportAndResolveCollisions(
...
 		if (unrolledAliasesMap[absFromPath].length == 0) {
 			delete unrolledAliasesMap[absFromPath];
 		}
 	});
 
 	// Resolve collisions in multiple length aliases
+
 	Object.entries(unrolledAliasesMap)
-		.filter(([absFromPath, unrolledAliases]) => unrolledAliases.length > 1)
+		.filter(([_absFromPath, unrolledAliases]) => unrolledAliases.length > 1)
```
wincent, line 195: Could have done:

```ts
.filter(([, unrolledAliases]) => unrolledAliases.length > 1)
```

too. But didn't want to risk being obfuscated.

yarn.lock
```diff
@@ -7867,6 +8089,20 @@ globby@^6.1.0:
     pify "^2.0.0"
     pinkie-promise "^2.0.0"
 
+globby@^9.2.0:
```
wincent, line 8092: Current output of `yarn why globby | grep Found`:

```
=> Found "globby@10.0.2"
=> Found "@liferay/js-insights#globby@11.0.1"
=> Found "@liferay/npm-scripts#globby@11.0.1"
=> Found "mem-fs-editor#globby@9.2.0"
=> Found "@liferay/npm-scripts#liferay-npm-bundler#globby@10.0.2"
=> Found "@liferay/npm-scripts#liferay-npm-bridge-generator#globby@10.0.2"
=> Found "@typescript-eslint/typescript-estree#globby@11.0.1"
=> Found "stylelint#globby@11.0.1"
=> Found "liferay-theme-tasks#globby@4.1.0"
=> Found "metal-tools-soy#globby@6.1.0"
=> Found "react-dev-utils#globby@8.0.2"
=> Found "del#globby@6.1.0"
=> Found "yeoman-generator#globby@8.0.2"
=> Found "yeoman-environment#globby@8.0.2"
=> Found "yeoman-environment#yeoman-generator#globby@9.2.0"
```

🤦‍♂️

I'll see if I can do a wee bit of de-duping as I get the workspaces properly configured (in another PR), but realistically, we are going to be stuck with a few versions of this darn library no matter what we do.

wincent:
As this is a mostly automated change (some manual tweaks noted inline in comments), I'll just merge this in order to keep building on top of it. Will fast-follow with any fixes that are necessary.

PR #131, opened 2020-10-06
chore(js-toolkit): update info and links in package.json files

wincent:
- Remove stuff that doesn't make sense in private packages because they never get published to the registry (ie. anything with links). Likewise, we remove the "workspaces" from "projects/js-toolkit/package.json" because it is now participating in the monorepo's top-level workspaces configuration. I didn't touch the 2.x equivalent under "maintenance/projects/js-toolkit" because that one _is_ using its own workspace set-up.
- Add "author" to all public packages.
- Add "repository" info to all public packages.

Not sorting the keys yet, but will do that in a separate PR, for consistency with all the other package.json in the monorepo.

### DIFF
--- a/maintenance/projects/js-toolkit/package.json
+++ b/maintenance/projects/js-toolkit/package.json
@@ -1,13 +1,4 @@
 {
-	"homepage": "https://github.com/liferay/liferay-js-toolkit#readme",
-	"license": "LGPL-3.0",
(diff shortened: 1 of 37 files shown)


wincent:
"No logic changes", so I'm going to get this in in order to continue building on top of it. If I broke anything will follow-up with a fix.

PR #132, opened 2020-10-06
chore(js-toolkit): sort package.json files

wincent:
Done using:

    find . \
      -name package.json \
      -path '*/js-toolkit/*' \
      -not -path '*/node_modules/*' \
      -not -path '*/__tests__/*' \
      -not -path '*/__fixtures/__*' \
      -not -path '*/qa/samples/*' | node support/sortJSON.js

and the included helper script.

We may want to eventually integrate something like the `sortJSON.js` script into our formatting runs, but I expect these package.json files to stay in order for at least a while now, so I don't want to prematurely overcomplicate things.

Test plan: Make this commit and visually inspect the changes.

### DIFF
--- a/maintenance/projects/js-toolkit/package.json
+++ b/maintenance/projects/js-toolkit/package.json
@@ -1,8 +1,20 @@
 {
+	"dependencies": {
+		"xml-js": "^1.6.8"
+	},
+	"devDependencies": {
+		"@types/jest": "^24.0.18",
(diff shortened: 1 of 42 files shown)


wincent:
Still not expecting this to be green; need to fix build and test for that, but this should be able to go in anyway.

wincent:
As this is just a mechanical change, going to merge it and keep going from here.

PR #135, opened 2020-10-07
refactor(js-toolkit): hoist devDependencies out of project and packages to top level

wincent:
For the same reasons as in liferay-portal, we don't want `devDependencies` down in nested folders; we just want one "standard" version up the top.

After this PR, there are no `devDependencies` at all in the js-toolkit v3 top-level package.json or in the individual packages.

I'm not touching v2 because that is isolated off in a workspace all of its own.


wincent:
Just config changes, so going to merge.

Issue #129, opened 2020-10-06
sort-imports rule thinks "." is an absolute path


wincent:
### Issue type

-   :bug: Bug report

### Description

As described in [this comment](https://github.com/liferay/liferay-frontend-projects/pull/126/files#r500187918):

```js
import X from '.';
```

should be considered a relative import, but we are treating it like an absolute one.

Now, importing from ".", is weird and I am not sure that we should do it, but classifying "." as a non-relative path _is_ a bug, so we should fix it anyway. Will just need to add one special case branch to [our `isRelative()` implementation](https://github.com/liferay/liferay-frontend-projects/blob/d396bc98df0cb69ff29917647929c4ff5d03044b/projects/eslint-config/plugins/liferay/lib/common/imports.js#L154-L156).

PR #137, opened 2020-10-07
fix(eslint-config): special case import of "."

wincent:
I don't like these "." imports as there are not very idiomatic; I guess they're not common because they are subtle and potentially confusing. If you are in a folder and "./index" is the manifest of everything exportable in that folder, then that would normally include the file doing the import, which is circular. So the cognitive load of thinking that through is higher than it need be.

Anyway, even though I don't like them, it doesn't mean I shouldn't fix the bug. So here goes.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/129

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/copy.ts
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/copy.ts
@@ -3,9 +3,9 @@
  * SPDX-License-Identifier: LGPL-3.0-or-later
  */
 
-import {Project} from '.';
 import prop from 'dot-prop';
 
+import {Project} from '.';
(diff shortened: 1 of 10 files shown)


wincent:
As I wrote this, I was wondering if a simpler test might suffice, like `source.startsWith('.')`, but I was worried that it would give a false positive if you ever wound up with a package name that started with "." (npm shouldn't allow this, but just in case it ever did... or you created one manually...).

So I renamed a package in a `node_modules` folder to `.crazy` to test this out, and indeed you can see in the REPL that `node` lets you `require` it:

```
> require('.crazy')
{ default: [Function: isObject] }
```

In other words, I think this is the appropriate fix. It's pretty simple and the tests are green so I'm going to merge it. No hurry to cut an `eslint-config` release because of this because the only projects where we're using these weird imports is already right here in the monorepo, where the fix already applies.

PR #138, opened 2020-10-08
chore(js-toolkit): configure release messages consistently

wincent:
- Adjust existing .yarnrc files so that the release commit message matches the format used by other packages in this monorepo.
- Add .yarnrc files for packages that don't have them yet.
- Configure Lerna (still used in v2 of the js-toolkit) to produce commit messages that match the format produced by yarn; previously they produced commit messages consisting just of the bare version number (eg. see b493e109afa2c706ce5fcea79b81a54ac72578d9), which isn't appropriate for a multi-project monorepo.

### DIFF
--- a/maintenance/projects/js-toolkit/lerna.json
+++ b/maintenance/projects/js-toolkit/lerna.json
@@ -3,6 +3,9 @@
 	"command": {
 		"publish": {
 			"exact": true
+		},
+		"version": {
+			"message": "chore: prepare liferay-js-toolkit/%s release"
 		}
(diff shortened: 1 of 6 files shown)


wincent:
I don't think there is any good way to test the Lerna change (I don't think `lerna publish` has a "dry-run" switch), but I wanted to do a release anyway in order to verify that we actually can do so from the monorepo, so I will try that today.

wincent:
Going to merge this so I can try releasing.

PR #139, opened 2020-10-08
docs(monorepo): add a bunch more detail on how to import a project

wincent:
Given that the js-toolkit import was quite complicated (two branches, 35 packages), I figure it warrants noting more detail in this document about how to update that kind of thing.


wincent:
Just simple doc updates so going to merge this.

PR #140, opened 2020-10-08
docs(js-toolkit): update links prior to release

wincent:
As I want to cut a release (just a patch release, to prove that releasing from the monorepo works, and to update the README and metadata on the package pages on the npm registry) I'm just sweeping through and updating links and other information that needs to be tweaked now that we're in the monorepo.

The one link I didn't update was the link to the wiki, because we haven't migrated that yet (we still need to decide whether to migrate it, or whether to move the contents into a `docs/` folder).

For consistency, snuck in one edit to the top-level README to make sure we use the same alt text on our CI badges everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # liferay-frontend-projects monorepo
 
-![](https://github.com/liferay/liferay-frontend-projects/workflows/ci/badge.svg)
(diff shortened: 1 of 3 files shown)


wincent:
Just docs, nothing much to see here...

move-along

PR #141, opened 2020-10-08
chore(js-toolkit): make post-release tweaks

wincent:
Just some small changes to reflect what I learned while doing our first release of the v2.x series from the monorepo context.

As a bonus, you get a free set of Conventional Commits types!

- "chore" to describe the overall feeling I had when preparing this PR.
- "fix" because the Lerna config was wrong.
- "docs" because I edited docs.
- "style" because Lerna seemed to mess up the formatting during publish, so I had to run `yarn format`.


jbalsas:
> As a bonus, you get a free set of Conventional Commits types!

imnotcrying

PR #144, opened 2020-10-08
docs(js-toolkit): clean up imported js-toolkit wiki

wincent:
I already merged the history to `master` in 9d081a70 (note: CI will be red on that due to formatting issues).

In this PR I clean things up and (hopefully) fix all the links...


wincent:
I can see a few links that point at code in the old repo, so going to tackle those in an additional commit.

wincent:
Just moving docs around, so may as well merge it. If we find any more broken links can fix later.

Issue #128, opened 2020-10-06
feat: use liferay-npm-scripts to do linting and formatting inside the monorepo


wincent:
So in getting linting to work in the js-toolkit project I'm seeing the classic fight between Prettier and ESLint about how comments should be formatted.

Prettier wants:

```ts
export interface X {
  /**
   * Yo!
   */
  thing: string;
}
```

while ESLint wants:

```ts
export interface X {

  /**
   * Yo!
   */
  thing: string;
}
```

(There's a similar fight for `//` comments as well.)

Now, js-toolkit has a couple dozen of these darn comments, so I either add a suppression or I delete the comments. I'm going to add a suppression, and create this issue to remind me to remove it.

The definitive solution will be to dog-food liferay-npm-scripts and use it in the monorepo to do the formatting. As seen in bf5d84c475fb478bfa3378b9ab54d012e935a063, we don't have this problem in liferay-portal thanks to liferay-npm-scripts (because it runs ESLint + Prettier + post-processing, we can resolve little conflicts like this). Obviously, there's a benefit to dog-fooding anyway, especially now that other projects like liferay-learn are using npm-scripts too (ie. it is useful to us to be dog-fooding it all the time, so that we find out if we ever break the non-liferay-portal use case).

Possibly controversially, this change will bring the "Liferay" `else` formatting into the monorepo, which FWIW, is probably not a bad thing (one less thing for us to think about; just do it the same everywhere.)

jbalsas:
> [...] inside the monorepo

I swear I skimmed over the title in GMail and literally read "inside this monstrosity" 😂

jbalsas:
> The definitive solution will be to dog-food liferay-npm-scripts and use it in the monorepo to do the formatting

I think this is really the only reasonable path to take

wincent:
```js
if (brave) {
    beTheChangeYouWantToSeeInTheWorld();
}
else {
    doNothing();
}
```

![my-eyes](https://user-images.githubusercontent.com/7074/95195493-e61d2e80-07d6-11eb-88ff-7ccf8046a7d1.gif)

PR #146, opened 2020-10-09
feat(monorepo): use liferay-npm-scripts for formatting

wincent:
This resolves disagreements between Prettier and ESLint about how comments should be formatted; eg. Prettier's preference:

```ts
export interface X {
  /**
   * Yo!
   */
  thing: string;
}
```

vs ESLint's preference:

```ts
export interface X {

  /**
   * Yo!
   */
  thing: string;
}
```

by passing everything through liferay-npm-scripts. It provides us a Prettier-like wrapper which abstracts over all this, and effectively runs Prettier + a post-processing pass under the covers.

This also means that we'll use the exact same formatting style in this repo as we use in liferay-portal.

Includes some necessary follow-up commits to ensure that our VSCode and Vim (etc) integrations keep working as expected. Note that I had to teach our `prettier` proxy the `--list-different` option in order for it to be usable as a substitute in our `format:check` job.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/128


projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
```diff
@@ -1059,28 +1102,52 @@ function prepare(filepath) {
...
-		if (extension) {
-			dir = path.join(extension.extensionPath);
-			file = path.join(dir, 'node_modules/prettier');
+			if (extension) {
+				dir = path.join(extension.extensionPath);
+				file = path.join(dir, 'node_modules/prettier');
+			}
 		}
 	}
+
```
wincent, line 1138: This is the least satisfying part of the entire PR. I wrote it late at night, so maybe it could be cleaned up a bit, but I feel there is something inescapably ugly about all the hackery we have to do in this script.

forgive-me

projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.sh
```diff
@@ -5,11 +5,27 @@
 # "liferay-npm-scripts prettier" subcommand to plugins.
 #
 
-# Auto-detect liferay-portal checkout.
+LIFERAY_NPM_SCRIPTS=""
+LIFERAY_PRETTIER=""
+
+# Auto-detect a liferay-portal or liferay-frontend-projects checkout.
```
wincent, line 11: The following section is also pretty ugly. Not quite as bad as the VSCode one, but still not great.

shaking-head

wincent:
As this touches almost everything, we kind of have to get this in quickly (because it will conflict with anything that comes after it), so [given this resounding endorsement](https://github.com/liferay/liferay-frontend-projects/issues/128#issuecomment-704198635), the fact that this change is mostly the result of automation, and it's green, I'm going to merge it.

yeehaw

PR #153, opened 2020-10-13
chore(js-toolkit): move liferay-npm-bridge-generator (v3.x) to @liferay/npm-bridge-generator

wincent:
Step 4 of 5.

Previously:

1. https://github.com/liferay/liferay-frontend-projects/pull/150 (js-toolkit-core)
2. https://github.com/liferay/liferay-frontend-projects/pull/151 (generator-js)
3. https://github.com/liferay/liferay-frontend-projects/pull/152 (js-toolkit-scripts)

A draft because it depends on the previous ones and they need to go in first.

PR #152, opened 2020-10-13
chore(js-toolkit): move liferay-js-toolkit-scripts (v3.x) to @liferay/js-toolkit-scripts

wincent:
Step 3 of 5.

Previously:

1. https://github.com/liferay/liferay-frontend-projects/pull/150 (js-toolkit-core)
2. https://github.com/liferay/liferay-frontend-projects/pull/151 (generator-js)

A draft because it depends on the previous ones and they need to go in first.

PR #151, opened 2020-10-13
chore(js-toolkit): move generator-liferay-js (v3.x) to @liferay/generator-js

wincent:
Step 2 of 5. Previously:

1. https://github.com/liferay/liferay-frontend-projects/pull/150 (js-toolkit-core)

Marking this one as a draft because it depends on the other one going in first.

PR #150, opened 2020-10-13
chore(js-toolkit): move liferay-js-toolkit-core (v3.x) to @liferay/js-toolkit-core

wincent:
Step 1 of 5. Will migrate the other packages after this.

PR #154, opened 2020-10-13
chore(js-toolkit): move liferay-npm-bundler (v3.x) to @liferay/npm-bundler

wincent:
Step 5 of 5.

Previously:

1. https://github.com/liferay/liferay-frontend-projects/pull/150 (js-toolkit-core)
2. https://github.com/liferay/liferay-frontend-projects/pull/151 (generator-js)
3. https://github.com/liferay/liferay-frontend-projects/pull/152 (js-toolkit-scripts)
4. https://github.com/liferay/liferay-frontend-projects/pull/153 (npm-bridge-generator)

A draft because it depends on the previous ones and they need to go in first.


wincent:
@izaera: This is the most complicated/risky one of the lot so I'd appreciate it if you could look over it with your 🦅 👀  — The others were more of a mechanical thing, but in this one there are quite a few interdependencies... eg.

- Places where the code still refers to v2.x of liferay-npm-bundler, and so I didn't change those... vs...
- Places where it seemed straightforwardly obvious and appropriate to update to switch to the namespaced `@liferay/npm-bundler`...

And then there are lots of references sprinkled through here to packages from v2.x of the toolkit that were never on the `3.x-WIP` branch in the other repo, and it's not clear to me whether they will ever be ported, or whether the references themselves should just be removed (basically, stuff I found via `git grep liferay-npm-bundler` eg. presets and plugins).

izaera:
The "problem" is that the generator in `3.x-WIP` has not been migrated yet, it's just a copy of the one in `master`. That is why it refers to `2.x` versions.

I wouldn't bother too much in migrating it, as it will need a rewrite because the structure of the projects has changed. So I guess you may leave it in any state that doesn't break the CI.

Regarding the bridge generator it is also the same as in 2.x and it probably won't need to be changed, as it works the same for 3.x.

Finally the js-toolkit-scripts should disappear too, as everything done there will be done inside bundler 3, so nothing will be modular as in 2.x.

To sum up, the final bundler 3 should have:

1. toolkit-core: a library with all utilities to deal with files, transformations, project inspection, etc. It should also be used (in the final image) by other projects, like the Themes Toolkit, for example. It could even be used by third party tools like the Fragments CLI or anything that any customer/partner develops. In other words: it is our programmatic API.
2. bundler: relies on toolkit-core to bundle JS projects
3. generator: generates projects that are built by bundler
4. bridge generator (perhaps): we need this tool for backwards compatibility in the portal, but we will remove/deprecate it as soon as we don't need it any more

wincent:
Thanks @izaera. As per Slack discussion, as long as CI is green (it is), we're just going to cut some prereleases and see if everything works.

The important bits to get right are the bundler and the toolkit core (used by the bundler). They are used in about three places in DXP so we can test all of those.

Known issue: QA script doesn't work in v3 (because the generator isn't really migrated, like Iván said above).

wincent:
Rebasing this one now.

wincent:
It's green so it obviously works. Without a doubt. Guaranteed!

PR #155, opened 2020-10-14
docs(js-tookit): update CHANGELOG files for first named scope releases

wincent:
As these first releases under the `@liferay` named scope are all a bit manual and weird, I'm batching the (basically manual) updates for all the packages into this one commit. Note that there were a couple of packages there that we never officially released, so they get CHANGELOG files for the first time. (We may not end up continuing to release those packages in the future — they may get refactored and cease to exist independently — but in the short term we want to respect the "Principle of Least Surprise" and ensure that everything under `packages` in the monorepo is both publishable and published, even if only as an ephemeral prerelease).

Other than that, these packages should all be basically identical to the non-named-scope cousins with the same version number. Only the names have changed.


jbalsas:
I'll merge this, I guess... looks legit

PR #158, opened 2020-10-14
chore(js-toolkit): update release process docs and configuration

wincent:
This was originally going to be a "docs" type task to update this doc to rectify the staleness that I discovered while doing our first `yarn version`-powered releases of js-toolkit packages in the `@liferay/` named scope today:

- https://github.com/liferay/liferay-frontend-projects/releases/tag/js-toolkit-core%2Fv3.0.1-pre.0
- https://github.com/liferay/liferay-frontend-projects/releases/tag/npm-bundler%2Fv3.0.1-pre.0

In the end, though, going with "chore" because I also had to strip out some misleading config in the .yarnrc. The `--publish.tag` config is never going to be used, because now that we are on the `master` branch, we will use `@liferay/js-publish` just like any other project, which means you release by running `yarn version`, not `yarn publish`. And `@liferay/js-publish` sets a tag of "prerelease" by default for any prerelease version (ie. anything with a hyphen in it), so this config is never going to take effect during a standard release flow, and keeping it around in there is only going to cause confusion. No other projects in the monorepo have a `--publish.tag` config.

### DIFF
--- a/projects/js-toolkit/.yarnrc
+++ b/projects/js-toolkit/.yarnrc
@@ -1,2 +1,1 @@
---publish.tag next
 --version.preid pre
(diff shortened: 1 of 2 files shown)


wincent:
Trivs?

PR #196, opened 2020-10-16
feat(changelog-generator): add --major, --minor etc convenience switches

wincent:
Still a draft because I want to refactor this to clean it up.

But in short, what this gives us is a set of new options:

- `--major`
- `--minor`
- `--patch`
- `--premajor`
- `--preminor`
- `--prepatch`
- `--prerelease`
- `--preid`

That match the options and behavior of the corresponding `yarn` ones.

These can be used to make the release process for most packages look like this:

    npx @liferay/changelog-generator --dry-run  # preview changes, decide on release type
    npx @liferay/changelog-generator --patch    # actually generate changes
    git add -p                                  # stage changes
    yarn version --patch                        # do release

And then on top of that, I added another switch:

- `--interactive`/`-i`

That turns the release process into this:

    npx @liferay/changelog-generator --interactive
    yarn version --patch

ie. it basically does this:

1. Run like `--dry-run`, and show a preview of what _would_ be changed without actually writing to disk; based on this, the user can decide on a release type (major, minor etc).
2. Prompt user for a version number (can be "major" etc or an explicit number).
3. Actually write out changes with the version number in effect.
4. Show `git diff` output and offer to `git add`.

Now, as I said above, there is a bit of refactoring to do before I'll be happy with this. The `index.js` is now getting cluttered enough that some of the peripheral bits could be swept under the carpet somewhere else (ie. into other files, never to be looked at again), and the whole "state-machine" aspect of `--interactive` could use some clean up. Once I've done that I'll take this out of draft.

### DIFF
--- a/projects/npm-tools/packages/changelog-generator/src/index.js
+++ b/projects/npm-tools/packages/changelog-generator/src/index.js
@@ -9,6 +9,7 @@ const {promisify} = require('util');
 
 const {cleanup, error, info, log, prompt, warn} = require('./console');
 const git = require('./git');
+const printBanner = require('./printBanner');
 const readYarnrc = require('./readYarnrc');
 
 const readFileAsync = promisify(fs.readFile);
@@ -251,56 +252,6 @@ async function formatChanges(changes, remote) {
 		})
 		.filter(Boolean)
 		.join('\n\n');
-}
-
-/**
- * Prints `message` in a silly banner like this:
- *  ____________________________________
- * (_)                                  `
- *   |                                  |
- *   |   @liferay/changelog-generator   |
- *   |   ============                   |
- *   |                                  |
- *   |   Reporting for duty!            |
- *   |                                  |
- *   |__________________________________|
- *   (_)________________________________)
- *
- */
-function printBanner(message) {
-	const lines = message.split('\n').map((line) => line.trim());
-
-	const width = lines.reduce((max, line) => {
-		return Math.max(max, line.length);
-	}, 0);
-
-	const TEMPLATE = [
-		[' _____', '_', '___  '],
-		['(_)   ', ' ', '   ` '],
-		['  |   ', '*', '   | '],
-		['  |___', '_', '___| '],
-		['  (_)_', '_', '____)'],
-	];
-
-	let banner = '';
-
-	TEMPLATE.forEach(([left, middle, right]) => {
-		if (middle === '*') {
-			lines.forEach((line) => {
-				banner +=
-					left +
-					line +
-					' '.repeat(width - line.length) +
-					right +
-					'\n';
-			});
-		}
-		else {
-			banner += left + middle.repeat(width) + right + '\n';
-		}
-	});
-
-	log(banner);
 }
 
 function relative(filePath) {
--- a/projects/npm-tools/packages/changelog-generator/src/index.js
+++ b/projects/npm-tools/packages/changelog-generator/src/index.js
@@ -3,47 +3,17 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-const child_process = require('child_process');
 const fs = require('fs');
 const path = require('path');
 const {promisify} = require('util');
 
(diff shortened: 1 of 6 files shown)


projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -66,6 +71,50 @@ function run(command, ...args) {
 	});
 }
 
+/**
+ * Prompt the user for input.
+ */
+function prompt(question, options) {
```
wincent, line 77: Obvious candidate for extraction into another file, along with the `log` _et al_ functions.

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -749,6 +937,41 @@ async function main(_node, _script, ...args) {
 		}
 	}
 
+	try {
+		await go(options);
+	}
+	finally {
+		if (readline) {
+			readline.close();
+		}
```
wincent, line 946: If we extract `prompt()`, we'll also extract a clean-up function into the same place, to be called from here.

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -749,6 +937,41 @@ async function main(_node, _script, ...args) {
 		}
 	}
 
+	try {
+		await go(options);
```
wincent, line 941: This is an example of a thing to refactor. The name `go` sucks; I just need something so that I could have `--interactive` mode run once as a preview, and then recurse to run again "for realz". I'll see if I can find something cleaner.

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -749,6 +937,41 @@ async function main(_node, _script, ...args) {
...
+ *
+ * - Prints `preview` to standard output (eg. during `--dry-run`); or:
+ * - Prints `contents` to the specified `outfile`.
+ */
+async function write(options, preview, contents) {
+	if (options.dryRun || options.version.endsWith(PLACEHOLDER_VERSION)) {
+		if (options.dryRun) {
+			info(`[--dry-run] Preview of changes 👀`);
+		}
+		else if (options.interactive) {
```
wincent, line 961: These conditionals kind of suck, specifically how `options.interactive` explicitly represents a command-line option but the `endsWith()` check encodes application state (because in interactive mode we'll _usually_ start with a placeholder — although it's not compulsory — and then after we've seen the preview we'll have a non-placeholder version when we run again). Another thing to think of an improvement for... 🤔

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -838,22 +1059,64 @@ async function main(_node, _script, ...args) {
...
-		else {
-			await writeFileAsync(outfile, newContents);
-		}
+		await write(options, contents, newContents);
 
 		written = newContents.length;
 	}
 
-	if (options.dryRun) {
+	if (options.interactive && options.version.endsWith(PLACEHOLDER_VERSION)) {
```
wincent, line 1067: Naughty `PLACEHOLDER_VERSION` again.

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -838,22 +1059,64 @@ async function main(_node, _script, ...args) {
...
+		const answer = await prompt(
+			'Would you like to stage these changes? (y/n)',
+			[]
+		);
+
+		if (/^y(es?)?$/i.test(answer)) {
+			await git('add', '--', options.outfile);
+		}
+
+		// we appear to hang here... why?
```
wincent, line 1117: Whoops. Will blow this away. We were hanging at one point because I wasn't calling `readline.close()`.

wincent:
Well, I didn't a bit of refactoring and cleanup — not as much as I would like, but enough to leave it in a bit better state than it was.

julien (approved):

PR #198, opened 2020-10-19
chore(amd-loader): merge liferay-amd-loader/history into projects/ 

wincent:
Sharing this as a draft to show what is needed to get this integrated, but if it looks good, I'll just push it directly (ie. it already is a merge, so I don't want to merge a merge and make the history any harder to understand than it already is).


wincent:
Ok, semantic pull request error is expected because many of the old imported commits are non-conforming.

Windows failure is probably legit and went undetected previously because we have never run the tests on Windows before; I'll see if I can figure out what is going on there. Everything else is nice and green.

wincent:
Ok, [still failing on Windows](https://github.com/liferay/liferay-frontend-projects/pull/198/checks?check_run_id=1275806727)... I am going to stop guessing and start trying in a VM instead.

wincent:
Screenshot 2020-10-19 -180735-DGmFudpB@2x

Hooray... Windows is green now, so I am going to close this.

PR #200, opened 2020-10-19
chore(amd-loader): prepare to release under @liferay named scope

wincent:
This is going to be one of the simpler projects to move, as far as I can tell, because:

- I already named the directory "amd-loader" instead of "liferay-amd-loader" when I imported the history.
- The autolabeler is already set up to use the "amd-loader" label for that directory.
- The .yarnrc file is already set up to assume the new name.

But note that we must touch two other projects in this commit:

- The preset needs to know about the new name and version.
- The toolkit needs to know for the same reason.
- We update some docs.

One of the nice things about the monorepo is that we can make an atomic commit like this that touches three projects in a single change. I believe all this is safe because the changes to the loader are all safe (ie. backportable), which means that in turn the changes to the toolkit and the preset are safe (because they can be backported too).

### DIFF
--- a/maintenance/projects/js-toolkit/docs/How-to-create-mixed-Java-JavaScript-projects.md
+++ b/maintenance/projects/js-toolkit/docs/How-to-create-mixed-Java-JavaScript-projects.md
@@ -21,7 +21,7 @@ When you are ready to deploy your portlet, edit your `package.json` file and con
 
 The `build` script will be automatically run when you deploy your portlet to Liferay running `gradlew deploy`.
 
-After everything is transpiled (if necessary) to Ecmascript 5+ and CommonJS you must run the [liferay-npm-bundler](../packages/liferay-npm-bundler) tool to pack all JavaScript code (including npm dependencies) and transform it to AMD so that [Liferay AMD Loader](https://github.com/liferay/liferay-amd-loader) may grab it from the server to use it in the browser.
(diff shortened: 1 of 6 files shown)


wincent:
Draft because we should do one last release of the non-scoped package first so that the page on the npm registry gets updated links and we can prove that releasing it from the monorepo works. Waiting to get ownership privs in order to do that.

wincent:
Ok, did the test release of 4.3.1 — this is the last release before we proceed to release under the named scope, just to show that everything is working from the monorepo with all the changes me made as part of the migration.

- [Diff of v4.3.0 and v4.3.1, as downloaded from npm](https://gist.github.com/wincent/27acc9a461ebf91613cebd1ec42ecc4c).
- [Diff `--stat`](https://gist.github.com/wincent/a4d8f56caa325bb07eb4028411eb33d1).

As you can see from the stat, we aren't shipping test files or coverage junk any more, so although the diff is large, when you ignore the thousands of lines of junk that got stripped out and look at what's left, it all looks minimal, as you would expect.

So with that, I don't think we need to bother integrating v4.3.1 itself into DXP in order to validate it — we can instead just proceed with the change in this PR and integrate the new `@liferay/amd-loader` instead (which will also be v4.3.1).

wincent:
Seeing as I am a bit blocked on this, I'm just going to merge it and if I make a huge mess will just have to clean it up later.

cowboy-coding

That is literally how I work.

PR #201, opened 2020-10-19
chore(amd-loader): remove an unnecessary line from the .gitignore

wincent:
Because this one is covered by the top-level one in the root of the monorepo.


wincent:
Just going to merge this because it is trivial.

PR #202, opened 2020-10-19
chore(amd-loader): prepare for releases via js-publish

wincent:
Will use this to make one last non-named-scope release, and then [a named-scope one](https://github.com/liferay/liferay-frontend-projects/pull/200) after that.


wincent:
Not really any way to _fully_ test this without actually cutting a release, so I suggest we just merge and 🙏

PR #203, opened 2020-10-19
chore(js-toolkit): make release process uniform

wincent:
Do all of these the same way.

Note that we are not running `yarn ci` in preversion because we'll often wind up releasing more than one package in a batch, and we expect the maintainer to see a green CI run (and probably a local one too) before doing that. I expect that later on we'll have some Lerna-like automation in place around this (but not Lerna itself, because its dependency footprint is off the charts).


wincent:
With stuff like this you can't really test it without doing a publish, so I'm going to pull the trigger on it. I double checked that the paths are right, so I am pretty confident that at least that aspect is good.

PR #204, opened 2020-10-20
docs(amd-loader): add changelog

wincent:
This project has never had an in-repo changelog, so we generate one from scratch:

    git checkout 17a5ab58
    npx @liferay/changelog-generator --regenerate --minor
    git checkout master
    mv CHANGELOG.md projects/amd-loader
    git add projects/amd-loader/CHANGELOG.md

So, two or three oddities to note here:

- The use of `--minor` is arbirtrary; I just needed to pick a version that was above the previously released one. The changes that have been made since that last release get stripped out by hand and will be included in the next release.

- Given that the changelog generator relies on the `--first-parent` switch, I checked out 17a5ab58 because that is a commit that predates the subtree merge; it means we can look at the history from the perspective of what the project used to look like in the old repo.

- Anomalously, this project basically never used merge commits, so most of the versions don't have any info that can be extracted. Such is life; they will have it going forward.


wincent:
Just an autogenerated doc update, so going to merge this in so I can cut a real release and see it working in the monorepo.

PR #205, opened 2020-10-20
docs(monorepo): make CONTRIBUTING.md docs more consistent

wincent:
- Added one for amd-loader because it didn't have one.
- Recommend the use of the new `--interactive` switch to `@liferay/changelog-generator`.
- Tweak the js-toolkit one to reflect our monorepo-wide commit conventions.

One small thing to note: in the amd-loader document I refer to the project as `@liferay/amd-loader` because we're about to migrate it into the named scope and we may as well skate towards where the puck is going.

### DIFF
--- a/projects/amd-loader/CONTRIBUTING.md
+++ b/projects/amd-loader/CONTRIBUTING.md
@@ -1,0 +1,37 @@
+# Contributing guidelines
+
+## Release process
+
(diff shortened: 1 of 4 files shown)


wincent:
A benign documentation update, so it may as well go in.

PR #206, opened 2020-10-20
docs(monorepo): add example of using version range with `npm deprecate`

wincent:
Seeing as I just had to use it like this, documenting it for posterity.


wincent:
Applying The GIF Rule™ — if you include a GIF, it's cool to merge your own PRs.

smart

PR #209, opened 2020-10-20
fix(changelog-generator): match `yarn version` behavior more closely

wincent:
So I noticed while preparing the 3.0.1 release of the npm-bundler that we had a discrepancy between how we determine the next version number and how Yarn does it.

Namely:

- We were at 3.0.1-pre.1.
- We ran `yarn run liferay-changelog-generator -i` and selected a "patch" release; that bumped the version number to 3.0.2 in the changelog.
- We then ran `yarn version --patch` and that bumped the version number to 3.0.1 in the package.json; you can see both of those changes in 113b7addc894.
- We then fixed the mismatching number by editing the changelog in 54b66db412b8.

So, I went back and ran `yarn version` in a test project in a bunch of combinations and observed the behavior. Added some tests to show what the expected behavior is (the list of combinations is long so I won't repeat it here, but you can see it in the table in the tests).

Test plan: run the tests.

### DIFF
--- a/projects/npm-tools/packages/changelog-generator/bin/liferay-changelog-generator.js
+++ b/projects/npm-tools/packages/changelog-generator/bin/liferay-changelog-generator.js
@@ -5,7 +5,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-const main = require('../src');
+const {main} = require('../src');
 
 main(...process.argv).catch((err) => {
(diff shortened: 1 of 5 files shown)


wincent:
Last call: I just pushed a small refactor commit to make one of the tests look a bit less ugly, but I'll merge this once I see CI green.

wincent:
It's green...

green

PR #210, opened 2020-10-21
test(monorepo): consistently run local tests scoped to project/package

wincent:
The idea is that when you run `yarn ci` from anywhere in the repo we run all the tests in the repo to make sure that everything works well together, but when you are inside a project or a package directory, you can run `yarn test` to run only the tests for that package and get a quicker feedback loop. Running `yarn test` at the top level continues to run everything. Running it at an intermediate level, like say, "projects/npm-tools/" will run it for all the packages nested below that point.


wincent:
adjustments

PR #211, opened 2020-10-21
feat(monorepo): add liferay-workspace-scripts

wincent:
This new private package is to the monorepo what liferay-npm-scripts is to the liferay-portal repo: that is, it presents a simplified interface over the top of some common operations (build, format, lint, test etc) that we can call from all over without having to pass complicated parameter lists.

In terms of prior art, I did [something similar in a monorepo over here](https://github.com/wincent/js/tree/master/packages/workspace-scripts) a while back, and [we talked about doing this in Clay](https://github.com/liferay/clay/issues/1567), but never got around to it.

The solution taken in Clay is to use a Yeoman generator to produce its 40+ packages. This makes creation easy, but not maintenance, because if you want to change what the scripts do, or what scripts there are, you have to touch every single package.json file. Hence, the alternate approach taken here.

Why not just use liferay-npm-scripts, then? Well, that is currently very specifically tailored for use in liferay-portal (apart from some minor use for formatting in this repo and in the liferay-learn one), and there are some kinds of things we want to do in this repo that we'll never want to do in liferay-portal, as far as I know (eg. publishing npm packages). So, the plan then, is to continue investing in liferay-npm-scripts and making it more generally useful, but to have a specialized tool in this repo that calls liferay-npm-scripts whenever appropriate (you can already see this because our `format` task calls liferay-npm-scripts).

Test plan: ran a bunch of commands; I don't know if I managed to run every command from every directory, but I did try to hit a good sample of them... added some `console.log()` to things I can't really run yet for real (like the `publish`) tasks to convince myself that they are probably going to work.


support/packages/workspace-scripts/src/index.js
```diff
@@ -0,0 +1,264 @@
...
+		try {
+			const json = JSON.parse(fs.readFileSync(pkg, 'utf8'));
+
+			if (Array.isArray(json.workspaces)) {
+				return cwd;
+			}
+		}
+		catch {
+
+			// Keep looking.
```
jbalsas, line 100: 👀

jbalsas (approved):
I can't neither confirm nor deny it, but I think this PR doesn't make your mama cry

wincent:
> I can't neither confirm nor deny it, but I think this PR doesn't make your mama cry

I just trimmed off some bytes in https://github.com/liferay/liferay-frontend-projects/pull/211/commits/47b00146680b337bd88ee68eed441bb7f030ae46, which was the most likely thing that was gonna make her cry.

PR #213, opened 2020-10-21
test(monorepo): fill in test gaps

wincent:
Some packages didn't have tests, but in order to make things consistent in the monorepo it is convenient if we just maintain the invariant that _every_ package will have tests.

So this PR adds some tiny stub/scaffolding tests to the packages which didn't have any (some minor refactoring required in a couple of places to make things actually possible to test) and applies some other consistency fixes noted along the way.


jbalsas (approved):
More tests are often better than less tests

wincent:
> More tests are often better than less tests

Very true, as is the corollary:

> Fewer tests are often better than more tests.

😂

judgment-call

PR #214, opened 2020-10-22
chore(npm-scripts)!: drop IE support from Babel config

wincent:
**BREAKING CHANGE:** This commit takes our 7.3 compatibility matrix, and drops IE from the list.

~In order for this to be safely used in pre-7.4 versions, we will have to add `.babelrc.js` config there that overwrites the settings added here. (Well, that's the plan for now at least; if the overriding proves to be too hard, I'll figure out another way.)~

Note that [Babel has always recommended being explicit anyway](https://babeljs.io/docs/en/babel-preset-env):

> Since one of the original goals of preset-env was to help users easily transition from using preset-latest, it behaves similarly when no targets are specified: preset-env will transform all ES2015-ES2020 code to be ES5 compatible.
>
> We don't recommend using preset-env this way because it doesn't take advantage of its ability to target specific environments/versions.

And also:

> `preset-env`'s behavior is different than browserslist: it does _not_ use the `defaults` query when there are no targets are found in your Babel _or_ browserslist config(s). If you want to use the defaults query, you will need to explicitly pass it as a target:
>
> ```
> {
>   "presets": [["@babel/preset-env", {"targets": "defaults"}]]
> }
> ```

This is breaking because, even in 7.4, there will be places where we want to override, such as analytics-client-js, because that is code that happens to live in the liferay-portal repo but is not distributed with DXP (rather, it is published separately via CDN, and we'll want to maintain compatibility with IE there).

[The compatibility matrix is currently at this link](https://www.liferay.com/documents/10182/246659966/Liferay+DXP+7.3+Compatibility+Matrix.pdf/30fb6a1e-431d-7c84-e211-f2a0590013bf) (not sure if it is a permanent link or not, due to all the junk in the URL). In any case, you can easily find it if you Google for ("Liferay DXP compatibility matrix").

It says:

-   Desktop Browser:
    -   Chrome 65
    -   Chrome (Latest)
    -   Edge (Latest)
    -   Firefox (Latest)
    -   Firefox ESR (Latest)
    -   Internet Explorer 11 (Limited)
    -   Safari 12
    -   Safari 13
-   Mobile Browser:
    -   Chrome (Latest)
    -   Safari (Latest)

Note that I am asking for "Safari >= 12" instead of "Safari 12, Safari 13", because Safari [is already up to v14](https://en.wikipedia.org/wiki/Safari_version_history) and it seems absurd to ignore it.

Likewise, note from [the Browserslist docs](https://github.com/browserslist/browserslist) that mobile Safari translates to "iOS", and mobile Chrome translates to "ChromeAndroid" (Chrome on iOS is actually just a WebKit wrapper, so is covered by "iOS").

**Test plan:**

Run `npx browserslist 'Chrome 65, Firefox ESR, Safari >= 12, last 1 Chrome version, last 1 ChromeAndroid version, last 1 Edge version, last 1 Firefox version, last 1 iOS version'` and see:

    and_chr 85
    chrome 86
    chrome 65
    edge 86
    firefox 81
    firefox 78
    ios_saf 14
    safari 14
    safari 13.1
    safari 13
    safari 12.1
    safari 12

Deploy this in liferay-portal and check the generated output looks as you would expect (eg. it has arrow functions in it).

See comments further down for additional tests that I ran as I added commits.


wincent:
Draft because I haven't actually tested it yet, and I am anticipating that as soon as I do I will discover how the way we merge Babel configs chokes on this. 🤞

jbalsas:
I'm not sure where it is, but we used to keep a different ruleset for JSPs since that was not safe. Dropping IE11 means we could drop some of those as well. Just bringing this up since I'm not sure now where that lives.

wincent:
> I'm not sure where it is, but we used to keep a different ruleset for JSPs since that was not safe. Dropping IE11 means we could drop some of those as well. Just bringing this up since I'm not sure now where that lives.

Yeah, I know where that is, and had it on my mental todo-list. Given that we don't transform JS in JSPs, it is not a Babel config; it's the `lintJSP()` function, which you can see in [these tests](https://github.com/liferay/liferay-frontend-projects/blob/6cf3084bf4d6b6bace79ba91976c7b7b88303cd5/projects/npm-tools/packages/npm-scripts/test/jsp/lintJSP.js#L137-L171).

jbalsas:
> Given that we don't transform JS in JSPs, it is not a Babel config; it's the lintJSP() function, which you can see in these tests.

That's why I couldn't find it... forgot that was there 🙈

wincent:
**Test results:**

Here is a tiny snippet from `build/node/packageRunBuild/resources/liferay/throttle.es.js` in `frontend-js-web`, showing that we have arrow functions in all their sweet, sweet purity in the build output:

```
        const schedule = () => {
          timeout = setTimeout(() => {
            timeout = null;
          }, interval);
          last = now;
          fn.apply(context, args);
        };
```

And here is one from `build/node/packageRunBuild/resources/liferay/global.bundle.js` (ie. something that went through webpack):

```
forEach(e=>{let r;r=n.has(e)?e:"".concat(u).concat(e),s.set(r,t[e])})
```

Finally, to see that we can actually override the `targets` config when we want, I added a `.babelrc.js` in `analytics-client-js`:

```js
module.exports = {
	presets: [
		[
			'@babel/preset-env',
			{
				targets: 'defaults',
			},
		],
	],
};
```

Before overriding, `build/analytics-all-min.js` contains arrows like:

```
()=>document.removeEventListener("DOMContentLoaded",e)
```

After:

```
function(){return document.removeEventListener("DOMContentLoaded",e)}
```

And searching for `=>` in the file produces no results.

So that's all good, I think. Going to look at the JSP side now, which might be a bit trickier, because there is no config file involved.

wincent:
So, about forking the settings for Babel between `master` and `7.3.x` (etc), it seems the least horrible way (to avoid having to create a bunch of `.babelrc.js` files scattered everywhere) is to do auto-detection based on `release.info.version` string in the `release.properties` file.

Would you 🤮 if I did that, @jbalsas? I could also leverage that to toggle the linting behavior in JSPs without having to deal with config files (which don't exist for JSP at the moment).

The alternatives all have too much indirection for my taste (eg. a new preset you could opt-in to from your `npmscripts.config.js` on the other branches, which in turn would select a new Babel config etc).

jbalsas:
> Would you 🤮 if I did that, @jbalsas? I could also leverage that to toggle the linting behavior in JSPs without having to deal with config files (which don't exist for JSP at the moment).

I wouldn't find that too horrible, no

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
```diff
@@ -96,20 +97,30 @@ function getMergedConfig(type, property) {
...
+					[baseConfig, getUserConfig('babel')],
+					deepMerge.MODE.BABEL
+				);
+
+				// (Temporary) special case required by:
+				//
+				// https://github.com/liferay/liferay-npm-tools/issues/303
+				//
+				// TODO: Remove once incremental-dom is no longer used in
+				// liferay-portal.
```
wincent, line 120: Seems like something we will be able to skip, at least for 7.4 and above, real soon now...

matuzalemsteles:
hey @wincent can we know how much transformation code we can eliminate from the files with this? even if it's 1% it's a victory 😂.

wincent:
> hey @wincent can we know how much transformation code we can eliminate from the files with this? even if it's 1% it's a victory 😂.

Excellent question; let me do a before-and-after comparison on a couple of modules to see the difference... I'll report back in a little while.

jbalsas:
It's probably not as much as we'd like. @bryceosterhaus looked into it before wrapping `7.3` and couldn't find big immediate improvements.

If I remember correctly, `babel` still adds quite a lot of boilerplate for utils like `_classCallCheck`, `_defineProperty`, `ownKeys`, `_objectSpread` or others that are probably duplicated across all of our files, but getting babel to stop outputting that didn't seem trivial.

wincent:
In `frontend-js-web` we see about 1,305 lines stripped, and 69,752 characters. Obviously, after gzip, the difference will be less. Totally unscientific results:

**Before:**

```
$ cloc build/node/packageRunBuild/resources

-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
JavaScript                     139           1589           2345           8142

$ wc build/node/packageRunBuild/**/*.js | tail -1
   12005   44520  580077 total
```

**After:**

```
$ cloc build/node/packageRunBuild/resources

-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
JavaScript                     139           1407           2529           6836

$ wc build/node/packageRunBuild/**/*.js | tail -1
   10701   39260  510325 total
```

wincent:
Like, I'm not sure why the "after" result has more lines of comments. Just to sanity check, here is one specific file being compared:

- [`side_navigation.es.js` before](https://gist.github.com/wincent/be6203a1e3032251a8974bcc633d3751)
- [`side_navigation.es.js` after](https://gist.github.com/wincent/b2b073003ce91b9256e1b86bbb1b8ac2)
- [Diff view](https://gist.github.com/wincent/408131bf38c070b10574e976930ad0aa)

matuzalemsteles:
> It's probably not as much as we'd like. @bryceosterhaus looked into it before wrapping 7.3 and couldn't find big immediate improvements.

Yeah, I had forgotten about that.

> In frontend-js-web we see about 1,305 lines stripped, and 69,752 characters. Obviously, after gzip, the difference will be less. Totally unscientific results:

Apparently good results, approximately 16% of lines removed in this case 🙂.

> Like, I'm not sure why the "after" result has more lines of comments. Just to sanity check[..]

Hmm this is very contradictory but looking more closely, it seems that it removes a lot from the beginning but later some lines end up being added in cases like this:

```diff
-      setClasses(element, _defineProperty({}, className, true));
+      setClasses(element, {
+        [className]: true
+      });
```

Apparently, this ends up making up for the number of lines throughout the file but maybe we can see a larger number of characters being removed.

---

Just a few ideas that came up now, I'm not sure if it would be a good metric but it would be really cool if we managed to monitor the overall size of the js files to identify the size of our codebase and find out each year how much it increased or decreased, maybe something to have closer. This can help us perhaps at times like this to find out how much code on the exit of the babel we are removing. I think it will be easier in the future to find better numbers for a "Performance Budget".

wincent:
> Hmm this is very contradictory but looking more closely, it seems that it removes a lot from the beginning but later some lines end up being added

So we can basically ignore the wrapping changes because in production all this gets minified anyway (both space and `\n` are one byte, and indenting gets stripped out); just look at the byte counts instead.

> Just a few ideas that came up now, I'm not sure if it would be a good metric but it would be really cool if we managed to monitor the overall size of the js files to identify the size of our codebase and find out each year how much it increased or decreased, maybe something to have closer. This can help us perhaps at times like this to find out how much code on the exit of the babel we are removing. I think it will be easier in the future to find better numbers for a "Performance Budget".

Yep, basically the analog of what @bryceosterhaus did for Clay in [the GitHub action](https://github.com/liferay/clay/blob/aefb430bd83de2056122ae2af3fb06cae0b2f6ae/.github/workflows/main.yml#L34-L76).

jbalsas:
> > [...]monitor the overall size of the js files to identify the size of our codebase and find out each year how much it increased or decreased, maybe something to have closer. This can help us perhaps at times like this to find out how much code on the exit of the babel we are removing. I think it will be easier in the future to find better numbers for a "Performance Budget".
>
> Yep, basically the analog of what @bryceosterhaus did for Clay in the GitHub action.

One key difference is that we never ship all of our codebase at once in DXP, so codebase size is likely not a valid metric in this case. 

I think the more obvious approach would be to pick specific scenarios, take a baseline and then compare as time goes by

wincent:
> I think the more obvious approach would be to pick specific scenarios, take a baseline and then compare as time goes by.

Obviously dovetails into whatever performance analysis infra we eventually roll out, [maybe](https://github.com/liferay/liferay-npm-tools/pull/396).

wincent:
BTW, seeing as nobody is commenting on the code, I am going to assume I an merge it. 😀 

Please contact customer support if you see any issues that need to be resolved.

jbalsas:
> BTW, seeing as nobody is commenting on the code, I am going to assume I an merge it. 😀

I was just commenting, but I'm fine shipping this as is :)

My only thought was wether we could simply set some process env variable with the version to simplify all that parsing thing... but let's get this in and then see if it can be improved further!

wincent:
> My only thought was wether we could simply set some process env variable with the version to simplify all that parsing thing... but let's get this in and then see if it can be improved further!

Doesn't sound simpler to me (we'd have to touch the Java side, right?).

jbalsas:
> Doesn't sound simpler to me (we'd have to touch the Java side, right?).

Yeah, I was wondering if it was already happening and there was something in the environment we could use

wincent:
> Yeah, I was wondering if it was already happening and there was something in the environment we could use

Problem with relying on that would be that you would get different results when run from Gradle (ie. environment variables present) from `yarn build` (etc) on the commandline.

jbalsas:
true, true

wincent:
- Release notes: https://github.com/liferay/liferay-frontend-projects/releases/tag/npm-scripts%2Fv34.0.0
- DXP PR is here: https://github.com/liferay-frontend/liferay-portal/pull/441

projects/npm-tools/packages/npm-scripts/src/config/babel.json
```diff
@@ -4,7 +4,15 @@
 			"plugins": ["transform-react-remove-prop-types"]
 		}
 	},
-	"presets": ["@babel/preset-env", "@babel/preset-react"],
+	"presets": [
+		[
+			"@babel/preset-env",
+			{
+				"targets": "Chrome 65, Firefox ESR, Safari >= 12, last 1 Chrome version, last 1 ChromeAndroid version, last 1 Edge version, last 1 Firefox version, last 1 iOS version"
```
bryceosterhaus, line 11: Not super important, but for readability I think we can make this something like 

```js
targets: {
    "chrome": "65",
    "safari": ">= 12",
    // ...
}
```

projects/npm-tools/packages/npm-scripts/src/utils/getDXPVersion.js
```diff
@@ -0,0 +1,48 @@
...
+const fs = require('fs');
+const path = require('path');
+
+const findRoot = require('./findRoot');
+
+/**
+ * Attempts to infer the current DXP version based on the contents of the
+ * `release.properties` file in the repo root directory.
+ */
+function getDXPVersion() {
```
bryceosterhaus, line 15: I hadn't thought of doing something like this and checking against dxp itself. Great idea!

projects/npm-tools/packages/npm-scripts/src/utils/getDXPVersion.js
```diff
@@ -0,0 +1,48 @@
...
+				return {
+					major: parseInt(major, 10),
+					minor: parseInt(minor, 10),
+					patch: parseInt(patch, 10),
+				};
+			}
+		}
+		catch {
+
+			// Oh well, it was worth a shot...
```
bryceosterhaus, line 42: Is it worth adding a log here for debug purposes? I'm sure we won't actually run into it, but just in case.

bryceosterhaus (commented):
Nothing super important, but thought I would at least note

projects/npm-tools/packages/npm-scripts/src/config/babel.json
wincent: AFAIK, that's not supported. [The docs](https://babeljs.io/docs/en/babel-preset-env) say that when you use key-value format you can supply:

> an object of minimum environment versions to support

and they give this example:

```
{
  "targets": {
    "chrome": "58",
    "ie": "11"
  }
}
```

which they contrast with the alternative (that I've used here) of:

> a browserslist-compatible query:
>
> ```
> {
>   "targets": "> 0.25%, not dead"
> }
> ```

So my reading of that is that the query-syntax can't be used when using object keys, only a concrete minimum version can.

projects/npm-tools/packages/npm-scripts/src/utils/getDXPVersion.js
wincent: Yeah, my thinking on this one is:

- This should never happen.
- If it does happen, anything we log will get lost in the flood of normal build output.
- If things do go wrong, the "graceful degrade" here is that we just use our latest Babel config, which isn't a bad thing because it means that `master` will behave exactly as it did before. (Presumably, breakage on older branches will be caught by one of our other safety nets, such as manual QA — the thing about breaking IE with arrow functions is that if you fail to transpile them across all of portal, you will notice pretty much immediately as soon as you try anything at all in IE.)

projects/npm-tools/packages/npm-scripts/src/config/babel.json
bryceosterhaus: Strange. I remember seeing an example of using key-value pairs with similar syntax, I can't seem to find where I saw that at though... I may just be delusional 🤷‍♂️

wincent: Just to rule this out as a possibility, I tried reforming it as:

```
"targets": {
	"Chrome": "65, last 1 version",
	"ChromeAndroid": "last 1 version",
	"Edge": "last 1 version",
	"Firefox ESR": "last 1 version",
	"Firefox": "last 1 version",
	"Safari": ">= 12",
	"iOS": "last iOS version"
}
```

But that gave errors like these:

```
Error: [BABEL] Invalid Option: 'Chrome' is not a valid target
        Maybe you meant to use 'chrome'?
Error: [BABEL] 'ChromeAndroid' is not a valid target
        Maybe you meant to use 'android'?
Error: [BABEL] Invalid Option: 'Edge' is not a valid target
        Maybe you meant to use 'edge'?
Error: [BABEL] Invalid Option: 'Firefox ESR' is not a valid target
        Maybe you meant to use 'firefox'?
Error: [BABEL] Invalid Option: 'Safari' is not a valid target
        Maybe you meant to use 'safari'?
Error: [BABEL] Invalid Option: 'iOS' is not a valid target
        Maybe you meant to use 'ie'?
```

So massaged it a bit further to this (note that for this experiment I dropped "iOS" beacuse I don't know what to put for that):

```
"targets": {
	"chrome": "65, last 1 version",
	"android": "last 1 version",
	"edge": "last 1 version",
	"firefox": "ESR, last 1 version",
	"safari": ">= 12"
}
```

But that yielded:

```
Error: [BABEL] Invalid Option: 'last 1 version' is not a valid value for 'targets.android'.
```

It looks like you can't use the query syntax in this way. My guess is that they are using `parseInt` internally to identify the minimum version, which is why `"65, last 1 version"` works (`65`) but `"last 1 version"` does not (`NaN`).

Issue #217, opened 2020-10-29
Error when calling @liferay/npm-bundler with npx


p2kmgcl:
### Issue type

-   :bug: Bug report

### Description

(Linked Jira issue https://issues.liferay.com/browse/LPS-122795)
(Reviewed with @izaera)

**Desired behavior:** call `@liferay/npm-bundler` with `npx` (`npx @liferay/npm-bundler`).

**Current behavior:** when the command is run, an error is shown in the console: 
![image](https://user-images.githubusercontent.com/800645/97543749-f20ca280-19c8-11eb-8547-69134b440d76.png)

**Repro instructions (if applicable):**

**Other information (environment, versions etc):**
- Tested on NodeJS 10 on Windows and NodeJS 10 on Linux.
- I checked the same process with `npx @angular/cli` and it works.
- It _might_ be the reference to the binary file in package.json ([bundler](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/js-toolkit/packages/npm-bundler/package.json#L4) vs [angular](https://github.com/angular/angular-cli/blob/master/package.json#L11))

wincent:
> * It _might_ be the reference to the binary file in package.json ([bundler](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/js-toolkit/packages/npm-bundler/package.json#L4) vs [angular](https://github.com/angular/angular-cli/blob/master/package.json#L11))

Seems like a reasonable guess. In `@liferay/npm-scripts` we [use the dot-relative form](https://github.com/liferay/liferay-frontend-projects/blob/993b70b9fc23165b023b531c6c00118ba9189d2f/projects/npm-tools/packages/npm-scripts/package.json#L4) and it works fine.

wincent:
Just to clarify, @p2kmgcl — I see you tested this on Linux and Windows, but the linked LPS only mentions Windows. Does that mean it works on Linux and not on Windows?

p2kmgcl:
Nope, I tested on Windows for the LPS and then I found that it doesn't work on Linux too (thanks to @izaera). I am going to update the Jira ticket :sweat_smile:

izaera:
I'm getting a different error in Windows:

![image](https://user-images.githubusercontent.com/3273630/97574544-ac150600-19eb-11eb-8c3d-fbbd40be1acb.png)

[...]

PR #218, opened 2020-10-29
fix: make binaries work from npx

izaera:
Right now, `npx @liferay/npm-bundler` fails because it cannot find the executable `.js` script.

We have made a very deep, intelligent, exhaustive, and thorough investigation (basically just peeking [how `@angular/cli` declares its binary](https://github.com/angular/angular-cli/blob/4169e303c3075df3bdb8c7f500a2d034d6532537/packages/angular/cli/package.json#L7)) and have come to the conclusion that adding `./` to the left and maybe removing the `.js` may help.

Unfortunately we need to merge this PR and release a new version to test this :-(.

### DIFF
--- a/projects/js-toolkit/packages/js-toolkit-scripts/package.json
+++ b/projects/js-toolkit/packages/js-toolkit-scripts/package.json
@@ -1,7 +1,7 @@
 {
 	"author": "Liferay Frontend Infrastructure Team <pt-frontend-infrastructure@liferay.com>",
 	"bin": {
-		"js-toolkit": "./bin/js-toolkit"
+		"js-toolkit": "./bin/js-toolkit.js"
 	},
 	"dependencies": {
 		"@babel/core": "^7.0.0",
--- a/projects/js-toolkit/packages/js-toolkit-scripts/package.json
+++ b/projects/js-toolkit/packages/js-toolkit-scripts/package.json
@@ -1,7 +1,7 @@
 {
 	"author": "Liferay Frontend Infrastructure Team <pt-frontend-infrastructure@liferay.com>",
 	"bin": {
-		"js-toolkit": "bin/js-toolkit.js"
+		"js-toolkit": "./bin/js-toolkit"
 	},
 	"dependencies": {
 		"@babel/core": "^7.0.0",
--- a/projects/js-toolkit/packages/npm-bridge-generator/package.json
+++ b/projects/js-toolkit/packages/npm-bridge-generator/package.json
@@ -1,7 +1,7 @@
 {
 	"author": "Liferay Frontend Infrastructure Team <pt-frontend-infrastructure@liferay.com>",
 	"bin": {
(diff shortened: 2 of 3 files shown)


izaera:
I'm going to merge this since we don't have any other option to make sure it is correct.

izaera:
According to [npx's documentation](https://www.npmjs.com/package/npx) the problem may be more in the name of the executable than on the path itself. Going to try to investigate it locally...

wincent:
> According to [npx's documentation](https://www.npmjs.com/package/npx) the problem may be more in the name of the executable than on the path itself. Going to try to investigate it locally...

It works fine with `npx @liferay/npm-scripts` though, that has exactly the same naming/structure, except `./bin` instead of `bin`.

wincent:
And it [uses the `.js` extension](https://github.com/liferay/liferay-frontend-projects/blob/993b70b9fc23165b023b531c6c00118ba9189d2f/projects/npm-tools/packages/npm-scripts/package.json#L4), so I am not sure it will work without it (it might, but I have no past experience with omitting the extension).

izaera:
I think the extension makes no difference in the end, because Windows' node creates a `.bat`|`.cmd` file under the hood to be able to execute things so I think it takes care of the extension for us...

However, I'm not sure now if this is the real error (see [this comment](https://github.com/liferay/liferay-frontend-projects/issues/217#issuecomment-718724923)).

wincent:
Can you leave the extension the way it was then? That, at least, is useful for users of IDEs who can hit a binding when their cursor is on the filename in order to jump to that file.

Even if this doesn't fix that other error, seems like the change from `bin` to `./bin` should still go in.

wincent:
Sounds good.

izaera:
Seeing that ["my" error](https://github.com/liferay/liferay-frontend-projects/issues/217#issuecomment-718724923) only happens when `npx` is used without any `node_modules` at all and, as that situation doesn't make sense for the tool and seems complex to fix, I won't be fixing that.

Instead, @p2kmgcl will [workaround it in Fragments CLI](https://github.com/liferay/generator-liferay-fragments/pull/176). In case that doesn't work, we will change the bundler to let users directly invoke its [main method](https://github.com/liferay/liferay-frontend-projects/blob/993b70b9fc23165b023b531c6c00118ba9189d2f/projects/js-toolkit/packages/npm-bundler/src/index.ts#L19) programmatically. That should be easier to hack and more useful for everybody.

So I guess we can merge this if you agree @wincent .

wincent:
> So I guess we can merge this if you agree @wincent .

I do, because:

> Even if this doesn't fix that other error, seems like the change from `bin` to `./bin` should still go in.

wincent (approved):
Don't know if the title should be edited before merging. Seeing as it doesn't seem to actually fix the problem. Maybe it should be "refactor(js-toolkit): make `bin` properties consistent".

izaera:
OK, we shouldn't merge this then, because the commit messages are also misleading. I will resend it.

PR #219, opened 2020-10-29
feat(js-toolkit): convert package.json version classfier to OSGi version classifer

aschaeffer:
Hello there :-)

This small PR solves a problem with version numbers having a qualifier.

### The problem

None of these version numbers work:

```json
{
  "version": "7.12.0-clavis-17"
}
```

```json
{
  "version": "7.12.0.clavis-17"
}
```

### The errors

***`7.12.0-clavis-17`***

1. `yarn run build` works
2. The jar gets created
3. The `MANIFEST.MF` contains the `Bundle-Version: 7.12.0-clavis-17`
4. OSGi-Deployment fails because of an OSGi-incompatible version number

```
INFO  [com.liferay.portal.kernel.deploy.auto.AutoDeployScanner][AutoDeployDir:271] Processing xxx-7.12.0-clavis-17.jar
[...]
java.lang.IllegalArgumentException: invalid version "7.12.0-clavis-17": non-numeric "0-clavis-17"
[...]
```

***`7.12.0.clavis-17`***

1. Building fails because of an npm-incompatible version number

```bash
$ yarn run build
[...]
Error: Invalid version: "7.12.0.clavis-17"
    at Object.fixVersionField (/xxx/node_modules/normalize-package-data/lib/fixer.js:191:13)
    at /xxx/node_modules/normalize-package-data/lib/normalize.js:32:38
[...]
```

### Valid version ranges

It's possible to define a version qualifier in a `package.json` and in the `MANIFEST.MF` but the accepted format differs:

| *Version number in ...* | *package.json* | *OSGi MANIFEST.MF* |
| --- | --- | --- |
| 1.2.3 | valid | valid |
| 1.2.3-classifier | valid | invalid |
| 1.2.3-classifier-1 | valid | invalid |
| 1.2.3.classifier | invalid | valid |
| 1.2.3.classifier-1 | invalid | valid |

### The solution

The first occurence of an `-` in the version number will be converted to a `.`

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/__tests__/manifest.test.js
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/__tests__/manifest.test.js
@@ -45,6 +45,34 @@ Tool: liferay-npm-bundler-${version}
 	);
 });
 
+it('includes Bundle-Version with valid simple classifier when provided', () => {
+	const manifest = new Manifest();
+
+	manifest.bundleVersion = '1.0.0-classifier';
+
+	expect(manifest.content).toEqual(
+		`Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Version: 1.0.0.classifier
+Tool: liferay-npm-bundler-${version}
+`
+	);
+});
+
+it('includes Bundle-Version with valid extended classifier when provided', () => {
+	const manifest = new Manifest();
+
+	manifest.bundleVersion = '1.0.0-classifier-1-2-3';
+
+	expect(manifest.content).toEqual(
+		`Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Version: 1.0.0.classifier-1-2-3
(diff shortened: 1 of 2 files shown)


wincent:
@aschaeffer: Can you please run `yarn format` on this to satisfy CI?

I'm asking @izaera to look at this.

izaera:
We had some issues with `bnd.bnd` and `package.json` version numbers before, mostly because, as @aschaeffer shows in his table, OSGi and npm version number formats are incompatible. However, they were related to the server side and the build process.

This looks like a mostly safe fix to me. There's of course the possibility of some version numbers colliding if the developers use `-` and then `.` for different versions of the same package, but it seems so implausible that I wouldn't worry about it.

So I guess we can merge it :+1: (once the `yarn format` issue is fixed).

Thx for the contribution @aschaeffer :-).

wincent:
@izaera: Would this fix be applicable to 3.x as well?

izaera:
> @wincent : Would this fix be applicable to 3.x as well?

Yes. I will send a PR for 3.x too :+1:

PR #234, opened 2020-11-11
feat(npm-scripts): add jest mock for createRange and createContextualFragment

bryceosterhaus:
This PR has a handful of failing unit tests due to createRange not existing in js-dom. Rather than add a setupfile for each module to include this, I figured its best to have as a global mock.

https://github.com/liferay-frontend/liferay-portal/pull/468

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/jest/setup.js
+++ b/projects/npm-tools/packages/npm-scripts/src/jest/setup.js
@@ -10,3 +10,22 @@ global.Liferay = require('./mocks/Liferay');
 global.fetch = require('jest-fetch-mock');
 
 global.themeDisplay = global.Liferay.ThemeDisplay;
+
+if (!global.createRange) {
+	global.createRange = () => ({
+		createContextualFragment(htmlString) {
+			const tempDiv = document.createElement('div'); // eslint-disable-line no-undef
+
+			tempDiv.innerHTML = `<br>${htmlString}`;
+			tempDiv.removeChild(tempDiv.firstChild);
+
+			const fragment = document.createDocumentFragment(); // eslint-disable-line no-undef
+
+			while (tempDiv.firstChild) {
+				fragment.appendChild(tempDiv.firstChild);
+			}
+
+			return fragment;
+		},
+	});
+}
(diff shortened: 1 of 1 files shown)


wincent (approved):
lg2m. I'm going to apply some small tweaks on top and merge as soon as I see it green.

wincent:
So I didn't test this but I assume you did @bryceosterhaus 😂 — there are any number of different ways `createRange` could be mocked, depending on the context ([example here](https://stackoverflow.com/a/57874167/2103996)), so let's just assume that this is going to be good enough to meet our use cases until we do the update, and if it isn't, then we'll tweak it. It's green now, so I'm going to pull the trigger.

If you want to try cutting a release, [the CONTRIBUTING.md](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/CONTRIBUTING.md) should be up-to-date. Things have probably changed a bit since the last time you did it, but not too much. It's basically going to boil down to:

```
cd projects/npm-tools/packages/npm-scripts

# Preview the changelog changes, to decide on release type (major, minor etc),
# and stage them:
yarn run liferay-changelog-generator --interactive

# Update the version number using the same release type as decided in previous step:
yarn version --minor
```

bryceosterhaus:
@wincent are there any docs for updating npm-scripts in portal?

wincent:
> @wincent are there any docs for updating npm-scripts in portal?

No because it is pretty straight forward; from `modules/`:

1. `yarn add -W --dev @liferay/npm-scripts@x.y.z`
2. Make sure lockfile has no unwanted duplication.
3. Test/sanity check according to the changes in the release.

I can add this to the contributing doc though. The only tricky part is that sometimes Yarn requires a bit of coaxing to deduplicate the lockfile, as [noted here](https://github.com/liferay/liferay-portal/commit/d5e9d058e483ac733d55c538c74e85c6009a4ce6):

> Yarn is a bit fickle about these things, so there are four ways you can do this and not all of them always work.
>
> - `npx yarn-deduplicate yarn.lock` (didn't work here)
> - `yarn install` (worked for this case)
> - `yarn install --force` (not needed)
> - `rm yarn.lock && yarn install` (last resort, very aggressive)

I usually start at the first and go down that list until I get the result I want. It used to be that step 1 was always enough; but lately I've found I have to do 2 quite often as well. I might heav had to do 3 once, and I've never done 4.

wincent:
Docs added [here](https://github.com/liferay/liferay-frontend-projects/pull/239).

PR #233, opened 2020-11-11
feat(npm-scripts): add minification to build process

wincent:
~This is a far-from-finished WIP, but parking it here anyway for safe-keeping:~ (upgrading this from WIP to POC)

Final status:

-    [Discussed in Slack here](https://liferay.slack.com/archives/C04D8HMEF/p1606233408059000):
    -   The JS-in-JSP minification works if you run `yarn build` by hand, but when run as part of a `gradle` invocation the jobs happen in the wrong order for our purposes — we can't minify the JS in JSP in `classes/` because it only gets put there _after_ `packageRunBuild`. The conclusion of the Slack thread is that we can leave the code in here as a kind of benign no-op, and if we decide to prioritize the necessary changes on the Gradle side it will "magically" start working. If it's still there months from now and it looks like we're never going to get to it we can just rip it out.
    
        Excerpts from the Slack thread:
    
        > its possible
        >
        > we would need to change the src path to the new output files first when we compile
        >
        > and also replace the jsps used to build the jars
    
        and:
    
        > in terms of whether or not it's worth doing. my quick-and-dirty metrics show that for typical JS we see a reduction of about 42.7% (in pre-gzip byte count), and in terms of how much JS-in-JSP we have on the table, at the moment it is about this much under modules/:
        >
        > - Files: 890
        > - Tags: 1102 (<script> or <aui:script>)
        > - Lines: 43634 (inside tags)
        > - Bytes: 1237555 (inside tags)
        >
        > so, at the moment with runtime minification done by the Java JS minifier, we get a fair bit of compression for that stuff, but once LPS-122883 is done the expectation is that we turn runtime minification off by default and do it all ahead of time as I'm describing above.

-   Also discussed release [here](https://liferay.slack.com/archives/G3JBR21HA/p1606290393489100). I still haven't got source maps working, but so as to keep this rolling forward I am going to proceed with the merge and fast-follow with a fix as soon as I find it. Otherwise this could drag on for days or more.

---

Here's the gist. This implements the first couple of parts of the idea described [in the research doc](https://docs.google.com/document/d/1JYSSSVFQlEwy1CRv1ZD6hfql8AzuXTS39qY5HMnvUe4/edit).

[...]

PR #221, opened 2020-10-30
feat(npm-scripts): fix storybook

oliv-yu:
Just some small fixes:
- update storybook's node path
- rename `index.es.js` to `index.js`
- update using `http-proxy-middleware` https://stackoverflow.com/questions/58431098/create-react-app-http-proxy-middleware-not-working

@thektan

PR #236, opened 2020-11-13
feat: support for webpack 5 in npm-scripts/bundler

izaera:
This is a PR to implement [IFI-2123](https://issues.liferay.com/browse/IFI-2123).

It has been tested by locally releasing (to a Verdaccio installation) versions of npm-bundler and npm-scripts and running an `ant all` in `liferay-portal`. Then started the portal and checked manually that the home page and control panel was in a correct state.

I have had to migrate `amd-loader` to webpack 5 but haven't tested the result. All I know is that the build works and passes the test. I don't think this will be a source of problems, but if it is, I don't want to tackle it now given the very low probability that we have to release a new version of the loader (because we will most probably get rid of it thanks to webpack 5 federation).


izaera:
Oops, the `yarn.lock` file is broken because of the local npm repo :-S. I will fix it and remove the draft marker.

projects/npm-tools/packages/npm-scripts/src/index.js
```diff
@@ -30,7 +30,10 @@ module.exports = async function () {
 		},
 
 		storybook() {
-			require('./scripts/storybook')();
+
+			// Storybook is temporarily disabled until it supports webpack 5
```
jbalsas, line 34: Are we using it in DXP currently?

wincent: I think in two places.

We [discussed the trade-off in Slack](https://liferay.slack.com/archives/DFUBFSDGT/p1605183277123500). My thought is that it's an actively maintained project, so we can be optimistic that this is a temporary measure — they will upgrade to webpack 5 (and by the way, we should check their issue tracker to add our `:+1:` and track the status).

Two observations though:

- We should log something to the console here with a link to where all this is fully explained.
- (Also as discussed on Slack) If we get pushback for dropping support, there are workarouds, like doing a separate build that bundles webpack v4 "inside" without adding it to the Yarn workspace, effectively hiding the dependency. It would be best if we could avoid that complexity/hassle though.

jbalsas: okay_gosling

izaera: Not sure about the console thing. I mean, I like the idea, but with the current portal's build nobody will see it. 

But it's OK, I'll add the log linking to this comment. It's better than nothing I guess :-). I will also notify the team using storybook about this. 

Nevertheless, keep in mind that we are not releasing anything yet, so this won't make effect in portal until we release a new version of `npm-tools`...

wincent: > Not sure about the console thing. I mean, I like the idea, but with the current portal's build nobody will see it.

The team that uses storybook runs it by hand, so they will see it (it is not part of the build process). It will be very helpful to them to understand why their stuff has suddenly stopped working.

Proof that they care about it: [this PR from 13 days ago unbreaking it](https://github.com/liferay/liferay-frontend-projects/pull/221).

> Nevertheless, keep in mind that we are not releasing anything yet, so this won't make effect in portal until we release a new version of `npm-tools`...

Obviously, no effect until you take it out of draft and merge it. Note that @bryceosterhaus is [probably going to do a release soon](https://github.com/liferay/liferay-frontend-projects/pull/234).

izaera: Perfect then :+1:

izaera:
So, this is ready for review and merge (if we like it).

izaera:
Given that [LPS-123144](https://issues.liferay.com/browse/LPS-123144) requires to use this in the portal's build. I will release a new version of `npm-scripts` once we merge, if nobody is against it.

projects/amd-loader/webpack.config.js
```diff
@@ -36,12 +36,11 @@ module.exports = ({flavor}) => {
 		},
 		min: {
 			mode: 'production',
+			optimization: {
+				minimize: true,
+				minimizer: [new TerserPlugin()],
+			},
```
izaera, line 42: We use `terser` instead of `uglify` because, given that it is used for minification, it should be equivalent. There's however a way to [configure the terser plugin to use uglify](https://webpack.js.org/plugins/terser-webpack-plugin/#custom-minify-function), just in case it is needed.

wincent: FWIW, uglify [is dead](https://github.com/mishoo/UglifyJS/issues/3156#issuecomment-392943058) (we'll be [using Terser for our own build-time minification](https://github.com/liferay/liferay-frontend-projects/pull/233)).

projects/js-toolkit/packages/npm-bundler/src/adapt/bundler-project/index.ts
```diff
@@ -74,7 +74,7 @@ async function transformBundles(): Promise<void> {
 		const sourceFile = bundlerWebpackDir.join(fileName);
 
 		if (!fs.existsSync(sourceFile.asNative)) {
-			break;
+			continue;
```
wincent, line 77: 😱

izaera: :shrug: We are using it in so many little builds that it didn't cause any error.

PR #241, opened 2020-11-16
feat(monorepo): complain if lockfile is not fresh

wincent:
This should ensure that we don't run into the kind of issue [mentioned here](https://github.com/liferay/liferay-frontend-projects/pull/233#issuecomment-727938944).

Note that `--frozen-lockfile` is supposed to take care of this for us, but it is [a known issue](https://github.com/yarnpkg/yarn/issues/5840) that it doesn't work properly in monorepos. There is [a fix for it](https://github.com/yarnpkg/yarn/pull/6554), but the issue is nearly two years old and the PR is almost as old. Maintainer [replies here](https://github.com/yarnpkg/yarn/pull/6554#issuecomment-532667774) that:

> The change appears sound, but we're currently working on the v2 (scheduled before the end of the year) and I'm somewhat worried about introducing subtle bugs right before releasing the next major (the way I see it, the current v1 is, despite its shortcomings, relatively stable)."

We face a similar concern in liferay-portal and have [an LPS ticket](https://issues.liferay.com/browse/LPS-110313) for that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,27 @@ env:
     yarn-cache-path: .yarn
 
 jobs:
+    check-lockfile:
+        runs-on: ubuntu-latest
+
+        strategy:
+            matrix:
+                node-version: [12.x]
+
+        steps:
+            - uses: actions/checkout@v2
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v1
+              with:
+                  node-version: ${{ matrix.node-version }}
+            - name: Use or update Yarn cache
+              uses: actions/cache@v2
+              with:
+                  path: ${{ env.yarn-cache-path }}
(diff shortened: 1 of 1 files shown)


wincent:
So this failed as expected, because `master` is currently out of shape:

Screenshot 2020-11-16 -134628-ojrsgFn0@2x

Once master is updated, I'll rebase this and we should see it green.

Screenshot 2020-11-16 -134730-3P80edqb@2x

cc @izaera

izaera:
LGTM. In fact, it's failing for this PR, which means it works :sweat_smile: .

wincent:
Ok, now that [the lockfile fix](https://github.com/liferay/liferay-frontend-projects/pull/242) went into `master`, I rebased this and force-pushed. Let's hopefully see it toggle to green now.

wincent:
Screenshot 2020-11-16 -145006-MxmOcLNW@2x

Aww yis.

PR #242, opened 2020-11-16
chore: fix yarn.lock

izaera:
This should fix the problem with duplicated deps in yarn.lock.


projects/js-toolkit/package.json
```diff
@@ -3,12 +3,9 @@
 	"private": true,
 	"scripts": {
 		"build": "node scripts/build.js && node scripts/copyfiles.js --all",
-		"check-deps": "node scripts/check-deps.js",
-		"ci": "yarn clean && yarn check-deps && yarn build && yarn test",
```
wincent, line 7: You probably shouldn't entirely delete the `ci` job; we have that in all the other `package.json` so you can run it from a project (or package) directory for convenience.

- Example project-level: [npm-scripts](https://github.com/liferay/liferay-frontend-projects/blob/e24eae81b7fb14180538ad22589ae245b5e9933c/projects/npm-tools/package.json#L14).
- Example package-level: [npm-bundler](https://github.com/liferay/liferay-frontend-projects/blob/e24eae81b7fb14180538ad22589ae245b5e9933c/projects/js-toolkit/packages/npm-bundler/package.json#L34).

wincent (approved):
I pulled his down and ran `yarn` and it doesn't modifiy the lockfile, so I think this is good, apart from that one inline comment I think this is good.

projects/js-toolkit/package.json
izaera: Done.

done

izaera:
LGTG (looks good to Greg) as discussed in Slack, so merging.

wincent:
lies

PR #245, opened 2020-11-19
POC: dart sass for building css

bryceosterhaus:
This is an initial idea of how we can use dart-sass instead of our current gradle task for building sass. I'm sure I missed some nuances to how buildCSS works and some performance checks like making sure the file has been updated. I wanted to get this out there for an idea though. Check out the tests to see some examples of use.

If we want to adopt this in portal, we would run this via `liferay-npm-scripts build`.

fyi, this is just POC for the idea and direction to go. I'm not super concerned with CI and tests at the moment, we can work on those more specifically once we agree on the direction of this.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/package.json
+++ b/projects/npm-tools/packages/npm-scripts/package.json
@@ -44,6 +44,7 @@
 		"liferay-lang-key-dev-loader": "^1.0.3",
 		"liferay-npm-bridge-generator": "2.19.4",
 		"liferay-npm-bundler": "2.19.4",
+		"liferay-r2": "^0.4.4",
 		"liferay-theme-tasks": "10.0.2",
 		"metal-tools-soy": "4.3.2",
 		"mini-css-extract-plugin": "0.11.2",
@@ -55,6 +56,7 @@
 		"regenerator-runtime": "0.13.7",
 		"resolve": "^1.14.2",
 		"rimraf": "^3.0.2",
+		"sass": "^1.29.0",
 		"sass-loader": "^8.0.2",
 		"script-ext-html-webpack-plugin": "2.1.4",
 		"style-ext-html-webpack-plugin": "4.1.2",
@@ -91,5 +93,8 @@
 		"preversion": "liferay-workspace-scripts ci",
 		"test": "liferay-workspace-scripts test"
 	},
-	"version": "36.0.1"
+	"version": "36.0.1",
+	"devDependencies": {
+		"@clayui/css": "^3.20.0"
+	}
 }
--- a/projects/npm-tools/packages/npm-scripts/package.json
+++ b/projects/npm-tools/packages/npm-scripts/package.json
@@ -93,8 +93,5 @@
 		"preversion": "liferay-workspace-scripts ci",
 		"test": "liferay-workspace-scripts test"
 	},
-	"version": "36.0.1",
-	"devDependencies": {
-		"@clayui/css": "^3.20.0"
-	}
+	"version": "36.0.1"
 }
--- a/projects/npm-tools/packages/npm-scripts/src/sass/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/sass/build.js
@@ -29,24 +29,22 @@ const SASS_EXCLUDE = [
 	'**/tmp/**',
 ];
 
-async function collectSassFiles(baseDir, excludes = []) {
+function collectSassFiles(baseDir, excludes = []) {
 	const excludePaths = [
 		...SASS_EXCLUDE.map((exclude) => path.join(baseDir, exclude)),
 		...excludes.map((exclude) => path.join(baseDir, exclude)),
 	];
 
-	return await globby([
+	return globby([
 		path.join(baseDir, '**/*.scss'),
 		...excludePaths.map((exclude) => `!${exclude}`),
 	]);
 }
 
 function buildSass(file, output, includePaths) {
-	const newFileName = file
-		.split('/')
-		[file.split('/').length - 1].replace('.scss', '.css');
+	const cssFileName = path.basename(file).replace(/\.scss$/, '.css');
 
-	const outputFile = path.join(output, newFileName);
+	const outputFile = path.join(output, cssFileName);
 
 	const result = sass.renderSync({
 		file,
@@ -55,13 +53,11 @@ function buildSass(file, output, includePaths) {
 		sourceMap: true,
 	});
 
-	return [outputFile, result.css, result.map];
+	return {...result, outputFile};
 }
 
 function copyCss(filePath, newDirectory) {
-	if (!fs.existsSync(newDirectory)) {
-		fs.mkdirSync(newDirectory, {recursive: true});
-	}
+	fs.mkdirSync(newDirectory, {recursive: true});
 
 	fs.copyFileSync(filePath, path.join(newDirectory, path.basename(filePath)));
 }
@@ -80,32 +76,28 @@ async function main(
 		return;
 	}
 
-	for (let i = 0; i < files.length; i++) {
-		const file = files[i];
-
+	for (const file of files) {
 		const baseBuildDirectory = path.join(
-			baseDir,
+			path.isAbsolute(outputDir) ? '' : baseDir,
 			outputDir,
-			path.parse(file).dir.replace(baseDir, '')
+			path.dirname(path.relative(baseDir, file))
 		);
 
 		const sassBuildDirectory = path.join(baseBuildDirectory, SASS_DIR);
 
 		copyCss(file, baseBuildDirectory);
 
-		const [outputFile, css, sourceMap] = buildSass(
+		const {css, outputFile, sourceMap} = buildSass(
 			file,
 			sassBuildDirectory,
 			imports
 		);
 
-		if (!fs.existsSync(sassBuildDirectory)) {
-			fs.mkdirSync(sassBuildDirectory, {recursive: true});
-		}
+		fs.mkdirSync(sassBuildDirectory, {recursive: true});
 
 		if (rtl) {
 			fs.writeFileSync(
-				outputFile.replace('.css', '_rtl.css'),
+				outputFile.replace(/\.css$/, '_rtl.css'),
 				r2.swap(css.toString())
 			);
 		}
--- a/projects/npm-tools/packages/npm-scripts/src/sass/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/sass/build.js
@@ -1,0 +1,124 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const fs = require('fs');
+const globby = require('globby');
+const r2 = require('liferay-r2');
+const path = require('path');
+const sass = require('sass');
+
+const log = require('../utils/log');
+
+const SASS_EXCLUDE = [
+	'**/_*.scss',
+	'**/build/**',
+	'**/classes/**',
+
+	// Copy from CSSBuilderArgs.java
+
+	'**/_diffs/**',
+	'**/.sass-cache*/**',
+	'**/.sass_cache_*/**',
+	'**/_sass_cache_*/**',
+	'**/_styled/**',
+	'**/_unstyled/**',
+	'**/css/aui/**',
+	'**/css/clay/**',
+	'**/tmp/**',
+];
+
+const getClayPaths = () => {
+	try {
+		return require('@clayui/css').includePaths;
+	} catch (e) {
+		return [];
+	}
+};
+
+const clayIncludePaths = getClayPaths();
+
+async function collectSassFiles(baseDir, excludes = []) {
+	const excludePaths = [
+		...SASS_EXCLUDE.map((exclude) => path.join(baseDir, exclude)),
+		...excludes.map((exclude) => path.join(baseDir, exclude)),
+	];
+
+	return await globby([
+		path.join(baseDir, '**/*.scss'),
+		...excludePaths.map((exclude) => `!${exclude}`),
+	]);
+}
+
+function buildSass(file, output) {
+	const newFileName = file
+		.split('/')
+		[file.split('/').length - 1].replace('.scss', '.css');
+
+	const outputFile = path.join(output, newFileName);
+
+	const result = sass.renderSync({
+		file,
+		includePaths: clayIncludePaths,
+		outFile: outputFile,
+		sourceMap: true,
+	});
+
+	return [outputFile, result.css, result.map];
+}
+
+function copyCss(filePath, newDirectory) {
+	if (!fs.existsSync(newDirectory)) {
+		fs.mkdirSync(newDirectory, {recursive: true});
+	}
+
+	fs.copyFileSync(filePath, path.join(newDirectory, path.basename(filePath)));
+}
+
+const SASS_DIR = '.sass-cache';
+
+async function main(baseDir, {excludes, outputDir = '', rtl = false}) {
+	const files = await collectSassFiles(baseDir, excludes);
+
+	if (!files.length) {
+		log(`BUILD CSS: No files found.`);
+
+		return;
+	}
+
+	for (let i = 0; i < files.length; i++) {
+		const file = files[i];
+
+		const baseBuildDirectory = path.join(
+			baseDir,
+			outputDir,
+			path.parse(file).dir.replace(baseDir, '')
+		);
+
+		const sassBuildDirectory = path.join(baseBuildDirectory, SASS_DIR);
+
+		copyCss(file, baseBuildDirectory);
+
+		const [outputFile, css, sourceMap] = buildSass(
+			file,
+			sassBuildDirectory
+		);
+
+		if (!fs.existsSync(sassBuildDirectory)) {
+			fs.mkdirSync(sassBuildDirectory, {recursive: true});
+		}
+
+		if (rtl) {
+			fs.writeFileSync(
+				outputFile.replace('.css', '_rtl.css'),
+				r2.swap(css.toString())
+			);
+		}
+
+		fs.writeFileSync(outputFile, css);
+		fs.writeFileSync(outputFile + '.map', sourceMap);
+	}
+}
+
+module.exports = main;
--- a/projects/npm-tools/packages/npm-scripts/src/sass/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/sass/build.js
@@ -29,16 +29,6 @@ const SASS_EXCLUDE = [
 	'**/tmp/**',
 ];
 
-const getClayPaths = () => {
-	try {
-		return require('@clayui/css').includePaths;
-	} catch (e) {
-		return [];
-	}
-};
-
-const clayIncludePaths = getClayPaths();
-
 async function collectSassFiles(baseDir, excludes = []) {
 	const excludePaths = [
 		...SASS_EXCLUDE.map((exclude) => path.join(baseDir, exclude)),
@@ -51,7 +41,7 @@ async function collectSassFiles(baseDir, excludes = []) {
 	]);
 }
 
-function buildSass(file, output) {
+function buildSass(file, output, includePaths) {
 	const newFileName = file
 		.split('/')
 		[file.split('/').length - 1].replace('.scss', '.css');
@@ -60,7 +50,7 @@ function buildSass(file, output) {
 
 	const result = sass.renderSync({
 		file,
-		includePaths: clayIncludePaths,
+		includePaths,
 		outFile: outputFile,
 		sourceMap: true,
 	});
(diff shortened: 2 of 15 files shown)


projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+	'**/tmp/**',
+];
+
+const getClayPaths = () => {
+	try {
+		return require('@clayui/css').includePaths;
+	} catch (e) {
+		return [];
+	}
+};
```
bryceosterhaus, line 38: This is a little weird. Might be some other ways to do this, but this was my first thought.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+
+		const baseBuildDirectory = path.join(
+			baseDir,
+			outputDir,
+			path.parse(file).dir.replace(baseDir, '')
+		);
+
+		const sassBuildDirectory = path.join(baseBuildDirectory, SASS_DIR);
+
+		copyCss(file, baseBuildDirectory);
```
bryceosterhaus, line 101: This idea is copied from the gradle task, it both copies the original scss files and builds them into css files which get thrown into the `.sass-cache` directory

projects/npm-tools/packages/npm-scripts/test/sass/build.js
```diff
@@ -0,0 +1,151 @@
+/**
+ * @jest-environment node
+ */
```
bryceosterhaus, line 3: This was necessary for dart-sass to work and access the file system. If we drop this below our copyright, it doesn't work as expected.

projects/npm-tools/packages/npm-scripts/package.json
```diff
@@ -44,6 +44,7 @@
 		"liferay-lang-key-dev-loader": "^1.0.3",
 		"liferay-npm-bridge-generator": "2.19.4",
 		"liferay-npm-bundler": "2.19.4",
+		"liferay-r2": "^0.4.4",
```
jbalsas, line 47: @wincent, do you remember where this currently lives? 😂 

Maybe time to bring this over to `liferay-frontend-projects`?

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+const fs = require('fs');
+const globby = require('globby');
+const r2 = require('liferay-r2');
+const path = require('path');
+const sass = require('sass');
+
+const log = require('../utils/log');
+
+const SASS_EXCLUDE = [
+	'**/_*.scss',
```
jbalsas, line 15: I thought Sass did this by default... 🤔

Based on https://sass-lang.com/guide:

> Partials
>
> You can create partial Sass files that contain little snippets of CSS that you can include in other Sass files. [...] A partial is a Sass file named with a leading underscore. [...] **The underscore lets Sass know that the file is only a partial file and that it should not be generated into a CSS file**. Sass partials are used with the @use rule.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
jbalsas: What we have right now is that we allow passing `includePaths` to sass. So, we'd want https://github.com/liferay/liferay-frontend-projects/pull/245/files#diff-f3abf7ec1699e822b3182c10c3315fa3bd6516406a6c299b82c5b27f6d99155fR63 to be a configurable param.

That way, we could rely on whatever `@clayui/css` version is in `liferay-portal` at the time and not in what `npm-tools` is bringing over.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+function buildSass(file, output) {
+	const newFileName = file
+		.split('/')
+		[file.split('/').length - 1].replace('.scss', '.css');
+
+	const outputFile = path.join(output, newFileName);
+
+	const result = sass.renderSync({
+		file,
+		includePaths: clayIncludePaths,
```
jbalsas, line 63: This should be some type of passed-in param so we can add includepaths by module.

It would definitely be great to default this to `@clayui/css` for simplicity, although that means all sass compilation times will be slower regardless of them using Clay or not.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
jbalsas: Yeah, should be okay for now since we're mainly just focused on moving to `dart-sass`. We can take care of this when we work on the minification scheme architecture which will surely change more things.

projects/npm-tools/packages/npm-scripts/test/sass/build.js
```diff
@@ -0,0 +1,151 @@
...
+		});
+
+		expect(fs.existsSync(OUTPUT_SASS_DIR_PATH + '/excluded.css')).toBe(
+			false
+		);
+		expect(fs.existsSync(OUTPUT_SASS_DIR_PATH + '/included.css')).toBe(
+			true
+		);
+	});
+});
```
wincent, line 151: Very nice tests!

projects/npm-tools/packages/npm-scripts/test/sass/build.js
wincent: I didn't even know this existed. 😂 [Docs](https://jestjs.io/docs/en/configuration#testenvironment-string) show an example of it working below a comment, so it's weird that it doesn't work.

Instead of adding it inline though (where the linter will try to re-add the copyright above it), you _might_ be able to configure it elsewhere, by setting up an override in the [`jest.config.js`](https://github.com/liferay/liferay-frontend-projects/blob/623715898e497fad03bf7cb489fec7537f057bf8/projects/npm-tools/packages/npm-scripts/jest.config.js) for this project (not sure if that will break other tests; it might not seeing as this package is all about node).

projects/npm-tools/packages/npm-scripts/package.json
wincent: I [commented on this elsewhere](https://github.com/liferay/liferay-frontend-projects/pull/245#discussion_r527554630). It's here: https://github.com/liferay-labs/r2

projects/npm-tools/packages/npm-scripts/src/sass/build.js
bryceosterhaus: > That way, we could rely on whatever @clayui/css version is in liferay-portal at the time and not in what npm-tools is bringing over.

@wincent,I didn't want to specify clay as a dependency in this package since we would have to release with every new clay version. By having this check, we protect from test failures without having to have clay as a dependency. We really wouldn't expect this to ever fail in the context of liferay-portal(only place this is used), but by doing this we don't require clay as a dependency in this package.

Not sure if this is the best way to go about it, but it made sense to me at the time.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
bryceosterhaus: @jbalsas this was also specified in our java implementation. I haven't tested it without it, but I think it seems reasonable to make it an explicit exclude.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
bryceosterhaus: 👍

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const fs = require('fs');
+const globby = require('globby');
```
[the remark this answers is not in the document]
bryceosterhaus: 👍  I didn't delve to deep into this, I just noticed it as an already existing dependency and went with it.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+	const excludePaths = [
+		...SASS_EXCLUDE.map((exclude) => path.join(baseDir, exclude)),
+		...excludes.map((exclude) => path.join(baseDir, exclude)),
+	];
+
+	return await globby([
+		path.join(baseDir, '**/*.scss'),
+		...excludePaths.map((exclude) => `!${exclude}`),
+	]);
+}
```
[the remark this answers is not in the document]
bryceosterhaus: ah I didn't know that. Thanks for the heads up

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+
+function buildSass(file, output) {
+	const newFileName = file
+		.split('/')
```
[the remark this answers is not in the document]
bryceosterhaus: This was more of just something to work as I tested this out. Likely should just use `path`api to get filename

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+async function main(baseDir, {excludes, outputDir = '', rtl = false}) {
+	const files = await collectSassFiles(baseDir, excludes);
+
+	if (!files.length) {
+		log(`BUILD CSS: No files found.`);
+
+		return;
+	}
+
+	for (let i = 0; i < files.length; i++) {
```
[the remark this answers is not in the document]
bryceosterhaus: @wincent do we allow this in portal code now? For some reason I remember a "liferay rule" where we didn't use `for of` or `for in`

projects/npm-tools/packages/npm-scripts/test/sass/build.js
```diff
@@ -0,0 +1,151 @@
...
+);
+
+describe('sass', () => {
+	beforeEach(() => {
+		fs.mkdirSync(OUTPUT_DIR_PATH);
+	});
+
+	afterEach(() => {
+		rimraf.sync(OUTPUT_DIR_PATH);
+	});
```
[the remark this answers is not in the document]
bryceosterhaus: Ah, I always forget about `os.tmpdir()`

projects/npm-tools/packages/npm-scripts/src/sass/build.js
bryceosterhaus: Just tested it, if we don't include this here then the partials files will build into their own css files

projects/npm-tools/packages/npm-scripts/src/sass/build.js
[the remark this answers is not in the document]
bryceosterhaus: @wincent I tried using `getPaths` but it seems to require that the directory to glob from always starts at `'.'`[here](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/packages/npm-scripts/src/utils/expandGlobs.js#L206).

For processing sass, I think we need to be able to specify the absolute path of where the scss is located. I could be missing something here, but I think we need to update `getPaths` and `expandGlobs` to accommodate for a different base directory and also needs to support a list of excludes rather than a file path for excluded files.

I can go through and make these changes, but I didn't want to mess with those utils quite yet in case I broke something that was dependent on the starting directory for traverse.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const fs = require('fs');
+const globby = require('globby');
+const r2 = require('liferay-r2');
```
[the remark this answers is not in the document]
bryceosterhaus: I don't know much about it other than it was mentioned in the issue. I haven't looked too deeply at the rtl support and nuances of the java implementation we use

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+			sassBuildDirectory
+		);
+
+		if (!fs.existsSync(sassBuildDirectory)) {
+			fs.mkdirSync(sassBuildDirectory, {recursive: true});
+		}
+
+		if (rtl) {
+			fs.writeFileSync(
+				outputFile.replace('.css', '_rtl.css'),
```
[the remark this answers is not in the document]
bryceosterhaus: @wincent what would be a case where `/\.css$/` would succeed but `.css` would fail? Are you thinking if the filepath was something like `some/base/dir.css/contains/main.css`?

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,124 @@
...
+		return;
+	}
+
+	for (let i = 0; i < files.length; i++) {
+		const file = files[i];
+
+		const baseBuildDirectory = path.join(
+			baseDir,
+			outputDir,
+			path.parse(file).dir.replace(baseDir, '')
```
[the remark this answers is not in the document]
bryceosterhaus: The last portion is used to append any extra nested file structure found in the glob. Using `path.relative(baseDir, File)` is a good idea, but it will still return the filename and we just want the directory. So we can do `path.dirname(path.relative(baseDir, file))`

projects/npm-tools/packages/npm-scripts/src/sass/build.js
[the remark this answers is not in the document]
wincent: On 7.4 we no longer have to worry about it; it was only ever a problem on IE (and [only in specific edge cases](https://github.com/liferay/liferay-frontend-guidelines/issues/53#issuecomment-508029423)). In repos like this one you don't have to worry about it at all:

- The `no-for-of` rule is [still turned on in the preset](https://github.com/liferay/liferay-frontend-projects/blob/801917a87c8260a01c84ddb68b839f8c92ed04ac/projects/eslint-config/index.js#L92); but...
- We [have it turned off in this repo](https://github.com/liferay/liferay-frontend-projects/blob/801917a87c8260a01c84ddb68b839f8c92ed04ac/.eslintrc.js#L25).

At some point in the future we should just drop it from the preset. When 7.3 was released we laid the groundwork that would enable us to apply different Babel rules on different branches (ie. so that we can use the same version of `@liferay/npm-scripts` everywhere), so it would be a natural extension of that to allow the same kind of thing for ESLint (and even if we didn't want to do that, there's always `.eslintrc.js`).

projects/npm-tools/packages/npm-scripts/src/sass/build.js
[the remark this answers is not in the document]
wincent: > For processing sass, I think we need to be able to specify the absolute path of where the scss is located. I could be missing something here, but I think we need to update `getPaths` and `expandGlobs` to accommodate for a different base directory and also needs to support a list of excludes rather than a file path for excluded files.
>
> I can go through and make these changes, but I didn't want to mess with those utils quite yet in case I broke something that was dependent on the starting directory for traverse.

I guess we can make the change to not traverse from `'.'` _if we are sure that we need to_ — that was just a simplifying assumption made in the name of robustness (because if you want to support building "project x" from some other directory belonging "project y", then you suddenly have 100s of places in the code base where you have to uphold the convention rather than just assuming that `process.cwd()` is always set to "the right place"... which sounds like a recipe for bugs). What's the reason we need absolute paths here again?

As for:

> also needs to support a list of excludes rather than a file path for excluded files

It already supports [negated globs](https://github.com/liferay/liferay-frontend-projects/blob/542e42c5b24897a1e63016afae7c44e7351e526e/projects/npm-tools/packages/npm-scripts/src/utils/getRegExpForGlob.js#L63), so maybe you don't need that?

projects/npm-tools/packages/npm-scripts/src/sass/build.js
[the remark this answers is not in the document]
wincent: Yep, that's exactly the kind of case I was thinking about. We might say, "we don't have any crazy paths like that", but months or years from now we _could_ wind up with one and if that happens, stuff will break. This kind of unanticipated thing happens often enough to be annoying.

As a rule of thumb, you don't have to be paranoid about things that _might_ happen in the future, but in cases like this, where the cost of saying _exactly what you mean_ as opposed to _something approximately like what you mean_, then you should do the former; eg.

- **Exactly what you mean:** "path ends with `.css`" → use an anchored regex → file suffix gets transformed.
- **Something approximately like what you mean:** "path contains `.css` as a substring" → use a string → first instance of substring gets replaced, which is currently what you want, but not guaranteed to be in the future.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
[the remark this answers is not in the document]
wincent: Thanks for the explanation. This stuff is hard to read. Your `path.dirname(path.relative(baseDir, file))` looks good to me.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
wincent: Having it as a `devDependency` like you do in this PR seems ok, especially if you make the version range pretty loose (makes it easier to update with `yarn upgrade-interactive` — and if all you need is to have it there for the purposes of running the tests, it wouldn't even need to be kept up-to-date as far as I can tell); as a dev dep, the actual version used here won't influence DXP at all.

The other option if it's just for testing is to create a scratch directory under `os.tmpdir()` and install Clay there from the network during the test run. Would be slower if you did it in a `beforeEach()`, but probably not in a `beforeAll()` (because CI is either installing it once as part of initial `yarn` set-up, or once as part of test run).

jbalsas:
Our bureaucracy department created https://github.com/liferay/liferay-frontend-projects/issues/248 so we can officially stop calling this a PoC ;)

projects/npm-tools/packages/npm-scripts/src/sass/build.js
bryceosterhaus: I'm actually just going to remove the clay stuff and instead test against scss using `imports` and making my own file to include variables. 

I'm removing clay because its not always included when portal is building scss files. Any module that wants to use clay-css, references it directly via

```
buildCSS {
	imports = [
		new File(npmInstall.nodeModulesDir, "@clayui/css/src/scss")
	]
}
```

projects/npm-tools/packages/npm-scripts/src/sass/build.js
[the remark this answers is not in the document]
bryceosterhaus: > What's the reason we need absolute paths here again?

The funny thing is, we might not need to in portal to reference anything outside of `'.'` in portal, but when it comes to our unit tests, we have to move back to the root of this mono-repo to reference the `__fixtures__` directory.


> It already supports negated globs, so maybe you don't need that?

👍  didn't realize that

projects/npm-tools/packages/npm-scripts/src/sass/build.js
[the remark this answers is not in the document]
bryceosterhaus: Makes sense, just wanted to make sure I was thinking through it similarly, thanks!

projects/npm-tools/packages/npm-scripts/src/sass/build.js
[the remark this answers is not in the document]
bryceosterhaus: > On 7.4 we no longer have to worry about it
> We have it turned off in this repo.

bryceosterhaus:
@wincent @jbalsas updated from your comments

So `expandGlobs` doesn't seem to accept an absolute path when comparing a path. I added a test [here](https://github.com/liferay/liferay-frontend-projects/pull/245/files#diff-cf084c0e1a629755211fc83e10e32538b4f43635321386077086c12f824925d4R33-R36) that shows it won't work with an absolute path. So in order to get around this I added [this test path](https://github.com/liferay/liferay-frontend-projects/pull/245/commits/da6af3d17bf8b1a5cff4b06ee7c841a7c451dc7d#diff-2067b8c7a8417bf8c2735f55ba1118f01e8f7afcc8c0259a856167322c5c4f76R133). I'm not sure this is the best way around it, but it seemed safe to go that route right now rather than update your regex util for globs.


My next steps are integrating this build process into the executable `build` script and allowing for proper configs that we already have in portal.

wincent:
> So `expandGlobs` doesn't seem to accept an absolute path when comparing a path. I added a test [here](https://github.com/liferay/liferay-frontend-projects/pull/245/files#diff-cf084c0e1a629755211fc83e10e32538b4f43635321386077086c12f824925d4R33-R36) that shows it won't work with an absolute path. So in order to get around this I added [this test path](https://github.com/liferay/liferay-frontend-projects/pull/245/commits/da6af3d17bf8b1a5cff4b06ee7c841a7c451dc7d#diff-2067b8c7a8417bf8c2735f55ba1118f01e8f7afcc8c0259a856167322c5c4f76R133). I'm not sure this is the best way around it, but it seemed safe to go that route right now rather than update your regex util for globs.

Seems reasonable to me.

bryceosterhaus:
follow up at https://github.com/liferay/liferay-frontend-projects/pull/282

PR #251, opened 2020-11-26
fix(npm-scripts): provide (inactive-by-default) tooling for formatting JS in JSP inside build

wincent:
Just testing the [claim made in this Slack thread](https://liferay.slack.com/archives/C04D8HMEF/p1606323974062700?thread_ts=1606233408.059000&cid=C04D8HMEF) that the reason why my earlier attempts to minify-in-place inside "build/" may have seemed to not be working due to Gradle caching.

While the cache may cause problems, I am skeptical that this will work because in my testing I found that `processResources` happens too late (after `npm-scripts` runs), but in the interests of thoroughness I am going to try this out and test it in a full `ant all` run anyway.

---

Final summary: we tested this out and the minify-in-`build/` trick does work, but even with minification turned way down to avoid reordering issues, testing in DXP shows that it is too risky to turn on by default (for example, we see other parts of the toolchain choking as a result of tags + lines following each other on the same line — probably something we could work around with some careful semicolon insertion, but it would be very fiddly...). So last commit in this PR gates the functionality behind an environment variable which will be empty by default.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/jsp/padLines.js
+++ b/projects/npm-tools/packages/npm-scripts/src/jsp/padLines.js
@@ -15,7 +15,7 @@ const PADDING_MARKER = '\u00abpad\u00bb';
  */
 const DEFAULT_PADDING = `void 0; /* ${PADDING_MARKER} */`;
 
-const PADDING_LINE = /(?:\s*void\s+0;\s*)?\/\*\s*\u00abpad\u00bb\s*\*\/\n/g;
+const PADDING_LINE = /(?:\s*void\s+0;?\s*)?\/\*\s*\u00abpad\u00bb\s*\*\/\n?/g;
 
 /**
  * Prefixes `string` with lines containing the `padding` string such that the
--- a/projects/npm-tools/packages/npm-scripts/src/utils/minify.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/minify.js
@@ -89,6 +89,7 @@ async function minify() {
 								// Add `beautify: true` here to debug tests, if
 								// necessary.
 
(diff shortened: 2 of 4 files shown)


wincent:
Happily, this looks like it is going to work _and_ additionally it uncovered a latent bug as seen [in this diff](https://gist.github.com/wincent/9c98f3d77fcca8a3753e680137f5f3a8).

You can see that the minifier is freely reordering stuff which means that some of our JSP syntax substitutions get restored into the wrong "slots" after minification.

For example, `data = Liferay.Util.ns('_<%= HtmlUtil.escapeJS(portletResource) %>_', data)` gets restored as `(data = Liferay.Util.ns('_<portlet:namespace />_', data))`; and the corresponding inverse error is `<portlet:namespace />resetPreview` getting restored as `<%= HtmlUtil.escapeJS(portletResource) %>resetPreview`.

So I'm going to turn the minification waaaay down for JSP so that it is basically only whitespace changes (it already was dialled way down, but I didn't anticipate that it would make this kind of reordering), add a regression test, and retry.

wincent:
Will obviously need to rebase and drop the pre-release commit(s) before calling this one final.

jbalsas:
> You can see that the minifier is freely reordering stuff which means that some of our JSP syntax substitutions get restored into the wrong "slots" after minification.

didntseethatcoming

wincent:
Ok, based on testing [over in my fork](https://github.com/wincent/liferay-portal/pull/371), I am "de-fanging" this by turning it off by default. We can look at turning it on later if we feel brave, but for now it remains a harmless no-op. Taking it out of draft.

jbalsas:
finishline

Issue #250, opened 2020-11-24
Add support for webpack 5 federation zeroconf


izaera:
### Issue type

-   :gift: Feature request

### Description

Given the manual config shown in [this PR](https://github.com/liferay-frontend/liferay-portal/pull/527) we need to internalize as much as possible inside `npm-scripts` to make things automagic.

PR #252, opened 2020-11-26
feat(npm-scripts): add webpack federation support

izaera:
This is implementation of #250 .

It simply adds a `--federation` switch to `npm-scripts build` so that selected projects from `liferay-portal` have a webpack federation configuration injected (in addition to any other webpack configuration they may have defined).

Note that if the project does not declare any webpack build, the `--federation` switch will force it even in the absence of a `webpack.config.js` file.

By default, the webpack configuration is automatic (as per this [comment](https://github.com/liferay-frontend/liferay-portal/pull/527#issuecomment-732148545)). However it is possible that in the future we may need overriding the injected config: we will see then, if the need appears, how to do it.

This can be tested with [this PR](https://github.com/liferay-frontend/liferay-portal/pull/527), but I **strongly advise** to make a pre-release (whether in a local npm registry or the default public one) and refrain from using `yarn link`, `file:///` URLs or any similar technique, because it will fail with very strange symptoms (because of the portal's dependency graph).

I have tested it manually by running `ant all` in a portal and it builds and seems to work.


wincent:
> It simply adds a `--federation` switch to `npm-scripts build`

Going to look at the code, but one thing to bear in mind straight away is that adding a switch may require changes to the Java Source Formatter because it [expects the build string to have one of four exact formats](https://github.com/liferay/liferay-portal/blob/5aceff3ec63a5118178d7967f0903964b769a773/modules/util/source-formatter/src/main/java/com/liferay/source/formatter/checks/JSONPackageJSONCheck.java#L67-L88):

- `build`
- `liferay-npm-bundler`
- `theme build`
- `webpack`

Specifically, it asserts that the strings _end_ with that. So we either need to modify the formatter, or perform this configuration via `npmscripts.config.js`.

Anyway, I'm going to look at the rest of the code now.

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const fs = require('fs');
+
+/**
+ *
+ */
```
wincent: No empty comments please.

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
```diff
@@ -0,0 +1,32 @@
...
+function parseBnd(filePath = './bnd.bnd') {
```
wincent, line 18: You don't need the `./`.

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
```diff
@@ -0,0 +1,32 @@
...
+	const content = fs.readFileSync(filePath).toString();
```
wincent: Pro-Tip™: You can avoid the `toString()` call by changing this to:

```js
const content = fs.readFileSync(filePath, 'utf8');
```

(I think that's the pattern we use in most places in `npm-scripts`, and it's shorter too.)

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
```diff
@@ -0,0 +1,32 @@
...
+	return lines.reduce((entries, line) => {
+		line = line.trim();
+
+		if (line === '') {
+			return entries;
+		}
+
+		const i = line.indexOf(':');
+		const key = line.substring(0, i);
+		const value = line.substring(i + 2);
```
wincent, line 35: I know we expect whitespace after the `:` and source formatter will require it, but what if we ever run on an unformatted file? Shouldn't we avoid hard-coding the `+ 2` here?

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
```diff
@@ -0,0 +1,32 @@
...
+
+		entries[key] = value;
+
+		return entries;
```
wincent, line 39: Should we handle line continuations (`\`)? It seems a bit dangerous to call the function `parseBnd` if it only produces a valid parse result for a subset of bnd files (ie. ones without properties that span multiple lines).

izaera:
>  So we either need to modify the formatter, or perform this configuration via `npmscripts.config.js`.

Yes, I saw the [other PR] failing because of the SF, so I'll move the switch to `npmscripts.config.js`. In fact it will simplify the code. Turns out that using a CLI flag wasn't a good idea :facepalm: .

projects/npm-tools/packages/npm-scripts/test/utils/parseBnd.js
```diff
@@ -0,0 +1,36 @@
...
+		return path.join(
+			__dirname,
+			'..',
+			'..',
+			'__fixtures__',
+			'utils',
+			'parseBnd',
+			'bnd.bnd'
+		);
+	}
```
wincent, line 21: If you wanted, there is a `getFixture` function that you can use.

[Example usage here](https://github.com/liferay/liferay-frontend-projects/blob/de5eb0773d8c19c86e1bcc4ff2a97f0eb024e5f6/projects/npm-tools/packages/npm-scripts/test/jsp/extractJS.js#L163).

projects/npm-tools/packages/npm-scripts/src/utils/tweakWebpackConfig.js
```diff
@@ -0,0 +1,166 @@
...
+const getMergedConfig = require('./getMergedConfig');
+const parseBnd = require('./parseBnd');
+
+const CORE_REMOTES = ['frontend-js-react-web'];
+const CORE_SHARES = ['react'];
+
+const BABEL_CONFIG = getMergedConfig('babel');
+
+function tweakWebpackConfig(webpackConfig, {federation} = {}) {
+	const ret = [];
```
wincent, line 29: In the guidelines we ask that people [avoid abbreviations](https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/naming.md#avoid-abbreviations).

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
izaera: We are handling them (here](https://github.com/liferay/liferay-frontend-projects/pull/252/files#diff-333307525d25dd06bb5a64f84cc557d8cb21337d0be2a934f1947bf107153e2eR22). Maybe I should add comments explaining what I'm doing, since regexps are so cryptical.

projects/npm-tools/packages/npm-scripts/src/utils/tweakWebpackConfig.js
```diff
@@ -0,0 +1,166 @@
...
+		module: {
+			rules: [
+				{
+					exclude: /node_modules/,
+					test: /\.js$/,
+					use: {
+						loader: 'babel-loader',
+						options: {
+							plugins: [
+								'@babel/plugin-proposal-class-properties',
```
wincent, line 79: Why do we include this specifically?

If we do need to include some extra options here, it would be nice if we could keep this in sync, at least as close as possible, with [our Babel config](https://github.com/liferay/liferay-frontend-projects/blob/de5eb0773d8c19c86e1bcc4ff2a97f0eb024e5f6/projects/npm-tools/packages/npm-scripts/src/config/babel.json#L17-L21) which has a broader list:

```
"@babel/proposal-class-properties",
"@babel/proposal-export-namespace-from",
"@babel/proposal-nullish-coalescing-operator",
"@babel/proposal-object-rest-spread",
"@babel/proposal-optional-chaining",
```

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
izaera: :thinking: Not sure. I thought it was a requirement (at least it is in the final `MANIFEST.MF` AFAIK). But it is probable that `bnd` doesn't complain if it is missing, so I will trim it instead, just to be safe :+1:

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
izaera: Where it says "here" it is referring to line 22 (my github UI doesn't jump to the expected place when I click the link :roll_eyes: ).

projects/npm-tools/packages/npm-scripts/src/utils/tweakWebpackConfig.js
izaera: True :+1: . I forgot to refactor that.

projects/npm-tools/packages/npm-scripts/test/utils/tweakWebpackOptions.js
```diff
@@ -0,0 +1,197 @@
...
+				'tweakWebpackConfig'
+			)
+		);
+	});
+
+	afterEach(() => {
+		fs.rmdirSync('build', {recursive: true});
+
+		process.chdir(savedCwd);
+	});
```
wincent, line 139: It's probably best not to write files inside the repo during a test run. This is something that I pointed out in Bryce's PR [over here](https://github.com/liferay/liferay-frontend-projects/pull/245#discussion_r527577883):

> If the tests crash while running, you'll wind up with a dirty work tree here. That's why in other tests in `npm-scripts` we create a temporary directory and do all our tests in there.
>
> If you wanted to take that approach, you could find examples [like this one](https://github.com/liferay/liferay-frontend-projects/blob/542e42c5b24897a1e63016afae7c44e7351e526e/projects/npm-tools/packages/npm-scripts/test/utils/expandGlobs.js) by searching for `os.tmpdir()`.
>
> You don't need to clean up the files in that case (the OS will clean the temp dirs), so you can avoid the dependency on `rimraf` too, as well as the time spent erasing the files.

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
```diff
@@ -130,6 +130,16 @@ function getMergedConfig(type, property) {
 			]);
 			break;
 
+		case 'bundler-with-federation':
+			mergedConfig = deepMerge([
+				require('../config/npm-bundler'),
+				getUserConfig('npmbundler'),
+				{
+					ignore: ['webpack/**/*'],
+				},
```
wincent, line 139: The order here seems a bit off compared to every other config type that we merge: ie. "our base stuff" → "user stuff" → "more of our stuff" (vs all the others, which let the user have the last say).

Could we just fold the "more of our stuff" into `../config/npm-bundler`? I know a user is never going to want to override this, but just for consistency... and the `ignore` would be harmless as far as I can tell the `case 'bundler'` block (because nobody is writing to `webpack/**/*`).

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -4,6 +4,7 @@
  */
 
 const fs = require('fs');
+const parseArgs = require('minimist');
```
wincent, line 7: I don't know whether we'll keep doing this via an option (given the source formatter thing), but in case we do...

I think using minimist is a bit of overkill here to grab a single option.

We're [already calling `minimist()` from `src/index.js`](https://github.com/liferay/liferay-frontend-projects/blob/de5eb0773d8c19c86e1bcc4ff2a97f0eb024e5f6/projects/npm-tools/packages/npm-scripts/src/index.js#L11-L13), and now calling it again here for what would be a simple `argv.includes('--federation')`.

(Personally, I never would have added `minimist` to this project in the first place because it is using [245 lines of JS](https://github.com/substack/minimist/blob/master/index.js) to do what can be done in a single expression... 🤦‍♂️)

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -95,11 +98,11 @@ module.exports = async function () {
 		'--source-maps'
 	);
 
-	if (fs.existsSync(path.join(CWD, 'webpack.config.js'))) {
-		webpack();
+	if (fs.existsSync(path.join(CWD, 'webpack.config.js')) || args.federation) {
```
wincent, line 101: I don't think you need the `path.join` call here. You are in the current working directory already.

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
izaera: Yeah, maybe that would simplify the code too. My first thought was that someone would override that ignore accidentally, but seeing it from the other side, if someone overrides it is because they are using `webpack` folder for something and that will surely lead to some kind of error (because of the conflict).

Maybe this is a sign that `webpack` is not a good choice for the name of this internal folder. Perhaps we should change it to something more internal. Like `_liferay_webpack` or something that denotes its technical nature. 

What do you think?

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: The fact is that I used `minimist` because it was there :sweat_smile: . My first implementation used `includes`.

It has happened the same with the `getFixture` code: I copied my code from some other test that didn't use it.

Maybe we should refactor these things every once in a while to make sure the project looks the same in every file. I know it's not easy and it will never be perfect, but maybe we can create low priority issues for these things and do them every once in a while if we have time to spare...

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: :roll_eyes: :joy: 

`path.relative(CWD, path.resolve( path.join( CWD, 'webpack.config.js') ) )`

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
```diff
@@ -15,37 +16,35 @@ const WEBPACK_DEV_CONFIG_FILE = 'webpack.config.dev.js';
...
  * Main function for running webpack within the liferay-portal repo.
  */
-module.exports = function (...args) {
-	const watch = args.indexOf('--watch');
-	if (watch !== -1) {
+module.exports = function (...argv) {
+	const extraArgs = argv.filter(
+		(arg) => !['--federation', '--watch'].includes(arg)
+	);
+	const args = parseArgs(argv);
```
wincent, line 23: I'd just skip `minimist` here, instead of calling it once in `src/index.js` and immediately after again here.

```js
const federation = argv.includes('--federation');
const watch = argv.includes('--watch');
```

Also, tiny nit: quadratic loop of `filter` + `includes`. I know it's not a performance problem because N is small, but just to avoid the pattern becoming normalized and then people replicating in other places where it actually does matter, I make a point of note writing nested loops like this. Instead, you could do something like:

```js
const SPECIAL_ARGS = new Set(['--federation', '--watch']);

module.exports = function (...args) {
   const extraArgs = args.filter((arg) => !SPECIAL_ARGS.has(arg));
}
```

(I don't know if the names are right, just an example).

One other minor thing, this thing that you're calling `argv` isn't actually `argv`: `argv` is the array of arguments passed to the executable when invoked — what you're getting here is a subset of those, so technically the name `argv` is giving a false impression of its contents.

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
```diff
@@ -16,8 +16,12 @@ const fs = require('fs');
  * @return {object} a hash of parsed key-value pairs.
  */
 function parseBnd(filePath = './bnd.bnd') {
-	const content = fs.readFileSync(filePath).toString();
-	const lines = content.replace(/\\\n/g, ' ').split('\n');
+	let content = fs.readFileSync(filePath).toString();
+
+	content = content.replace(/\r\n/g, '\n');
+	content = content.replace(/\\\n/g, ' ');
```
wincent, line 22: Ok, and now I see that I misread this on an earlier commit when I asked about line continuations... you were already handling it. 👍

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
wincent: Ignore this — I now see you actually are handling line continuations.

wincent (commented):
I left a bunch of small comments, but I don't see any major problems, only nits, other than the thing you're already dealing with about the `--federation` option. 👏

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
wincent: > Maybe I should add comments explaining what I'm doing

Nooo! It is better to live in the darkness of ignorance.

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
wincent: > Maybe this is a sign that `webpack` is not a good choice for the name of this internal folder. Perhaps we should change it to something more internal. Like `_liferay_webpack` or something that denotes its technical nature.

Sounds great. I was thinking that too, but didn't say it because I have already said too much.

A convention I've used elsewhere is `__generated__/` — which is a pretty obvious cue not to go putting anything hand-created in there, and translates to any context where we use tool-generated files. But really, any name would be less collision-prone than `webpack`.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
wincent: Yeah, "Boy Scout Rule" — when you see something that could be cleaned up, make it a little better along the way.

jbalsas:
I'm happy as long as we don't add the `--federation` flag. Since you've already reached that same conclusion, I'm happy :)

projects/npm-tools/packages/npm-scripts/test/utils/parseBnd.js
izaera: I've created a `getFixturePath` function, because I need the path, not the contents. That's probably why the place where I copied this code from was written that way.

I will change it too to let all the code in the same state.

izaera:
Pushed changes to address revision issues. However I will check this again running `ant all` locally because I've changed so many things. :crossed_fingers:

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -19,6 +18,7 @@ const validateConfig = require('../utils/validateConfig');
 const webpack = require('./webpack');
 
 const BUILD_CONFIG = getMergedConfig('npmscripts', 'build');
+const FEDERATION_ENABLED = getMergedConfig('npmscripts', 'federation');
```
wincent, line 21: Thing about calling `getMergedConfig` twice, both times with a second param, is that it will throw if the key is not supplied. It seems fine for `build` to be required, but `federation` might be better as an optional key, seeing as absent is the same as `false`. (Especially because in DXP there are modules that don't use the preset, so you'll have to add an explicit `federation` key to all of them, or to the alternative preset that they do use.)

So maybe:

```js
const CONFIG = getMergedConfig('npmscripts');

if (!CONFIG.build) {
  throw new Error('npmscripts.config.js is missing required "build" key');
}
```

and from then on just use `CONFIG.build` and `CONFIG.federation`.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -98,11 +96,11 @@ module.exports = async function (...argv) {
...
 	);
 
-	if (fs.existsSync(path.join(CWD, 'webpack.config.js')) || args.federation) {
-		webpack(...argv);
+	if (fs.existsSync('webpack.config.js') || FEDERATION_ENABLED) {
+		webpack(...args);
 	}
 
-	runBundler(args.federation ? 'bundler-with-federation' : 'bundler');
+	runBundler();
```
wincent, line 103: Noice.

nice-jazz-club

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
```diff
@@ -76,7 +67,7 @@ function withWebpackConfig(filename, {federation}, callback) {
 		module.exports = tweakWebpackConfig(
 			webpackConfig,
 			{
-				federation: ${federation}
+				federation: ${FEDERATION_ENABLED}
```
wincent, line 71: Maybe `!!FEDERATION_ENABLED`.

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
```diff
@@ -18,7 +18,12 @@ const fs = require('fs');
 function parseBnd(filePath = './bnd.bnd') {
 	let content = fs.readFileSync(filePath).toString();
 
+	// Convert Windows line feeds to Unix
+
 	content = content.replace(/\r\n/g, '\n');
+
+	// Handle line continuation (`\` at the end of a line)
+
```
wincent, line 26: I don't know why you made this so readable when it could have been obfuscated.

```js
const lines = fs
    .readFileSync(filePath, 'utf8')
    .replace(/\\\r?\n/g, ' ')
    .split(/\r?\n/);
```

`#LostOpportunties`

projects/npm-tools/packages/npm-scripts/support/getFixture.js
```diff
@@ -4,16 +4,14 @@
...
+const getFixturePath = require('./getFixturePath');
+
 const readFile = util.promisify(fs.readFile);
 
 async function getFixture(fixture) {
-	return await readFile(
-		path.join(__dirname, '../__fixtures__', fixture),
-		'utf8'
-	);
+	return await readFile(getFixturePath(fixture), 'utf8');
```
wincent, line 14: I know this was like this before, but you don't need `async` or `await` here. From the callers' perspective, these functions are identical promise-returning functions:

```js
async function getFixture1(fixture) {
    return await readFile(getFixturePath(fixture), 'utf8');
}

function getFixture2(fixture) {
    return readFile(getFixturePath(fixture), 'utf8');
}
```

And you can await on both in exactly the same way:

```js
await getFixture1('blah');
await getFixture2('blah');
```

projects/npm-tools/packages/npm-scripts/support/getFixturePath.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const path = require('path');
+
+function getFixturePath(...fixture) {
+	return path.join(__dirname, '../__fixtures__', ...fixture);
```
wincent, line 9: While we're at it, should we change `'../__fixtures__'` to `'..', '__fixtures__'`?

I know it works (and tests on Windows in CI prove it), but maybe to be consistent?

projects/npm-tools/packages/npm-scripts/test/scripts/theme.js
```diff
@@ -7,30 +7,29 @@ const os = require('os');
...
+
+const FIXTURES = getFixturePath('scripts', 'theme');
+const MODULES = path.join(FIXTURES, 'modules');
+const APPS = path.join(MODULES, 'apps');
+const NODE_MODULES = path.join(MODULES, 'node_modules');
+const FJORD = path.join(APPS, 'frontend-theme-fjord', 'frontend-theme-fjord');
+const CLASSIC = path.join(APPS, 'frontend-theme', 'frontend-theme-classic');
+const BAD = path.join(FIXTURES, 'not', 'the', 'modules', 'you', 'are', 'looking', 'for');
+const FRONTEND = path.join(APPS, 'frontend-theme');
+const STYLED = path.join(
```
wincent: I hope these still work. I don't remember exactly why, but that comment about "Use path.normalize to make tests behave uniformly on Linux and Windows" makes me think I had to add to use `normalize` here for some reason. 🤷

projects/npm-tools/packages/npm-scripts/test/utils/parseBnd.js
```diff
@@ -3,25 +3,14 @@
...
-			'utils',
-			'parseBnd',
-			'bnd.bnd'
-		);
-	}
+	const bndPath = getFixturePath('utils/parseBnd/bnd.bnd');
 
-	it('correctly parses a bnd.bnd file', () => {
-		const entries = parseBnd(getBndPath());
+	it('correctly parses a bnd.bnd file', async () => {
```
wincent, line 12: Why `async`? I can't see any `await` in here.

wincent:
> Pushed changes to address revision issues. However I will check this again running `ant all` locally because I've changed so many things. 🤞

Looks pretty good to me!

projects/npm-tools/packages/npm-scripts/test/utils/parseBnd.js
izaera: Oops, that's a leftover because I first tried with `getFixture`. 
I thinks there's an eslint rule for that :thinking: ...

projects/npm-tools/packages/npm-scripts/test/scripts/theme.js
izaera: We are going to see it when CI finishes :sweat_smile: . The theory is that it should work, but with paths you never know...

When I saw the code I strongly empathized with the person behind it :-D. It reminded me of my adventures with path in the bundler.

projects/npm-tools/packages/npm-scripts/support/getFixturePath.js
izaera: Yes. We should never use '/'  in any place, just `path.join` with multiple args. I learned that the hard way after several iterations in the bundler.

The root cause of the problem seems to be that we assume that we can convert paths from posix to windows by changing `/` and `\` when that's impossible because it is not a bijective mapping :-(.

projects/npm-tools/packages/npm-scripts/src/utils/parseBnd.js
izaera: :thinking: Maybe the source formatter... I'm sure I wouldn't have done that on purpose.

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
izaera: It's false by default in the preset, but it is true that the overrides may turn it `undefined` :+1: .

izaera:
As expected, this has broken portal's build :roll_eyes: 

broken

wincent:
> As expected, this has broken portal's build 🙄

Well, the good news is that you found out quickly. 😀

izaera:
Working again. I will push new changes to address the new review so that I can break it again, refix and get caught in an [infinite loop forever](https://en.wikipedia.org/wiki/Eternal_return).

izaera:
This is ready to ship, once it is green, if no one has anything else to say...

Issue #253, opened 2020-11-27
Add "@function", "@return" to list of accepted stylelint at-rules


wincent:
### Issue type

-   :gift: Feature request

### Description

Came up in [this Slack thread](https://liferay.slack.com/archives/CNBG06JS3/p1606416319333100).

We decided to add the `at-rule-no-unknown` rule [here](https://github.com/liferay/liferay-npm-tools/issues/344#issuecomment-576756682) and customized the whitelist of allowed rules [here](https://github.com/liferay/liferay-npm-tools/pull/347/commits/be48c275d5377a1b1f1a0cb34a37a1c807c1846c) based on the built-in rules that come with Sass and that were currently in use in DXP at the time.

@eduardoallegrini [says](https://liferay.slack.com/archives/CNBG06JS3/p1606466305337500?thread_ts=1606416319.333100&cid=CNBG06JS3) that `@function` is kosher to add to the list, although the truth is we may not need it in many places. Right now `master` has exactly one usage and one corresponding lint suppression.

Probably no need to take any action with respect to [the other `at-rules`](https://sass-lang.com/documentation/at-rules), although just as I typed this, I see Edu [suggesting we add `@return` as well](https://liferay.slack.com/archives/CNBG06JS3/p1606475187339400?thread_ts=1606416319.333100&cid=CNBG06JS3).

PR #254, opened 2020-11-27
feat(npm-scripts): add `@function` and `@return` to list of stylelint at-rules

wincent:
Came up in [this Slack thread](https://liferay.slack.com/archives/CNBG06JS3/p1606416319333100).

We decided to add the `at-rule-no-unknown` rule [here](https://github.com/liferay/liferay-npm-tools/issues/344#issuecomment-576756682) and customized the whitelist of allowed rules [here](https://github.com/liferay/liferay-npm-tools/pull/347/commits/be48c275d5377a1b1f1a0cb34a37a1c807c1846c) based on the built-in rules that come with Sass and that were currently in use in DXP at the time.

@eduardoallegrini [says](https://liferay.slack.com/archives/CNBG06JS3/p1606466305337500?thread_ts=1606416319.333100&cid=CNBG06JS3) that `@function` is kosher to add to the list, although the truth is we may not need it in many places. Right now `master` has exactly one usage and one corresponding lint suppression.

#### Test plan

Add a `@function` containing a `@return` to an .scss file in DXP and run linter (`yarn run checkFormat`):

    Example.scss
      7:1  error  Unexpected unknown at-rule "@function" (at-rule-no-unknown)  at-rule-no-unknown
      9:3  error  Unexpected unknown at-rule "@return" (at-rule-no-unknown)  at-rule-no-unknown

Then apply the change in this commit and re-run; see no errors.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/253

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/config/stylelint.json
+++ b/projects/npm-tools/packages/npm-scripts/src/config/stylelint.json
@@ -10,12 +10,15 @@
 					"else",
 					"extend",
 					"for",
+					"forward",
 					"function",
 					"if",
 					"include",
 					"mixin",
 					"return",
-					"warn"
+					"use",
+					"warn",
+					"while"
 				]
 			}
 		],
--- a/projects/npm-tools/packages/npm-scripts/src/config/stylelint.json
+++ b/projects/npm-tools/packages/npm-scripts/src/config/stylelint.json
@@ -10,9 +10,11 @@
 					"else",
 					"extend",
 					"for",
+					"function",
 					"if",
 					"include",
 					"mixin",
+					"return",
 					"warn"
 				]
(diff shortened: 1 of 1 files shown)


pat270:
If we end up using Dart Sass, we might see `@use` and `@forward`. We're also missing `@while`.

A list of at-rules from Sass docs for reference: https://sass-lang.com/documentation/at-rules.

wincent:
Ok, to avoid having to touch this multiple times, I'll add `@use`, `@forward` and `@while` as well.

wincent:
[GitHub actions are degraded right now](https://www.githubstatus.com/incidents/sxthxvdlljk4), but as previous run was green, going to merge this.

PR #257, opened 2020-12-02
feat(js-toolkit): RequireCapability: convert package.json version classfier to OSGi version classifer

aschaeffer:
Follow-up to PR #219

This PR also converts the version in the manifest header `RequireCapability`.

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
@@ -12,6 +12,7 @@ import path from 'path';
 
 import * as ddm from './ddm';
 import Manifest from './manifest';
+import * as osgi from './osgi';
 import * as xml from './xml';
 
 const pkgJson = project.pkgJson;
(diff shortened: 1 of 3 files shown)


maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
```diff
@@ -300,3 +302,18 @@ function getSystemConfigurationJson() {
...
+ */
+function getBundleVersionAndClassifier(pkgJsonVersion) {
+	const parts = pkgJsonVersion.split('-');
+	if (parts.length > 1) {
+		return parts[0] + '.' + parts.slice(1).join('-');
+	}
+	else {
+		return pkgJsonVersion;
+	}
+}
```
wincent, line 319: So this is a near duplicate of [these lines](https://github.com/liferay/liferay-frontend-projects/blob/fc0a3cb5f972758db6277eb1c876a20a2e7b01b7/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/manifest.ts#L33-L39), so we should move this function to somewhere common where the same logic can be used by both places. The other instance of this already has test coverage too, as a bonus. Requesting changes @aschaeffer to get your thoughts on that.

wincent (changes requested):
@izaera: another one that could be replicated for 3.x?

izaera:
> another one that could be replicated for 3.x?

I would wait for the webpack 5 federation epic to finish before doing anything in bundler 3. Because it is very possible that we have to restructure it.

wincent:
lg2m. Let me just grab this and rebase it because CI is red but I believe it will be green with a rebase.

wincent (approved):

wincent:
Included [in v2.21.0 release](https://github.com/liferay/liferay-frontend-projects/releases/tag/liferay-js-toolkit%2Fv2.21.0).

Issue #256, opened 2020-12-01
Let compression level be configured


izaera:
### Issue type

-   :gift: Feature request

### Description

We should allow user define the compression level to use when creating the JAR file. By default we are using `STORE` compression, which doesn't compress anything but this leads to huge JARs and some customers have problems with cloud deployments (see [LPP-39585](https://issues.liferay.com/browse/LPP-39585)).

We should probably use compression level 6 (a [compromise between compression speed and artifact size](https://stackoverflow.com/questions/16485573/what-exactly-is-default-compression)) by default and let the value be configurable in case someone wants to tweak it further.

**Repro instructions (if applicable):**

Create a vue project with the generator and build the JAR. It is around 8,8MB. With `DEFLATE` set to compression level 9, it gets to around 2.6MB instead.

Issue #249, opened 2020-11-24
Bring over liferay-r2 RTL package


jbalsas:
As evidenced at [POC: dart sass for building css](https://github.com/liferay/liferay-frontend-projects/pull/245#pullrequestreview-535265795) we need to bring over the [liferay-labs/r2](https://github.com/liferay-labs/r2) project into `liferay-frontend-projects`. 

The project is currently only used inside the `liferay-js-themes-toolkit` but it should also be used within `liferay-npm-scripts` once we start handling sass compilation. We could decide not to even make it a public package but simply roll it locally into the ones that need it.

An alternative option to bringin that old fork here would be to try and find more up-to-date libraries such as [rtlcss](https://github.com/MohammadYounes/rtlcss) and evaluate their readiness...

jbalsas:
Hey @bryceosterhaus, I'm assigning this to you too so we don't forget to get this done before moving any further after the Dart Sass integration!

wincent:
Funnily enough, `liferay-js-themes-toolkit` uses [a copy-pasted version of it](https://github.com/liferay/liferay-js-themes-toolkit/blob/45ca23003efba9a23e1ea75fd736a00d57390961/packages/liferay-theme-tasks/lib/r2/index.js#L14) as far as I can tell. It was inlined [in this commit](https://github.com/liferay/liferay-js-themes-toolkit/commit/a843516ffb18bbeb9b3c3b1c7de8a2622d233b12#diff-a016e05e9164926c610083bf8d96c2a4d473221ecda45c00522bdef0c5d97585).

Some relevant discussion about [licensing in this thread](https://github.com/liferay/liferay-js-themes-toolkit/pull/424#discussion_r368860354) (although GitHub might hide it) — we need to be careful to preserve [the original third-party license](https://github.com/liferay-labs/R2/blob/a4551f5bd3b22f8c2f3d2df6f3f25baa300631dd/r2.js#L2-L5). We broke the rules a bit in the themes toolkit, mostly because I was conflict averse and didn't want to keep arguing about it. But let's do it properly this time.

In a sense moving it here is only solving half of the problem. Unless we un-inline it in `liferay-js-themes-toolkit`, we will have two divergent copies floating around in our graph. Maybe we should look at moving themes toolkit in here as well — the only reason I didn't move it in the initial burst of imports was that other work came up.

[...]

PR #258, opened 2020-12-02
fix(js-toolkit): let compression level be configured

izaera:
( See issue #256 )

We should allow user define the compression level to use when creating the JAR file. By default we are using STORE compression, which doesn't compress anything but this leads to huge JARs and some customers have problems with cloud deployments (see [LPP-39585](https://issues.liferay.com/browse/LPP-39585)).

We should probably use compression level 6 (a compromise between compression speed and artifact size) by default and let the value be configurable in case someone wants to tweak it further.

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
@@ -29,14 +29,22 @@ export default function createJar() {
 	addSystemConfigurationFiles(zip);
 	addPortletInstanceConfigurationFile(zip);
 
-	return zip.generateAsync({type: 'nodebuffer'}).then((buffer) => {
-		fs.mkdirpSync(project.jar.outputDir.asNative);
+	const {compression, compressionOptions} = getCompressionConfiguration();
 
-		fs.writeFileSync(
-			project.jar.outputDir.join(project.jar.outputFilename).asNative,
-			buffer
-		);
-	});
+	return zip
+		.generateAsync({
+			compression,
+			compressionOptions,
+			type: 'nodebuffer',
+		})
+		.then((buffer) => {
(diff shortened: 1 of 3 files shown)


maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
```diff
@@ -29,14 +29,22 @@ export default function createJar() {
...
 
-		fs.writeFileSync(
-			project.jar.outputDir.join(project.jar.outputFilename).asNative,
-			buffer
-		);
-	});
+	return zip
+		.generateAsync({
+			compression,
+			compressionOptions,
```
wincent, line 37: So this will sometimes be equivalent to `compressionOptions: undefined`. Would it be better to just omit the property in that case? ie.

```js
return zip
    .generateAsync({
        ...getCompressionConfiguration(),
        type: 'nodebuffer',
    })
```

wincent (approved):
lg2m. One non-blocking question inline.

wincent (changes requested):
Temporarily requesting changes on this because I have three big subtree merges for https://github.com/liferay/liferay-frontend-projects/issues/249 that I want to get in before this (redoing them would be painful). It might take a few hours, but 🤞 nothing too long...

wincent (approved):
[Merge happened](https://github.com/liferay/liferay-frontend-projects/commits/7ac5515d77edede86ea013a187c90370d0c5dd67), so this is all clear now (I have some unrelated follow-up fixes to do tomorrow though).

maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
izaera: I sometimes wonder what goes on inside my brain... :roll_eyes:

PR #276, opened 2020-12-04
chore(monorepo): clean up orphaned Travis config files

wincent:
Three that I knew about from the recent js-themes-toolkit migration, and one from js-toolkit that had escaped my attention back when we imported that.

We're now running all CI tasks using GitHub actions under "projects/", and all that remains is to set up actions for the js-themes-toolkit projects that live under "maintenance/", but I'll leave that for another PR as it is likely to be fiddly.


wincent:
It's green.

PR #278, opened 2020-12-09
style(js-toolkit): fix formatting

wincent:
Because [Lerna mangles it](https://github.com/liferay/liferay-frontend-projects/runs/1523806083?check_suite_focus=true) whenever we release.

We're going to have to figure out a way to stop it from doing that... or we might simply hope to stop using Lerna or releasing "legacy" projects which use it... (We could ignore the file, but that defeats the purpose of having Prettier *in charge* of formatting everything in the repo.)

### DIFF
--- a/maintenance/projects/js-toolkit/lerna.json
+++ b/maintenance/projects/js-toolkit/lerna.json
@@ -9,9 +9,7 @@
 		}
 	},
 	"npmClient": "yarn",
-	"packages": [
-		"packages/*"
-	],
+	"packages": ["packages/*"],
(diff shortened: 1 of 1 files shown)


wincent:
Too trivial to bother anybody with...

izaera:
> We're going to have to figure out a way to stop it from doing that...

Given that we release [from a script](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/package.json#L28) we may simply restore the file to the desired format after running lerna... :thinking:

PR #279, opened 2020-12-09
chore(js-themes-tookit): clean up irrelevant .github directories

wincent:
These came across when we imported the history but we don't need them now (because we're covered by the config at the top level of the monorepo).


wincent:
Trivial.

Issue #280, opened 2020-12-09
Standardize test directory structure


wincent:
At the moment we have:

1.  Projects that nest tests close to the corresponding source in `__tests__/` directories:
    -   `maintenance/projects/js-themes-toolkit-v8`
    -   `maintenance/projects/js-themes-toolkit-v9`
    -   `maintenance/projects/js-toolkit`
    -   `projects/js-themes-toolkit`
    -   `projects/js-toolkit`
2.  Projects that put tests inside `test/` at the top level, maintaining a parallel hierarchy to `src/`:
    -   `projects/amd-loader`
    -   `projects/npm-tools`
3.  Projects that are unique and beautiful snow flakes and do their own thing:
    -   `projects/eslint-config`: this one uses the structure provided by [ESLint's `generator-eslint` Yeoman plugin](https://www.npmjs.com/package/generator-eslint):
        -   `$PLUGIN_NAME/lib/rules` (source files); and:
        -   `$PLUGIN_NAME/tests/lib/rules` (test files).

Projects outside of this repo that may eventually get imported here currently follow these patterns:

-   [Clay](https://github.com/liferay/clay): nested `__tests__` folders (ie. **variant 1**)
-   [liferay-ckeditor](https://github.com/liferay/liferay-ckeditor): what tests?  (ie. **variant 0**)
-   [AlloyEditor](https://github.com/liferay/alloy-editor): top-level `test` folder (ie. **variant 2**)
-   [liferay-frontend-guidelines](https://github.com/liferay/liferay-frontend-guidelines): what tests?  (ie. **variant 0**).

Numerically speaking, `__tests__` predominates, and for good reason (it's a pretty common pattern out in the broader JavaScript ecosystem, and it makes it easy to jump back and forth between implementation and test files), so moving everything in that direction would be less work overall.

On the flip side, [our guidelines recommend the use of `test/`](https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/file_names.md#why-testjs) because that's the way we do it in Liferay DXP (and the reason we do it that way there is to ensure that test files never make it into the built JAR files).

So we have a choice here and could go either way. I "like" `__tests__` better, but I also know that if we ever get asked to move the code into this repo into `liferay-portal` as part of some future hypothetical mega-monorepo-ization then it would be a difficult decision to unwind.

[...]

PR #281, opened 2020-12-09
refactor(js-themes-toolikt): hoist lint and formatting config to repo root

wincent:
Didn't have time to do this the other day when I did the import so catching up now; this just gets rid of local ESLint config and delegates the task of linting and formatting to the top-level, global scripts.

Note that I needed to expand the "files" patterns so that we correctly identify files in "__tests__" directories as places where we can use Jest globals like `it` and `describe`. We should standardize the pattern here across the entire repo, but that will involve moving a *lot* of stuff, so I'm leaving that for a PR another time. In fact, there are quite a few trade-offs to consider so I've done a braindump for that in [this issue here](https://github.com/liferay/liferay-frontend-projects/issues/280).


wincent:
This one is small enough to just go in, I think.

PR #282, opened 2020-12-09
feat(npm-scripts): add script for building sass

bryceosterhaus:
follow up from https://github.com/liferay/liferay-frontend-projects/pull/245

Few notable changes:
- I removed RTL support since we handle that via a filter in portal rather than building the files ahead of time. We can add this down the road, but figured its best to get the basics in first.
- Supports adding timestamps for `@import url()`
- Only re-builds sass if the scss source files have been modified
- Moved it to its own script rather than including it in build. This would allow for easier integration of modules that don't use npm-scripts already. I am thinking of `frontend-css-web`.

Next steps:
- Review code
- Publish a release (possibly just a preview release)
- Update a few modules in portal for proof of concept.

Updates to a module would be adding `liferay-npm-scripts sass` to the build script and then updating the build.gradle to be
```groovy
// from this
buildCSS {
	imports = [
		new File(npmInstall.nodeModulesDir, "@clayui/css/src/scss")
	]
}

// to this
copyCSS.onlyIf {false}
buildCSS.onlyIf {false}
```

note: I'll update commit history once we think this is a good enough place to merge


projects/npm-tools/packages/npm-scripts/package.json
```diff
@@ -76,6 +77,9 @@
 	"license": "BSD-3-Clause",
 	"main": "index.js",
 	"name": "@liferay/npm-scripts",
+	"devDependencies": {
+		"@clayui/css": "^3.x"
+	},
```
wincent, line 82: We only want `devDependencies` at the root of the repo (I have a few of those to clean up in the recently imported js-themes-toolkit, but the other projects are all free of `devDependencies`).

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,181 @@
...
+
+	'**/_diffs/**',
+	'**/.sass-cache*/**',
+	'**/.sass_cache_*/**',
+	'**/_sass_cache_*/**',
+	'**/_styled/**',
+	'**/_unstyled/**',
+	'**/css/aui/**',
+	'**/css/clay/**',
+	'**/tmp/**',
```
wincent, line 27: Doesn't look fully sorted?

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,181 @@
...
+];
+
+const CWD = process.cwd();
+
+const CSS_IMPORT_REGEX = /@import\s+url\s*\(\s*['"]?(.+\.css)/g;
```
wincent, line 32: Probably more precise to do a non-greedy match here. Your regex would match:

```
@import url("a.css"); @import url("b.css");
^--------------------------------------^ match
             ^-------------------------^ capture group
```

ie. change it to:

```
/@import\s+url\s*\(\s*['"]?(.+?\.css)/g
                              ^
                              |
                      note the "?" here
```

projects/npm-tools/packages/npm-scripts/src/scripts/sass.js
```diff
@@ -0,0 +1,45 @@
...
+const getMergedConfig = require('../utils/getMergedConfig');
+const setEnv = require('../utils/setEnv');
+const validateConfig = require('../utils/validateConfig');
+
+const {build: BUILD_CONFIG} = getMergedConfig('npmscripts');
+const CWD = process.cwd();
+
+if (!BUILD_CONFIG) {
+	throw new Error('npmscripts.config.js is missing required "build" key');
+}
```
wincent, line 18: IIRC you can skip this check by writing

```js
const BUILD_CONFIG = getMergedConfig('npmscripts', 'build');
```

instead, due to [this code](https://github.com/liferay/liferay-frontend-projects/blob/b9a5a054c3d91a10f0881a49c130eed9f3e015dc/projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js#L24-L26) which we added in 7b5f5376d0727d2e38fa7bd. It should `throw` if the `build` property is missing.

projects/npm-tools/packages/npm-scripts/test/sass/build.js
```diff
@@ -0,0 +1,265 @@
...
+	beforeAll(() => {
+		tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'scripts-'));
+
+		tempSassDir = path.join(tempDir, OUTPUT_SASS_DIR);
+	});
+
+	afterEach(() => {
+		log.mockReset();
+
+		rimraf.sync(tempDir);
```
wincent, line 32: You don't need this. One of the main reasons to use `os.tmpdir()` is that the OS takes care of cleaning things up at some opportune time (usually after a reboot, but it varies by OS; some will reclaim space on a schedule, or when space runs low, I believe) — this in turn means that the tests run faster because they do less work.

projects/npm-tools/packages/npm-scripts/src/utils/expandGlobs.js
```diff
@@ -9,6 +9,7 @@ const path = require('path');
 const getRegExpForGlob = require('./getRegExpForGlob');
 
 const DEFAULT_OPTIONS = {
+	baseDir: '.',
```
wincent, line 12: Would be nice to see these changes covered in [the tests](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/packages/npm-scripts/test/utils/expandGlobs.js).

wincent (commented):
> * Moved it to its own script rather than including it in build. This would allow for easier integration of modules that don't use npm-scripts already. I am thinking of `frontend-css-web`.

One trade-off to bear in mind here:

- Adding a new command will require changes to the Java Source Formatter, which expects us to always say `liferay-npm-scripts build` (plus a couple of other permitted ones: `liferay-npm-scripts webpack` and `liferay-npm-scripts theme build`).
- There is some value in making _all_ modules have the same build entry point; those modules which don't use `@liferay/npm-scripts` could start saying `liferay-npm-scripts build` instead of `liferay-npm-scripts build-variation-3` (just thinking about a future world where we have `liferay-npm-scripts build-variation-11` etc). Basically, trying to manage the complexity of configuring build tooling _inside_ `@liferay/npm-scripts` intead of strewn across the `package.json` files of the liferay-portal repo; we already have build commands... they grow like weeds.

> Next steps:
> 
> * Review code

I left a few small comments but this is looking good.

> * Publish a release (possibly just a preview release)

If you go with adding a new subcommand, then maybe you can skip the preview (because it is an almost purely additive change).

In general about previews in this project; we've done it two ways so far:

- Merge to `master` and release: it doesn't matter whether its a prerelease or not because if you break it in a prerelease, `master` is still broken, so you still have to fast-follow with a fix, or revert. Note that the value of communicating the status/quality of the release by using a pre-release designation is also diminished in this project, because nobody uses it but us, so we're not really "telling" anybody anything at all.
- Cut a prerelease before merging, and use a non-`latest` dist-tag on npm. This is good for risky changes because you keep master in a good state until you've proven that things are good. The cost is that releasing a "real prerelease" like this is [a bit more manual](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/CONTRIBUTING.md#publishing-a-preview-release) — we could look at tweaking `@liferay/js-publish` to make it a bit less manual in the future though.

> * Update a few modules in portal for proof of concept.

🥳

projects/npm-tools/packages/npm-scripts/src/sass/build.js
bryceosterhaus: I copied this over from `CSSBuilderUtil.java`

```java
	private static final Pattern _cssImportPattern = Pattern.compile(
		"@import\\s+url\\s*\\(\\s*['\"]?(.+\\.css)");
```

Assuming my java->js conversion was accurate, do you think its worth changing this?

projects/npm-tools/packages/npm-scripts/test/sass/build.js
bryceosterhaus: @wincent I ended up needing to clear the directory because I didn't want built css files to hang around between tests. If we remove this, we start to run into issues where a built file already exists.

If the tmp directory was cleared between tests we wouldn't need this, but it only cleans the tmp directory after all the tests run.

bryceosterhaus:
> There is some value in making all modules have the same build entry point;

My concern for not throwing this into `build` was for modules that either either had only js or only css. For an only css module(frontend-css-web), I didn't want to waste time by checking for js files. Same goes for css, I thought we could reduce build time by not checking for css unless explicitly specified. This isn't really an issue for a single module, but if we run this over 20+ modules via `ant all`, I thought it might start to add unnecessary time.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
wincent: Yes, that's why I said it.

The Java pattern obviously works, but it is fragile, because it makes assumptions about the input being well formed. By adding a single `?` character, you can make the regex more robust and handle badly formed input as well, so why not do it?

In an ideal world, we'd use AST-based transforms for everything and we would never have to worry about any of this. But in the actual world were we still use regexes for a lot of things, it's just a matter of best practice to tighten them up — in so far as is practical — to match what you want them to match and _not_ what you don't want them to match.

projects/npm-tools/packages/npm-scripts/test/sass/build.js
wincent: Ah sorry, I missed that you weren't using `mkdtemp` as well in a `beforeEach` instead of `beforeAll`, like you can [see here](https://github.com/liferay/liferay-frontend-projects/blob/ed8a348fe0bbad3ac8bf28ddd5074b56f9bdd64b/projects/npm-tools/packages/npm-scripts/test/scripts/format.js#L35) in this repo. Pristine set of files for each test and you can forget about cleanup and the need to use a dependency.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
bryceosterhaus: Sounds good to me, just wanted to clarify and make sure we were good with making those sort of changes while "porting" over to new tech. I'll be sure to include it in a separate commit so we can as least have a reference to a change in the regex in case there are issues in the future.

projects/npm-tools/packages/npm-scripts/test/sass/build.js
bryceosterhaus: Ah, that makes sense to create a temp before each test, my brain seemed to skip the idea that each tmpdir would be different if I did it beforeEach. Thanks for clarifying!

bryceosterhaus:
@wincent updated and squashed commit history.

I added the updates from your comments.

I removed the sass script and added it to build, as well as a flag within our npmscripts build config to disable the sass compilation if necessary. The intent is to not keep this flag in the config for long term and is only for the migration process. I did this because I want to be able to quarantine a specific module if we are struggling to get the modules css built. This way we can get this feature into DXP quickly and migrate over without having to wait on any outlier modules.

One thing to note, we may need to update the build script to only run babel and the bundler if `js` files exist, that way we don't waste time running it in purely css modules.


Next Steps:
- Going to leave this PR in draft state
- Cutting a preview release of npm-scripts
- Going to send a POC PR for dxp with the preview release.

The goal is once the PR for dxp looks good, we can then cut a full release of npm-scripts and merge it relatively easy.

projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -0,0 +1,178 @@
...
+const expandGlobs = require('../utils/expandGlobs');
+const log = require('../utils/log');
+
+const SASS_EXCLUDE = [
+	'**/_diffs/**',
+	'**/_sass_cache_*/**',
+	'**/_styled/**',
+	'**/_unstyled/**',
+	'**/.sass_cache_*/**',
+	'**/.sass-cache*/**',
```
wincent, line 19: Unless my eyes deceive me, line 19 makes line 18 unnecessary.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -70,6 +71,19 @@ function runBridge() {
...
 
+// Utility for getting paths to @clayui/css variables
+// This shouldn't ever fail, but is necessary so that we don't require
+// '@clayui/css' as a dependency in this package.
+
+const getClayPaths = () => {
+	try {
+		return require('@clayui/css').includePaths;
+	}
+	catch (e) {
```
wincent, line 82: I was going to say [don't abbreviate](https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/naming.md#avoid-abbreviations), but you can just delete the whole `(e)`, I think, because AFAIK, [optional catch binding](https://v8.dev/features/optional-catch-binding) works in Node 10, which is the minimum version we're using, realistically.

projects/npm-tools/packages/npm-scripts/test/sass/build.js
```diff
@@ -0,0 +1,262 @@
...
+		buildSass(path.join(FIXTURES, 'partial'), {
+			outputDir: tempDir,
+		});
+
+		expect(fs.existsSync(path.join(tempSassDir, '_some-partial.css'))).toBe(
+			false
+		);
+	});
+
+	xit('builds with rtl support', () => {
```
wincent, line 186: Just to double-check here, plan is to turn this back on via the code from liferay-r2 that's now in liferay-theme-tasks? Is that right?

As you probably saw, I imported liferay-js-theme-toolkit into this repo now, and all we have to do, I think, is officially expose the functions [in this file](https://github.com/liferay/liferay-frontend-projects/blob/22ab943c0663e04da755ad4cd88b433c105a8265/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/r2/liferay-r2/index.js) so that we can call them from `@liferay/npm-scripts`.

Seeing as I moved stuff around in the import, I was going to do a test release anyway just to make sure I didn't break anything.

projects/npm-tools/packages/npm-scripts/test/utils/expandGlobs.js
```diff
@@ -220,4 +220,22 @@ describe('expandGlobs()', () => {
...
+			path.join(os.tmpdir(), 'scripts-basedir-')
+		);
+
+		const tempFilePath = path.join(randomTempDir, 'test.js');
+
+		fs.writeFileSync(tempFilePath, '');
+
+		const doesntMatch = expand(['*']);
+
+		expect(doesntMatch).not.toEqual([tempFilePath]);
```
wincent, line 235: I find the double-negative here ("does not match not equal") hard to read. Might be easier as:

```js
let files = expand([*]);

expect(files).not.toInclude(tempFilePath);

files = expand(['*'], [], {baseDir: randomTempDir});

expect(files).toEqual([tempFilePath]);
```

bryceosterhaus:
update:

Trying new preview release with bourbon dependency today.

Also updated some gradle plugins to not run the buildCSS task if there is a `liferay-npm-script build`.

I'll follow up with a portal PR if things start to work... 🤞

projects/npm-tools/packages/npm-scripts/test/sass/build.js
wincent: So I cut "test" releases yesterday (not really tests; just whatever small inconsequential stuff was on `master` since the last release), although the truth is you could already prior to the release use r2 hackily just by doing (something like):

```js
const {swap} = require('liferay-theme-tasks/lib/r2/liferay-r2');
```

If we want to use it, we should just [export it properly here](https://github.com/liferay/liferay-frontend-projects/blob/6fb97ab7a80dfd9d87a31c1d6c9dfb9a1c606b8a/projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/index.js#L82-L84) to elevate it to "official/public" API. Only reason I didn't do this is because of your prior comment somewhere which leaves me uncertain of when we want to do it:

> I removed RTL support since we handle that via a filter in portal rather than building the files ahead of time. We can add this down the road, but figured its best to get the basics in first.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -130,6 +144,13 @@ module.exports = async function (...args) {
 		cleanSoy();
 	}
 
+	if (!BUILD_CONFIG.disableSass) {
+		buildSass(path.join(CWD, BUILD_CONFIG.input), {
```
wincent, line 148: Hey @bryceosterhaus, we can wrap `buildSass` in a call to `instrument` to get build-time performance metrics for it.

Basically, you use `let` when you `require` it ([like this](https://github.com/liferay/liferay-frontend-projects/blob/7a869cded3aed96f3bd72b22b464b9f18694aca9/projects/npm-tools/packages/npm-scripts/src/scripts/build.js#L12)), and then add `buildSass` to [the list of instrumented functions](https://github.com/liferay/liferay-frontend-projects/blob/7a869cded3aed96f3bd72b22b464b9f18694aca9/projects/npm-tools/packages/npm-scripts/src/scripts/build.js#L26-L48).

Demo and sample results of this in [the PR that added instrumentation](https://github.com/liferay/liferay-frontend-projects/pull/299) (and [example analysis here](https://docs.google.com/spreadsheets/d/1gV6ZDO2lkkQSd4yAf5fXM5xDwSNc3SOtO0ywORfwbh4/edit)).

bryceosterhaus:
Updated script to check for js files before running babel or bundler. I was running into some errors in some modules with either

`Error: Loader 'undefined' failed: Cannot determine web context path for the project: either specify it in .npmbundlerrc or bnd.bnd`

or

`src/main/resources/META-INF/resources does not exist`


These were for modules that had scss but no js and where I added a package.json and build script

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -82,6 +83,13 @@ module.exports = async function (...args) {
 		'liferay-npm-scripts: `build`'
 	);
 
+	const hasJS =
+		fs.existsSync(BUILD_CONFIG.input) &&
+		!!expandGlobs(
+			['*.js', '*.jsx', '*.ts', '*.tsx'],
+			['build', 'classes', 'node_modules']
```
wincent, line 90: Should you be anchoring this search in `BUILD_CONFIG.input`? Just in case there is some module out there that has a top-level `.js` file (eg. a config file?)

Kind of sad that we have to do this, and that the two lists here have to be so hard-coded (well, really, the hard-coding of the second list is the one that is a bit sad).

I wonder how feasible it is to fix this "upstream". For example, maybe the bundler could be taught to handle the no-JS case gracefully/silently instead of:

```
Error: Loader 'undefined' failed: Cannot determine web context path for the project: either specify it in .npmbundlerrc or bnd.bnd
```

@izaera: thoughts?

About Babel, presumably that's the one that is printing this:

```
src/main/resources/META-INF/resources does not exist
```

Here again I wonder exactly where that error is coming from and whether we could just teach it to be graceful. 🤔

izaera: The error about the web context is thrown from `liferay-npm-bundler-loader-css-loader` when it tries to generate the `.js` module for a `.scss` file.

I would say there's no way to fix that upstream since you are calling the bundler in a project that is not meant to call the bundler so, even if we ignore that, we may get into lots of unpredictable problems. Also, what's the point of running a tool that, in the best case, it's going to do nothing? :sweat_smile: 

Maybe we can detect this by looking for a `bnd.bnd` file for example, in addition to `.js` files. That would fix the bundler problem.

Regarding babel, I suppose we only need to check if `src/main/resources/META-INF/resources` exists. Given that it is given as configuration to `npm-scripts` we don't have to hardcode it, I think (I think that is inside `BUlLD_CONFIG.input`...).

izaera: Yes, it is:
image

wincent: > Maybe we can detect this by looking for a `bnd.bnd` file for example, in addition to `.js` files. That would fix the bundler problem.
>
> ...
>
> I suppose we only need to check if `src/main/resources/META-INF/resources` exists.

This sounds better than recursively traversing an entire directory hierarchy, if looking at the `bnd.bnd` alone is enough... basically just looking for a way to do the check in `O(1)` instead of `O(n = size of module)`.

bryceosterhaus: Looks like babel is happy enough with just the check if the `BUlLD_CONFIG.input` path exists.


For the bundler...

> looking for a bnd.bnd file for example, in addition to .js files.

@wincent sounds like we would still need to check for the js files _in addition_. But I think if we anchor the search for js files with `BUlLD_CONFIG.input` it should cut costs considerably.

I'll investigate this a bit more and see if we can reduce the overhead of traversing the file structure multiple times (once for the check, additionally for bundler & babel).

wincent: > But I think if we anchor the search for js files with `BUlLD_CONFIG.input` it should cut costs considerably.

Yeah, will avoid any potentially nasty nested `node_modules` subtrees that might be hiding in the module root.

projects/npm-tools/packages/npm-scripts/test/sass/build.js
wincent: Added https://github.com/liferay/liferay-frontend-projects/pull/301 which will mean you can do:

```js
const {r2} = require('liferay-theme-tasks');

r2.swap(...);
```

projects/npm-tools/packages/npm-scripts/test/utils/expandGlobs.js
```diff
@@ -220,4 +220,22 @@ describe('expandGlobs()', () => {
 			/Redundant ignore patterns/
 		);
 	});
+
+	it('can match files in a different directory', () => {
+		const randomTempDir = fs.mkdtempSync(
+			path.join(os.tmpdir(), 'scripts-basedir-')
+		);
+
+		const tempFilePath = path.posix.join(randomTempDir, 'test.js');
```
bryceosterhaus, line 229: Anyone know why it is necessary to use `posix` here? Windows was failing earlier with this error, but adding posix seemed to fix it. However, looking at the output, I'm not totally sure why it only mattered for that last directory path.

```
    - Expected  - 1
    + Received  + 1

      Array [
    -   "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\scripts-basedir-0RbmjO\\test.js",
    +   "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\scripts-basedir-0RbmjO/test.js",
      ]
```

bryceosterhaus: Actually, I don't think its right to use `path.posix` because now I am just making the expected value to use both `\\`  and `/`.

cc @wincent @jbalsas any ideas?

izaera: It doesn't make sense to use `path.posix` there. In fact, it almost never makes sense unless you are reading posix paths from a configuration or something like that (in which case, I usually resort to string functions `split` and `join` passing `/` directly to make it explicit that we are not dealing with paths, but with a multi-platform logical representation).

Nevertheless, Windows APIs are POSIX, so you can use posix paths for Windows too (though it may lead to problems because some older APIs in libc are not POSIX and you never know for sure what APIs the applications are using).

Said that, I don't know why using `path.posix` there fixes the test. My bet is that the underlying functions you are using for expanding globs are mixing POSIX and Windows paths. For example: they can be taking the paths in the format you pass them but then completing paths with POSIX, or something like that...

wincent: > My bet is that the underlying functions you are using for expanding globs are mixing POSIX and Windows paths.

That's a new behavior. We have always expressed globs using POSIX format, and when expanded them we have always returned them in expanded form [in the same format](https://github.com/liferay/liferay-frontend-projects/blob/19e03ab7adbaf20d7ad23b3f6235bb18ab9430a2/projects/npm-tools/packages/npm-scripts/src/utils/expandGlobs.js#L130).

The difference is that in this PR @bryceosterhaus has taught `expandGlobs()` to take `baseDir` param. So, previously, we would always operate relative to the current working directory, which meant that given a glob like `**/*.js` we would expand that to things like `a/b.js`, `a/c.js` etc regardless of platform and everything was rosy. But with `baseDir` in effect, on Windows the glob now expands to `SOME_WINDOWS_PATH\a/b.js` etc.

So, we can do any of these:

- Call it harmless and leave the test like it is (with the `path.posix` call), possibly with a comment explaining why:

  ```
  // Note that globs are always expressed in (and expanded to) POSIX form.
  const tempFilePath = path.posix.join(randomTempDir, 'test.js');
  ```
- Back out the change that added `baseDir` support and instead `process.chdir()` in `collectSassFile()`; might need to do a `path.join()` in there if we definitely need absolute paths (I have no idea if we do).
- Change `expandGlobs()` to stop using `path.posix.join()` (but note that it's been like that [for two years now](https://github.com/liferay/liferay-frontend-projects/commit/5e76049d782b85614c274ffd95d064a1ba6c8332), and that commit flags at least one more thing we'd need to change in order to keep things working on Windows if we do that).

My preference is to do the minimal thing (ie. the first option) just to minimize the risk of regressions.

bryceosterhaus:
Squashed commit history. Tests look good now, just waiting on the PR to portal to pass before merging this and making a release https://github.com/liferay-frontend/liferay-portal/pull/617

bryceosterhaus:
Tests look to be passing in https://github.com/liferay-frontend/liferay-portal/pull/617. This is ready for a final review

bryceosterhaus:
For the sake of time and avoiding delays in getting a real PR to DXP, I am gonna merge this and cut a new release.We've addressed most comments above and if anything else comes up, we can make more adjustments or revert if something drastically breaks. 🤞

PR #283, opened 2020-12-10
docs(js-themes-toolkit): update links to point at new repo

wincent:
Labeling this as "docs" although there is one code change in there. This commit updates all references that were pointing at the old (archived, read-only) repo to point at the monorepo instead.

Bonus: soft-wrapped the v8 CONTRIBUTING.md because that's what we're doing in all the other Markdown files.


wincent:
Basic stuff.

PR #284, opened 2020-12-10
feat: extend webpack federation support

izaera:
This PR adds newer features to `npm-scripts` which are needed to implement [LPS-124287](https://issues.liferay.com/browse/LPS-124287).

I have tested it locally with the LPS-124287 PR for portal and everything work, but we need to merge this and release a new version for the other PR to work.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
@@ -17,7 +17,7 @@ const spawnSync = require('../utils/spawnSync');
 const validateConfig = require('../utils/validateConfig');
 const webpack = require('./webpack');
 
-const {build: BUILD_CONFIG, federation: FEDERATION_ENABLED} = getMergedConfig(
+const {build: BUILD_CONFIG, federation: FEDERATION_CONFIG} = getMergedConfig(
 	'npmscripts'
 );
 const CWD = process.cwd();
@@ -73,7 +73,7 @@ function runBridge() {
 /**
  * Main script that runs all all specified build tasks synchronously.
  *
- * Babel and liferay-npm-bundler are always run,
+ * Babel and liferay-npm-bundler are run unless the disable flag is set,
  * liferay-npm-bridge-generator and webpack are run if the corresponding
  * ".npmbridgerc" and "webpack.config.js" files, respectively, are
  * present, and soy is run when soy files are detected.
@@ -94,18 +94,31 @@ module.exports = async function (...args) {
 		buildSoy();
 	}
 
-	runBabel(
-		BUILD_CONFIG.input,
-		'--out-dir',
-		BUILD_CONFIG.output,
-		'--source-maps'
-	);
+	const disableOldBuild =
+		FEDERATION_CONFIG && FEDERATION_CONFIG.disableOldBuild;
 
-	if (fs.existsSync('webpack.config.js') || !!FEDERATION_ENABLED) {
+	if (!disableOldBuild) {
+		runBabel(
+			BUILD_CONFIG.input,
+			'--out-dir',
+			BUILD_CONFIG.output,
+			'--source-maps'
+		);
+	}
+
+	if (fs.existsSync('webpack.config.js') || !!FEDERATION_CONFIG) {
 		webpack(...args);
 	}
 
-	runBundler();
+	if (!disableOldBuild) {
+		runBundler();
+	}
+	else {
+		const {output} = BUILD_CONFIG;
+
+		fs.copyFileSync('package.json', path.join(output, 'package.json'));
+		fs.writeFileSync(path.join(output, 'manifest.json'), '{}');
+	}
 
 	translateSoy(BUILD_CONFIG.output);
 
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
@@ -106,7 +106,7 @@ module.exports = async function (...args) {
 		);
 	}
 
-	if (fs.existsSync('webpack.config.js') || !!FEDERATION_CONFIG) {
+	if (fs.existsSync('webpack.config.js') || FEDERATION_CONFIG) {
 		webpack(...args);
 	}
 
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
@@ -58,13 +58,7 @@ function withWebpackConfig(filename, callback) {
 			federation = '{}';
 		}
 		else if (typeof FEDERATION_CONFIG === 'object') {
-			federation = '{\n';
-
-			if (FEDERATION_CONFIG.main) {
-				federation += `main: '${FEDERATION_CONFIG.main}'\n`;
-			}
-
-			federation += '}\n';
+			federation = JSON.stringify(FEDERATION_CONFIG);
 		}
 	}
 
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
@@ -10,7 +10,7 @@ const getMergedConfig = require('../utils/getMergedConfig');
 const spawnSync = require('../utils/spawnSync');
 const withTempFile = require('../utils/withTempFile');
 
-const FEDERATION = getMergedConfig('npmscripts', 'federation');
+const FEDERATION_CONFIG = getMergedConfig('npmscripts', 'federation');
 const TWEAK_WEBPACK_CONFIG_PATH = require.resolve(
 	'../utils/tweakWebpackConfig'
 );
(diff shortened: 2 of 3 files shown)


projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -94,18 +94,31 @@ module.exports = async function (...args) {
...
 	}
 
-	runBabel(
-		BUILD_CONFIG.input,
-		'--out-dir',
-		BUILD_CONFIG.output,
-		'--source-maps'
-	);
+	const disableOldBuild =
+		FEDERATION_CONFIG && FEDERATION_CONFIG.disableOldBuild;
```
izaera, line 98: I couldn't find a better name. However, this should be temporary, so we don't need to be very careful with its definition, IMO.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -94,18 +94,31 @@ module.exports = async function (...args) {
...
+	if (fs.existsSync('webpack.config.js') || !!FEDERATION_CONFIG) {
 		webpack(...args);
 	}
 
-	runBundler();
+	if (!disableOldBuild) {
+		runBundler();
+	}
+	else {
+		const {output} = BUILD_CONFIG;
```
izaera, line 117: These steps are usually performed by the bundler so we need to "fake" them here.

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
```diff
@@ -51,6 +51,23 @@ function escapeLiteralString(str) {
...
 	const webpackConfigPath = path.resolve(filename);
 
+	let federation;
+
+	if (FEDERATION_CONFIG) {
+		if (typeof FEDERATION_CONFIG === 'boolean') {
+			federation = '{}';
+		}
+		else if (typeof FEDERATION_CONFIG === 'object') {
+			federation = '{\n';
```
izaera, line 61: Is there a clean way to generate a JS (not JSON) valid representation of an object?

projects/npm-tools/packages/npm-scripts/src/utils/tweakWebpackConfig.js
```diff
@@ -12,16 +12,28 @@ const createTempFile = require('./createTempFile');
...
+const CORE_SHARES = [
+	'@clayui/icon',
+	'classnames',
+	'formik',
+	'prop-types',
+	'react',
+	'react-dnd',
+	'react-dnd-html5-backend',
+	'react-dom',
+];
```
izaera, line 25: This will be moved to a better place once it's clear what we need. It is the equivalent of the `imports` section of the default preset.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
wincent: judging-you

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -94,18 +94,31 @@ module.exports = async function (...args) {
...
-	runBabel(
-		BUILD_CONFIG.input,
-		'--out-dir',
-		BUILD_CONFIG.output,
-		'--source-maps'
-	);
+	const disableOldBuild =
+		FEDERATION_CONFIG && FEDERATION_CONFIG.disableOldBuild;
+
+	if (!disableOldBuild) {
```
wincent, line 100: So you said this isn't permanent; what will the final state be?

If it's only temporary this is ok, but straight away, having a boolean that is only ever checked in negated form is a bit of a smell (eg. double negative "not disabled" instead of "enabled").

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -94,18 +94,31 @@ module.exports = async function (...args) {
...
+		runBabel(
+			BUILD_CONFIG.input,
+			'--out-dir',
+			BUILD_CONFIG.output,
+			'--source-maps'
+		);
+	}
 
-	if (fs.existsSync('webpack.config.js') || !!FEDERATION_ENABLED) {
+	if (fs.existsSync('webpack.config.js') || !!FEDERATION_CONFIG) {
```
wincent, line 109: nit: `!!` is unneeded

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
wincent: Why can't you use JSON (which is JS)?

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
```diff
@@ -68,7 +85,7 @@ function withWebpackConfig(filename, callback) {
 		module.exports = tweakWebpackConfig(
 			webpackConfig,
 			{
-				federation: ${FEDERATION_ENABLED}
+				federation: ${federation}
```
wincent, line 88: I don't know whether I would bother with the pretty printing (ie. the newlines) because when it gets interpolated here the indent is going to be off anyway...

wincent (commented):
Looks reasonable to me!

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: It's negative because the canonical way to build things has always been executing babel and bundler, so disabling it is the exception. Of course we can turn it into enable and default it to true, but it's a bit strange that you have to enable something that has always been there...

Nevertheless, the final state is that all this will be removed, because we won't need to run the bundler or babel alone.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: Oh, yes. We added it [here](https://github.com/liferay/liferay-frontend-projects/pull/252#pullrequestreview-539930843) but it was a different context. :+1:

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
izaera: :thinking: That's true. However I had issues when using JSON in the past but now that I realize it should work I don't know why :shrug: .

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
izaera: I added them just to be safe in case we add something in the future that needs line breaks (or if remove the spaces and turn it into a sum of strings). [OCD](https://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder).

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: I don't care

jbalsas: Screen Shot 2020-12-10 at 15 18 05

Doesn't even care enough to put an image

badass

projects/npm-tools/packages/npm-scripts/src/scripts/webpack.js
izaera: I've changed it into JSON.

izaera:
So, after the CI turns to green I'm going to merge this delicate piece of code, so carefully crafted.

exquisite

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
wincent: > It's negative because the canonical way to build things has always been executing babel and bundler, so disabling it is the exception.

I'm not saying you should _restructure_ the code; I'm saying your variable _name_ is bad, because literally _every time you use it_, you express it in a double-negative form. As a native English speaker, chained negatives sound "bad", and they require more cognitive effort to understand.

ie. if this were to stay in the code, the `if` would remain in exactly the same place, but you'd flip the name and drop the `!`; eg:

```js
const classicBuild = !FEDERATION_CONFIG || !FEDERATION_CONFIG.disableOldBuild;
//                                                            ^-------------^
//                                                                    |   
//                                         ignoring this for now, which is also "bad"    

if (classicBuild) {
  // etc...
```

All moot, of course, because this code will go away, but the point about double-negatives being undesirable is good to bear in mind for future occasions.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: It's GitHub, that doesn't embed it :sob:

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: Ah OK. That makes more sense, yes :+1: . 

Anyway, I'll change it in a future PR. It's not that I don't want to change the code, but every time we make any small change I have to do a full test cycle which includes an `ant all`, so unless we really need to change something before merging the PR I prefer to delay it... :grimacing: 

Also, in this case, we should be removing this soon :pray: . If we need it in the end, we will do a full refactor of the new things so that they stay coherent and clean (this is not the only temporary fix we will be merging, I suspect...

PR #288, opened 2020-12-10
fix(js-themes-toolkit): silence noisy yarn warning

wincent:
Silences:

    warning Missing name in workspace at "projects/js-themes-toolkit", ignoring.


wincent:
A change that couldn't hurt a fly.

PR #289, opened 2020-12-10
test(monorepo): adjust CI coverage

wincent:
Motiviation here is the first commit, which runs js-themes-toolkit tests for the v8 and v9 series in CI. Includes a bonus tweak in the second commit to make `yarn ci` a bit more powerful (ie. run more checks) when run in local checkouts.


wincent:
[Tests are green now](https://github.com/liferay/liferay-frontend-projects/runs/1532232336), although GitHub UI doesn't seem to be updating to reflect that.

wincent:
~Greed~ Green is good and all that.

PR #292, opened 2020-12-14
chore(js-themes-toolkit): make final tweaks for monorepo releases

wincent:
I'm just running through the CONTRIBUTING.md docs to prove that we can actually do releases from the monorepo, and this commit makes the changes I've identified as being necessary:

- Namely, we need to add a `version-tag-prefix` in the `.yarnrc` files for use by the changelog-generator.
- We need to switch to the latest not-yet-published version of the generator, because the last-published version doesn't want to run from subdirectories like these maintenance projects use (it expects to run from the repo root, or from a packages/ subdirectory).
- In order to distinguish between the two series, we need them to use distinct tag prefixes. We could teach the changelog-generator a new switch to deal with this, but I'd rather not add the complexity there for what is a decidedly odd use case (multiple projects on same branch happening to share the same prefix), so let's just use these somewhat ugly prefixes instead.

Test plan: Go through the prep to make a new 8.x release and produce [a diff like this](https://gist.github.com/wincent/d5890667390caf595d558aab4a809470). Could also do it for 9.x but I'm going to do that shortly anyway, and it should be just the same.

### DIFF
--- a/maintenance/projects/js-themes-toolkit-v8-x/.yarnrc
+++ b/maintenance/projects/js-themes-toolkit-v8-x/.yarnrc
@@ -1,1 +1,9 @@
 --publish.tag maintenance
+
+# We don't actually use `yarn version` to publish this project, but we still
+# specify the desired tag format here so that `@liferay/changelog-generator` can
+# pick it up.
+#
+# Note that prefix is based on the directory name so that
+# `@liferay/changelog-generator` distinguish it from the v9-x series.
+version-tag-prefix "js-themes-toolkit-v8-x/v"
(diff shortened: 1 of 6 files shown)


wincent:
`ci / format-check` failure can be ignored (is caused by Lerna mangling `maintenance/projects/js-toolkit/lerna.json`). I'll fix it [like I did here](https://github.com/liferay/liferay-frontend-projects/pull/2780) and make an actual issue for actually implementing a permanent fix.

wincent:
- Temporarily fix styles: https://github.com/liferay/liferay-frontend-projects/pull/294
- Issue for permanent fix: https://github.com/liferay/liferay-frontend-projects/issues/293

wincent:
Going to try out the release process, so I need to merge this.

Issue #290, opened 2020-12-11
Support for generating bridges from AMD to webpack federation 


izaera:
### Issue type

-   :gift: Feature request

### Description

Add a new feature to generate bridges from AMD to webpack federation as the ones included in [this commit](https://github.com/liferay-frontend/liferay-portal/pull/606/commits/47919370022f8ab06b421392b0f43b6fddc0c5e5).

wincent:
> * 🎁 Feature request

![it-is-a-gif](https://user-images.githubusercontent.com/7074/101931704-12d83280-3bda-11eb-9f5d-82f4f15fd63f.gif)

(gif -> gift)

izaera:
> (gif -> gift)

![oh my lord](https://media.giphy.com/media/3ov9jE3YKszddL2ak0/giphy.gif)

PR #295, opened 2020-12-14
feat(js-toolkit): teach publish.sh to revert unwanted formatting changes

wincent:
Added `set -e` to abort at the first sign of things going wrong, seeing as this commit involves pushing semi-blindly to the remote...

I inverted the sense of the remote check so that I could do the publishing + post-publish-fixing in a DRY way (and without having to stick it in a function).

Closes: https://github.com/liferay/liferay-frontend-projects/issues/293

PR #294, opened 2020-12-14
style(js-toolkit): fix formatting

wincent:
Same thing as [fixed here](https://github.com/liferay/liferay-frontend-projects/pull/278).

I've made [an issue for applying permanent fix](https://github.com/liferay/liferay-frontend-projects/issues/293).

PR #296, opened 2020-12-14
feat: support generation of bridges from AMD to webpack

izaera:
This is implementation of #290 . Needed for [this portal PR](https://github.com/liferay-frontend/liferay-portal/pull/606).

I have tested it locally and it's not horrible.

--- 

What this basically does is adding a new section inside `federation` of `npmscripts.config.js` containing the packages to bridge. So, for example, if we had this in `frontend-js-react-web`'s `npmscripts.config.js`:

```javascript
module.exports = {
	federation: {
		bridges: {
			'classnames': '__classnames',
			'formik': '__formik',
			'prop-types': '__propTypes',
			'react': '__react',
			'react-dnd': '__reactDnd',
			'react-dnd-html5-backend': '__reactDndHtml5Backend',
			'react-dom': '__reactDom'
		},
		main: 'js/index.federation.js',
	}
};
```

It would create bridges from `classnames` in the AMD loader to `classnames` in the webpack realm. The `__classnames` part is the name of the member as it is exported in the webpack's main entry point of `frontend-js-react-web`. So, that file would be `js/index.federation.js` and would contain something like:

```javascript
export * as __classnames from 'classnames';
export * as __formik from 'formik';
export {default as __propTypes} from 'prop-types';
export {default as __react} from 'react';
export * as __reactDnd from 'react-dnd';
export * as __reactDndHtml5Backend from 'react-dnd-html5-backend';
export {default as __reactDom} from 'react-dom';
```

I know we could do the variable name by convention, but:

1) It requires more code
2) (Especially important, at least for me) There are so many things by convention now, that it is easy to get lost and remember where things come from and where should they be defined, so having it explicitly configured helps a lot with this (because it reminds you what to grep for).

Also, given that this is temporary and will be removed when the epic is finished, I think this is fair enough: it covers what we need now (for example: it doesn't allow bridging explicit modules, just package default entry points, which is the only thing we need now) with minimal configuration.

For completeness, the generated bridges look like this:

```javascript
(function() {
	const getModule = window[Symbol.for('__LIFERAY_WEBPACK_GET_MODULE__')];

	getModule('frontend-js-react-web').then(
		function({__classnames}) {
			Liferay.Loader.define(
				"frontend-js-react-web$classnames@2.2.6/index", ['module'],
				function (module) {
					module.exports = __classnames;
				}
			);
		}
	);
})();
```

### DIFF
--- a/maintenance/projects/js-toolkit/lerna.json
+++ b/maintenance/projects/js-toolkit/lerna.json
@@ -9,9 +9,7 @@
 		}
 	},
 	"npmClient": "yarn",
-	"packages": [
-		"packages/*"
-	],
+	"packages": ["packages/*"],
 	"tagVersionPrefix": "liferay-js-toolkit/v",
 	"useWorkspaces": true,
 	"version": "2.22.0"
--- a/projects/js-toolkit/packages/js-toolkit-core/src/transform/js/index.ts
+++ b/projects/js-toolkit/packages/js-toolkit-core/src/transform/js/index.ts
@@ -121,8 +121,10 @@ async function mergeMaps(
 
 	const mergedMapGenerator = new SourceMapGenerator();
 
+	/* eslint-disable @typescript-eslint/await-thenable */
 	const oldMapConsumer = await new SourceMapConsumer(oldMap);
 	const newMapConsumer = await new SourceMapConsumer(newMap);
+	/* eslint-enable @typescript-eslint/await-thenable */
 
 	newMapConsumer.eachMapping((mapping) => {
 
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
@@ -9,14 +9,14 @@ const path = require('path');
 
 module.exports = function (bridges, dir) {
 	const projectPackageJson = JSON.parse(
-		fs.readFileSync('package.json', 'utf-8')
+		fs.readFileSync('package.json', 'utf8')
 	);
 
-	Object.entries(bridges).forEach(([packageName, exportName]) => {
+	for (const [packageName, exportName] of Object.entries(bridges)) {
 		const packageJson = JSON.parse(
 			fs.readFileSync(
 				require.resolve(`${packageName}/package.json`),
-				'utf-8'
+				'utf8'
 			)
 		);
 
@@ -69,7 +69,7 @@ module.exports = function (bridges, dir) {
 	getModule('${projectPackageJson.name}').then(
 		function({${exportName}}) {
 			Liferay.Loader.define(
-				"${namespacedVersionedPackageName}/${moduleName}", ['module'],
+				'${namespacedVersionedPackageName}/${moduleName}', ['module'],
 				function (module) {
 					module.exports = ${exportName};
 				}
@@ -79,5 +79,5 @@ module.exports = function (bridges, dir) {
 })();
 `
 		);
-	});
+	}
 };
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
@@ -1,0 +1,83 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const {addNamespace} = require('@liferay/js-toolkit-core');
+const fs = require('fs');
+const path = require('path');
+
+module.exports = function (bridges, dir) {
+	const projectPackageJson = JSON.parse(
+		fs.readFileSync('package.json', 'utf-8')
+	);
+
+	Object.entries(bridges).forEach(([packageName, exportName]) => {
+		const packageJson = JSON.parse(
+			fs.readFileSync(
+				require.resolve(`${packageName}/package.json`),
+				'utf-8'
+			)
+		);
+
+		const namespacedVersionedPackageName = addNamespace(
+			`${packageJson.name}@${packageJson.version}`,
+			projectPackageJson
+		);
+
+		const packageDir = path.join(
+			dir,
+			'node_modules',
+			namespacedVersionedPackageName
+		);
+
+		fs.mkdirSync(packageDir, {recursive: true});
+
+		fs.writeFileSync(
+			path.join(packageDir, 'package.json'),
+			JSON.stringify(
+				{
+					...packageJson,
+					name: addNamespace(packageJson.name, projectPackageJson),
+				},
+				null,
+				'\t'
+			)
+		);
+
+		let {main: moduleFileName} = packageJson;
+
+		moduleFileName = moduleFileName || 'index.js';
+		moduleFileName = moduleFileName.replace(/^\.\//, '');
+
+		if (!moduleFileName.toLowerCase().endsWith('.js')) {
+			moduleFileName += '.js';
+		}
+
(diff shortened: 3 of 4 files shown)


jbalsas:
Haven't looked at the code, but the explanation looks reasonable to me. 

From the portal [PR](https://github.com/liferay-frontend/liferay-portal/pull/606) I quite don't understand why we need all these:
- Federation bridge configuration
- `index.js`
- `index.federation.js`

I would expect this to actually also automate some of it, but it might just not be worth it. Hard to tell with my limited attention span 🤷

maintenance/projects/js-toolkit/lerna.json
```diff
@@ -9,9 +9,7 @@
 		}
 	},
 	"npmClient": "yarn",
-	"packages": [
-		"packages/*"
-	],
+	"packages": ["packages/*"],
```
wincent, line 12: Unrelated, so [already fixed separately here](https://github.com/liferay/liferay-frontend-projects/pull/294) along with [a proposal (ie. a PR) to stop it from happening again](https://github.com/liferay/liferay-frontend-projects/pull/295).

projects/js-toolkit/packages/js-toolkit-core/src/transform/js/index.ts
```diff
@@ -121,8 +121,10 @@ async function mergeMaps(
 
 	const mergedMapGenerator = new SourceMapGenerator();
 
+	/* eslint-disable @typescript-eslint/await-thenable */
 	const oldMapConsumer = await new SourceMapConsumer(oldMap);
 	const newMapConsumer = await new SourceMapConsumer(newMap);
+	/* eslint-enable @typescript-eslint/await-thenable */
```
wincent, line 127: What a horrible API... returns a Promise with the `new` operator 🤦‍♂️...

Seems like we might be able to avoid the ugly lint suppressions by using [`SourceMapConsumer.with()`](https://www.npmjs.com/package/source-map#sourcemapconsumerwith)?

projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
```diff
@@ -0,0 +1,83 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const {addNamespace} = require('@liferay/js-toolkit-core');
+const fs = require('fs');
+const path = require('path');
+
+module.exports = function (bridges, dir) {
+	const projectPackageJson = JSON.parse(
+		fs.readFileSync('package.json', 'utf-8')
```
wincent, line 12: By convention, [`'utf8'`, not `'utf-8'`](https://nodejs.org/api/fs.html) (don't know if it still works, but we should do what the docs do).

projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
```diff
@@ -0,0 +1,83 @@
...
+	);
+
+	Object.entries(bridges).forEach(([packageName, exportName]) => {
```
wincent, line 15: Don't forget that you rarely need `forEach()` nowadays:

```js
for (const [packageName, exportName] of Object.entries(bridge)) {
```

Avoids a little bit of function call overhead, and is more flexible in general (eg. permits early `return`).

projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
```diff
@@ -0,0 +1,83 @@
...
+		const packageJson = JSON.parse(
+			fs.readFileSync(
+				require.resolve(`${packageName}/package.json`),
+				'utf-8'
```
wincent, line 19: `'utf-8'` → `'utf8'`

projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
```diff
@@ -0,0 +1,83 @@
...
+		fs.writeFileSync(
+			path.join(packageDir, moduleFileName),
+			`
+(function() {
+	const getModule = window[Symbol.for('__LIFERAY_WEBPACK_GET_MODULE__')];
+
+	getModule('${projectPackageJson.name}').then(
+		function({${exportName}}) {
+			Liferay.Loader.define(
+				"${namespacedVersionedPackageName}/${moduleName}", ['module'],
```
wincent, line 72: Double and single quotes on this line — probably better to just stick with single.

wincent (commented):
Code looks good to me; only thing I'd definitely change is the `'utf-8'` to `'utf8'`.

I'll leave it up to you to make the call on @jbalsas' comments, which I can neither confirm nor deny. 😂

izaera:
> Federation bridge configuration

We need to know whether to automate bridge configuration or not. There's no way to tell from the code, unless we wanted to bridge everything, which is something complex that we don't need now (and hopefully :crossed_fingers: won't need ever).

>   index.js

This is the entry point for the AMD loader, which is now different from the one we had before because all it does now is to reexport the webpack module. That is, this is a bridge too, but hand-made because it's inside the project and I don't expect to have many of these. If more of them appear I will make a script to automate then.

>   index.federation.js

This is the entry point for webpack,  which is:

1. The old index.js plus
2. The exported webpack packages to be used by the bridges

Unfortunately we need two entry points because we only have one `main` field in the `package.json` file. In the end, the entry point configured here will be renamed as index.js and the other one removed (once we don't have AMD loader any more).

I agree we can automate all this, but I'm not 100% sure that it is worth doing it now, so I'm deferring it for the next LPS, where we can decide if we want to evolve this automatic configuration more. The thing is that, being temporary, I don't want to put much effort than the strictly needed and I don't mind doing some things by hand if it is faster.

Said that, if we want to change the strategy not because of efficiency, but because we see too much "garbage" in `liferay-portal` we can give it a try, though it is going to be considerably more complex than the bridging code we have developed in this PR.

izaera:
I'm thinking that we could also export the bridges from a different module (not the index one, but a new generated one) and make everything automatic. It looks like it could work...

I would say let's merge this and try it in the next LPS if we need to generate more bridges. Does it sound good? @jbalsas ?

maintenance/projects/js-toolkit/lerna.json
izaera: :+1:

projects/js-toolkit/packages/js-toolkit-core/src/transform/js/index.ts
izaera: I think I tried may alternatives when I had to spend almost one day fighting with that, because:

1. As you say the API is horrible
2. Not happy with being horrible, they didn't make the correct typings for TS, so that nobody knew :facepalm: 

It is possible that `SourceMapConsumer.with()` fixes it (or not :sweat_smile: ) but I prefer not to try it in this PR. Also, we will remove this sooner or later, so maybe we don't even need to fix it ever.

projects/npm-tools/packages/npm-scripts/src/scripts/createBridges.js
izaera: I suppose it still works for people like me :rofl:

izaera: The thing is that node's convention is quite unfortunate, because the canonical name seems to be [UTF-8](https://en.wikipedia.org/wiki/UTF-8) :shrug: 

Anyway, the docs only refer to `utf8` as you said so I'm changing it.

jbalsas:
> I would say let's merge this and try it in the next LPS if we need to generate more bridges. Does it sound good? @jbalsas ?

Sounds good to me, but I have a feeling that maybe some things might be there to stay and we don't want to drag them through time. Just address the small issues @wincent pointed out and you have my blessing 👍

PR #298, opened 2020-12-15
fix(npm-scripts): support scoped packages correctly

izaera:
I thought using liferay-js-toolkit-core for namespacing would have done this, but it turns out that [it is done at the bundler level](https://github.com/liferay/liferay-frontend-projects/blob/2da0f09cc0c885db457f4237bd1ac0b432c92bf4/projects/js-toolkit/packages/npm-bundler/src/adapt/bundler-project/write-export-modules.ts#L44), so we have to mimic it here.

I'm not releasing yet, because I have spotted a new error in the [portal's PR](https://github.com/liferay-frontend/liferay-portal/pull/606) :-(


izaera:
I'm merging this and will release a fix because it looks like the PR is OK (after checking with the team [in slack](https://liferay.slack.com/archives/G3JBR21HA/p1608032814138800)).

PR #299, opened 2020-12-15
perf(npm-scripts): add instrumentation for build (etc) performance

wincent:
Because Brian wanted to know the cost of moving minification to build time.

Of course, if we're going to measure that, we may as well measure everything, seeing as it costs about the same to measure `N` things as it does to measure `1`. In this way, we get insight into where we're spending time during the build and can identify areas for improvement.

How it works:

- You set `LIFERAY_NPM_SCRIPTS_TIMING` to in the environment and then do something. It specifies a directory where npm-scripts will write out CSV file(s) containing timing information. A good example would be '/dev/shm/npm-scripts-test' (on Linux) or `/tmp/npm-scripts-test` (on macOS).
- Internally, the `instrument()` function wraps any function that we want to measure in a wrapper that measures start and end time.
- CSV file(s) get written out right before exit.
- You can inspect the CSV files manually, or, probably more usefully, run the `support/generateTimingsReport.js` script to generate an aggregate of all of them. It gets printed to standard output, so you will want to redirect that to a file or into your clipboard for pasting into a Google sheet.

A small sample result here for running a deploy on one module:

    module,build,soyExists,runBabel,webpack,runBundler,translateSoy,minify
    global:build,12.949,0.000,0.000,0.000,0.000,0.000,0.000
    global:soyExists,0.000,0.066,0.000,0.000,0.000,0.000,0.000
    global:runBabel,0.000,0.000,0.719,0.000,0.000,0.000,0.000
    global:webpack,0.000,0.000,0.000,3.648,0.000,0.000,0.000
    global:runBundler,0.000,0.000,0.000,0.000,1.076,0.000,0.000
    global:translateSoy,0.000,0.000,0.000,0.000,0.000,0.062,0.000
    global:minify,0.000,0.000,0.000,0.000,0.000,0.000,7.219
    module:frontend-js/frontend-js-react-web.csv,12.949,0.066,0.719,3.648,1.076,0.062,7.219

**NOTE:** To get the total runtime you can't just add up all the subcomponents. If you do that, you'll end up double-counting. For example, "build" takes 12.9s in the example above, because it's actually doing multiple steps (eg. "runBabel" takes 0.7s, "runBundler" takes 1s etc). If you add up all the steps you'll get roughly the right answer, but if you add all those to "build" you'll overcount.

Test plan:

To show that this works on multiple packages and will work with an `ant all` too once I actually do that:

    support/install-packages.sh ~portal
    yarn # to stop Gradle invocation from complaining about frozen lockfile
    export LIFERAY_NPM_SCRIPTS_TIMING=/tmp/blah
    portool deploy frontend-js-web segments-web
    support/generateTimingsReport.js /tmp/blah

Output:

    module,build,soyExists,runBabel,webpack,runBundler,translateSoy
    global:build,17.438,0.000,0.000,0.000,0.000,0.000
    global:soyExists,0.000,0.078,0.000,0.000,0.000,0.000
    global:runBabel,0.000,0.000,4.196,0.000,0.000,0.000
    global:webpack,0.000,0.000,0.000,6.780,0.000,0.000
    global:runBundler,0.000,0.000,0.000,0.000,5.707,0.000
    global:translateSoy,0.000,0.000,0.000,0.000,0.000,0.072
    module:segments/segments-web.csv,4.664,0.060,1.858,0.000,2.519,0.056
    module:frontend-js/frontend-js-web.csv,12.774,0.018,2.338,6.780,3.188,0.016


wincent:
I've only lightly tested this, and I see it conflicts with `master` anyway. But I'll report back once I've gathered some `ant all` data on a couple of machines. Want to try with laptop and desktop.

wincent:
Rebased to resolve merge conflicts; noted that `runBridge` wasn't instrumented so moved that into a separate file so that it can be treated like all the others.

projects/npm-tools/packages/npm-scripts/support/generateTimingsReport.js
```diff
@@ -64,12 +63,13 @@ function main() {
 			const label = fields[0];
 			const delta = fields[3];
 
-			const ms = parseInt(delta);
+			const ms = parseInt(delta, 10);
 
-			(labels[label] = (labels[label] || 0) + ms),
-				(modules[file] = modules[file] || {
-					'~total': 0,
-				});
```
wincent: I swear I didn't write it like this... and it doesn't look like the sort of thing Prettier would do... weird...

jbalsas: Shame on you

jbalsas (approved):
Looks awesome and super juicy to me!!!

delicious

I have some doubts:

> You set LIFERAY_NPM_SCRIPTS_TIMING to in the environment and then do something. It specifies a directory where npm-scripts will write out CSV file(s) containing timing information. A good example would be '/dev/shm/npm-scripts-test' (on Linux) or /tmp/npm-scripts-test (on macOS).

How common is this and how easy would it be to calculate the proper path for example as part of the build configuration in DXP compared to having a default-ish location?

> To show that this works on multiple packages and will work with an ant all too once I actually do that[...]

So, just to confirm my understanding here... with timings enabled we can aggregate time spent in the same task for different invocations.

How do we tell full invocations apart, though? How does it differentiate from `ant all` to `ant all`? Would it be necessary to clean up the file?

wincent:
> I have some doubts:
> 
> > You set LIFERAY_NPM_SCRIPTS_TIMING to in the environment and then do something. It specifies a directory where npm-scripts will write out CSV file(s) containing timing information. A good example would be '/dev/shm/npm-scripts-test' (on Linux) or /tmp/npm-scripts-test (on macOS).
> 
> How common is this and how easy would it be to calculate the proper path for example as part of the build configuration in DXP compared to having a default-ish location?

The idea is that:

- We probably want to run this only manually, at least at first.
- It's desirable to make it flexible in order to take advantage of platform differences; for example, on Linux `/dev/shm` gives us convenient access to a RAM-backed (ie. fast) filesystem, so it makes sense to use that, whereas on macOS setting up a RAM-disk is [a pain in the backside](https://stackoverflow.com/a/2033417/2103996).
- Given that `ant all` will probably actually have several `npm-scripts` jobs in flight in parallel, we want each one to be able to write to its own file so that they aren't clobbering each other — this is why we specify a _directory_ and then write separate CSV files into it named after the modules where `npm-scripts` is running, followed by the script you run at the end to aggregate it all into a single CSV report file.

> 
> > To show that this works on multiple packages and will work with an ant all too once I actually do that[...]
> 
> So, just to confirm my understanding here... with timings enabled we can aggregate time spent in the same task for different invocations.

Correct, the sample is tiny (two modules), but you can see the entries prefixed with `global:` correspond to time spent across _all_ modules on `build`, on `runBundler`, on `runBabel` etc. And then you see entries prefixed with `module:` that show the time scoped to a specific module.

At least, that's what it is supposed to do — 😂

> How do we tell full invocations apart, though? How does it differentiate from `ant all` to `ant all`? Would it be necessary to clean up the file?

A "full invocation" like an `ant all` will have a bunch of `module:` entries (many dozens, I guess, corresponding to all the modules which use `npm-scripts` during the build). In practice, if you want to compare two runs you'd do something like:

```
export LIFERAY_NPM_SCRIPTS_TIMING=/dev/shm/dev-ish
ant all
path/to/npm-scripts/support/generateTimingsReport.js "$LIFERAY_NPM_SCRIPTS_TIMING" > minification-off.csv

export LIFERAY_NPM_SCRIPTS_TIMING=/dev/shm/prod-ish
git clean -fdx
ant all
path/to/npm-scripts/support/generateTimingsReport.js "$LIFERAY_NPM_SCRIPTS_TIMING" > minification-on.csv
```

Assuming you stick the report files in `/dev/shm` or `/tmp` the OS will reclaim the space later on, so there isn't any clean-up to do really. If you want to re-use the same directory nothing is stopping you, but doing an `rm -r` of it first is more likely to be deterministic (for example, if you build subset A then subset B using the same `LIFERAY_NPM_SCRIPTS_TIMING` directory without a clean in the middle, your aggregrate report will include the sum of A and B).

Anyway, this is all theory. I am going to test it out in full shortly. 😬

jbalsas:
> Assuming you stick the report files in /dev/shm or /tmp the OS will reclaim the space later on, so there isn't any clean-up to do really. If you want to re-use the same directory nothing is stopping you, but doing an rm -r of it first is more likely to be deterministic (for example, if you build subset A then subset B using the same LIFERAY_NPM_SCRIPTS_TIMING directory without a clean in the middle, your aggregrate report will include the sum of A and B).

I was wondering if there was a way to annotate the individual csvs or create intermediate folders like rolling logs in a way that we could aggregate them automatically, but it certainly looks challenging to determine which are related... maybe by getting the `pid` of the parent process if available or something... probably not worth it 🤔

wincent:
> I was wondering if there was a way to annotate the individual csvs or create intermediate folders like rolling logs in a way that we could aggregate them automatically, but it certainly looks challenging to determine which are related... maybe by getting the `pid` of the parent process if available or something (would need to be the one forking the node process, not the node process itself)... probably not worth it 🤔

Yeah, I think I'd want that handled up in the calling process; eg. top-level Gradle or Ant would set a `LIFERAY_NPM_SCRIPTS_TIMING` at a "standard" location like `/$PREFIX/npm-scripts-$TIMESTAMP`, and you'd wind up with a catalog of subdirectories corresponding to different runs.

jbalsas:
> Yeah, I think I'd want that handled up in the calling process; eg. top-level Gradle or Ant would set a LIFERAY_NPM_SCRIPTS_TIMING at a "standard" location like /$PREFIX/npm-scripts-$TIMESTAMP, and you'd wind up with a catalog of subdirectories corresponding to different runs.

Sounds good enough 👍

wincent:
Ok, I think we can merge this as it definitely works.

-   Results/analysis [posted here](https://docs.google.com/spreadsheets/d/1gV6ZDO2lkkQSd4yAf5fXM5xDwSNc3SOtO0ywORfwbh4/edit).
-   Raw results:
    -   [First macOS build, minification off](https://gist.github.com/wincent/7f250dab1defbc5020a64b62d78e95fd)
    -   [Second macOS build, minification on](https://gist.github.com/wincent/24d757c9b051f2793835f7499d28a4ae) (11.7% speed hit)
    -   [First Linux build, minification off](https://gist.github.com/wincent/4eb92ca9c31896c0703a44f329ef0441)
    -   [Second Linux build, minification on](https://gist.github.com/wincent/c1d1b135a15de420e600a4d57a889566) (11% speed hit)
    -   [Third Linux build, minification on](https://gist.github.com/wincent/6b9872e6b82a7fe462ced1c55cc052ab) (note: that _almost everything_ came from cache; minification has _zero_ effect here).

jbalsas:
> Ok, I think we can merge this as it definitely works.
>
> -   Results/analysis [posted here]

noice

@bryceosterhaus, could you see to instrument the new sass task and crunch the numbers to get an idea of the difference it makes?

wincent:
> @bryceosterhaus, could you see to instrument the new sass task and crunch the numbers to get an idea of the difference it makes?

Commented on the relevant bits [here](https://github.com/liferay/liferay-frontend-projects/pull/282/files#r543565295).

PR #300, opened 2020-12-15
fix(js-toolkit): unable to deploy to 7.1 due to InvalidSyntaxException

izaera:
See [LPP-39704](https://issues.liferay.com/browse/LPP-39704)


izaera:
I'm merging this as it is trivial and a customer is depending on it.

PR #301, opened 2020-12-17
feat(js-themes-toolkit): expose r2 API

wincent:
As discussed in [the issue](https://github.com/liferay/liferay-frontend-projects/issues/249), we already have a copy of the liferay-r2 code originally hosted at [liferay-labs/R2](https://github.com/liferay-labs/R2), which got inlined into the themes toolkit in [this commit](https://github.com/liferay/liferay-js-themes-toolkit/commit/a843516ffb18bbeb9b3c3b1c7de8a2622d233b12#diff-a016e05e9164926c610083bf8d96c2a4d473221ecda45c00522bdef0c5d97585).

So, given that we needed access to r2's functionality in [this PR](https://github.com/liferay/liferay-frontend-projects/pull/282), we can just leverage the existing copy of the code that we have in the
theme tasks (npm-scripts already depends on the theme tasks).

Note that I added a couple of simple tests to show that you can actually grab and use the r2 API from the public interface of the package.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/249

### DIFF
--- a/projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/__tests__/index.test.js
+++ b/projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/__tests__/index.test.js
@@ -1,0 +1,29 @@
+/**
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+const {r2} = require('liferay-theme-tasks');
+
+describe('index.js', () => {
+	it('exposes the r2 API from the "main" package resource', () => {
+		expect(require('liferay-theme-tasks')).toBe(require('..'));
+
+		expect(Object.keys(r2).sort()).toEqual(['exec', 'swap', 'valueMap']);
+	});
+
+	describe('r2.swap()', () => {
+		const {swap} = r2;
+
+		it('swap text alignment', () => {
+			expect(swap('p{text-align:right;}')).toEqual(
+				'p{text-align:left;}',
+				'text-align: left => text-align: right'
+			);
+			expect(swap('p{text-align:left;}')).toEqual(
+				'p{text-align:right;}',
+				'text-align: right => text-align: left'
+			);
+		});
+	});
+});
--- a/projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/index.js
+++ b/projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/index.js
@@ -9,6 +9,7 @@ const _ = require('lodash');
 
 const checkNodeVersion = require('../lib/checkNodeVersion');
 const project = require('../lib/project');
+const r2 = require('../lib/r2/liferay-r2');
 const {getArgv} = require('../lib/util');
 const plugin = require('../plugin');
 const registerTaskBuild = require('./tasks/build');
@@ -80,5 +81,6 @@ function register() {
 }
 
 module.exports = {
+	r2,
 	registerTasks,
 };


projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/__tests__/index.test.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+const {r2} = require('liferay-theme-tasks');
+
+describe('index.js', () => {
+	it('exposes the r2 API from the "main" package resource', () => {
+		expect(require('liferay-theme-tasks')).toBe(require('..'));
```
wincent, line 10: Just proving that `'../index.js'` really _is_ the `"main"` export from the package.

projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/__tests__/index.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+		expect(Object.keys(r2).sort()).toEqual(['exec', 'swap', 'valueMap']);
```
wincent, line 12: And then showing what's in it has the expected shape.

projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/__tests__/index.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+		it('swap text alignment', () => {
+			expect(swap('p{text-align:right;}')).toEqual(
+				'p{text-align:left;}',
+				'text-align: left => text-align: right'
+			);
+			expect(swap('p{text-align:left;}')).toEqual(
+				'p{text-align:right;}',
+				'text-align: right => text-align: left'
+			);
```
wincent, line 26: Finally, a quick test to show you can use it to actually do something — this is a small subset of [the tests that over in the actual module itself](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/r2/liferay-r2/__tests__/index.test.js).

wincent:
As this is trivial, I'll merge this (if there are no objects in the next X minutes), so that I can cut a release and have it ready for @bryceosterhaus when his timezone comes online later today.

projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/__tests__/index.test.js
jbalsas: This feels like an original take to me. I get that it shows it has the expected shape, but we hardly ever do this, do we? What's the reason we want do really do it here?

We also know this will fail if we ever add other things to `r2` which I guess is expected?

Nothing blocking here, just looking to understand when would we want to assert exported APIs like this and when not.

wincent: In this case acceptance criteria were "I can use r2 API from theme tasks".

Now, before this PR, you could already do that hackily with:

```js
const r2 = require('liferay-theme-tasks/lib/r2/liferay-r2');
```

but that's fragile (path dependent) and breaks encapsulation. So given that I didn't actually add any new functionality in this PR, I just changed how it was exposed, I wanted to make assertions about _how_ it was exposed. So that leads me to these assertions:

- I can get at the API without reaching "inside" or making assumptions about package structure.
- The API I get is actually the one I think I'm getting (but without re-exercising all of its behavior, because that is already covered elsewhere; that's why we have the quick `swap()` test below, as a smoke test).

So that's why we "hardly ever do this", as you say — most likely, we "never do this — and probably won't ever have to do it again.

wincent: > We also know this will fail if we ever add other things to `r2` which I guess is expected?

And yes, that's right. I promise if this test ever fails for that reason, I will purchase you a beverage of your choice up to a monetary value of 5 _Lifebucks_.

PR #302, opened 2020-12-17
chore(js-themes-toolkit): restore missing license information

wincent:
As noted here:

- https://github.com/liferay/liferay-frontend-projects/issues/249#issuecomment-733010877
- https://github.com/liferay/liferay-js-themes-toolkit/pull/424#discussion_r368860354

We dropped some licensing information when we forked this package from [ded/R2](https://github.com/ded/R2) into [liferay-labs/R2](https://github.com/liferay-labs/R2). We later [inlined the code into the themes toolkit](https://github.com/liferay/liferay-js-themes-toolkit/commit/a843516ffb18bbeb9b3c3b1c7de8a2622d233b12), and did not restore the original copyright.

Seeing as we were looking at this code for [issue #249](https://github.com/liferay/liferay-frontend-projects/issues/249), we may as well put things right now.

Basically, there are only two files of interest in the original implementation (the index and the tests), so we put the correct copyright information in those via an override and leave everything else; you can see that the other files were added by us with original authorship by looking at [the commit listing in the liferay-labs repo](https://github.com/liferay-labs/R2/commits/master).

We make the change on v10 and v9 of the toolkit (v8 doesn't have an inlined copy of the module).

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -19,8 +19,8 @@
 #
 
 /copyright.js
-/maintenance/projects/*/copyright.js
-/projects/*/copyright.js
+/maintenance/projects/**/copyright.js
+/projects/**/copyright.js
 
 #
 # Third-party code
--- a/.prettierignore
+++ b/.prettierignore
@@ -22,5 +22,5 @@
 #
 
 /copyright.js
-/maintenance/projects/*/copyright.js
-/projects/*/copyright.js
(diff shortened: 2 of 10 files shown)


wincent:
Sending this in as it's the companion to https://github.com/liferay/liferay-frontend-projects/pull/301

PR #307, opened 2020-12-17
feat: support for ephemereal bridge generation

izaera:
Due to the errors at [this PR](https://github.com/liferay-frontend/liferay-portal/pull/606#issuecomment-745928072) we must generate all bridge scaffolding at build time so that we don't break the source tree.


izaera:
# Warning: don't review this yet, it needs a heavy refactor

jbalsas:
Hey @izaera, I obviously lack the context to review this in depth, so I'd suggest you abuse @wincent's absence and get this in as soon as you get that CI green 😂

izaera:
This is ready to go (I've tested it locally with [this PR](https://github.com/liferay-frontend/liferay-portal/pull/606) and a patched (according to [this fix](https://github.com/webpack/webpack/pull/12256)) webpack and everythings seems to work.

I will release a new version so that we can test everything in CI.

projects/npm-tools/packages/npm-scripts/src/utils/writeWebpackFederationEntryPoint.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const {transformJsSource} = require('@liferay/js-toolkit-core');
```
bryceosterhaus, line 6: I think we need to include this package in npm-scripts package.json

izaera: Yes, I saw it was missing yesterday, but forgot to send the PR. However it always works because in this repo, the package is linked inside the root `node_modules` and, in portal, because we use the bundler, it also appears :sweat_smile: .

But I'll send a PR to fix it :+1: .

PR #309, opened 2020-12-18
chore(monorepo): use "dependencies" label instead of "dev-dependency"

wincent:
We were using "dev-dependency" over in [the old repo](https://github.com/liferay/liferay-frontend-guidelines) but every time dependabot opens a PR with us it's going to spammily recreate a "dependencies" label, so we may as well just use that one.


wincent:
Trivial, so going to merge this.

PR #310, opened 2020-12-18
chore: set up autolabeling of "docs" and "guidelines" changes

wincent:
Any change to an "*.md" file is labeled as "docs". Additionally, changes under "guidelines/" are special-cased to reflect that they are meta-documentation about how we do development.


wincent:
This is one of those ones which we can't really end-to-end test without merging it, so I'll do that once it's green and we can follow up with a fix if it doesn't work as expected.

PR #361, opened 2020-12-18
refactor(monorepo): split GitHub workflows by Yarn workspaces

wincent:
Before:

-   `ci.yml`:
    -   Runs repo-wide lint and formatting checks.
    -   Runs tests for all four groups of yarn workspaces, corresponding to:
        -   `yarn.lock` (ie. everything under `projects/` on Linux and Windows).
        -   `maintenance/projects/js-themes-toolkit-v8-x/yarn.lock`
        -   `maintenance/projects/js-themes-toolkit-v9-x/yarn.lock`
        -   `maintenance/projects/js-toolkit/yarn.lock`
    -   Checks top-level `yarn.lock` only for unwanted changes during installation.
-   `guidelines.yml`: runs checks specific to "guidelines/" directory (specifically, broken-link checks).
-   `labeler.yml`: applies labels automatically based on paths touched by a PR.

After:

-   `guidelines.yml` and `labeler.yml`: same as before.
-   Split `ci.yml` into four files, one corresponding to each group of yarn workspaces:
    -   `js-themes-toolkit-v8-x.yml`: runs tests for the corresponding directory under "maintenance/"; also checks the lockfile.
    -   `js-themes-toolkit-v9-x`: runs tests for the corresponding directory under "maintenance/"; also checks the lockfile.
    -   `js-toolkit-v2.yml`: runs tests for the corresponding directory under "maintenance/"; also checks the lockfile.
    -   `global.yml`: effectively the remnant of "ci.yml"; now responsible for "everything else" (ie. global lints, format checks, tests under the "projects/" directory, and also checks the top-level lockfile).

So the net effect here should be that we:

-   Always run the "global" tests, lints, and format checks.
-   Always catch staleness of the top-level lockfile.
-   Run tests for projects under "maintenance/" only when touching files under those project folders; and when we do that, we check their lockfiles too.

I guess we'll see if this works... 🤷‍♂️

### DIFF
--- a/.github/workflows/global.yml
+++ b/.github/workflows/global.yml
@@ -1,6 +1,6 @@
 # Based on: https://github.com/actions/starter-workflows/blob/master/ci/node.js.yml
 
-name: ci
+name: global
 
 on:
     push:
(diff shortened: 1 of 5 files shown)


wincent:
About to go on PTO, so seems like a good time to merge this.

PR #365, opened 2020-12-22
feat(js-themes-toolkit): adds js minification step (Fixes #364)

jbalsas:
Hey @bryceosterhaus, @wincent, my gulp fu is pretty rusty (or my fu in general), but I think this should do it as far as minifying JS in themes goes.

> In production mode, `terser` will be invoked to minify `js` resources in a theme after every other processing has taken place.
>
> Terser options can be overriden by declaring a `terser` object in the `package.json` file of your theme. By default, it is invoked with defaults configuration.
>
> Sourcemaps are enabled by default and handled outside of terser if the minification takes place.

Everytime I get back to this I get an intense urge of rewriting the whole thing, but instead, I end up adding yet another thing. One day, though...

promise

PR #367, opened 2020-12-23
chore(npm-scripts): add missing dependency

izaera:
As per https://github.com/liferay/liferay-frontend-projects/pull/307#pullrequestreview-557358660

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/package.json
+++ b/projects/npm-tools/packages/npm-scripts/package.json
@@ -16,6 +16,7 @@
 		"@babel/preset-typescript": "^7.10.4",
 		"@liferay/eslint-config": "21.2.1",
 		"@liferay/jest-junit-reporter": "1.2.0",
+		"@liferay/js-toolkit-core": "3.0.1-pre.1",
 		"@liferay/npm-bundler-preset-liferay-dev": "4.6.5",
 		"@testing-library/dom": "^5.6.1",
 		"@testing-library/jest-dom": "^4.2.4",


yarn.lock
```diff
@@ -9820,6 +9820,51 @@ liferay-npm-bundler@2.19.4:
     xml-js "^1.6.8"
     yargs "^14.0.0"
 
+liferay-theme-tasks@10.1.0:
```
jbalsas, line 9823: I'm not confident about this... @wincent, did you send something to fix or not-fix but something this?

izaera: Yes. It looked correct in a first small analysis, but I'm not quite sure of why we get this. :thinking:

izaera: I guess this is fixed by #371 :thinking:

wincent: Yep, that's right.

> I'm not quite sure of why we get this

799f34240f5d6139e93582af6a530040ac77cb6a shows what Yarn is complaining about. If we release a new version of a workspace project (ie. in this case, `liferay-theme-tasks@10.2.0`) but continue to have an internal dependency on the old version of the project (ie. `liferay-theme-tasks@10.1.0`), Yarn is going to drag the old version into the lockfile. The fact that Yarn complains (and that CI surfaces the complaint) is a good thing, because it shows us that we forgot to update an internal reference. And the stale reference is a problem not just because of lockfile noise but because it means we have two copies of our own dependency floating around in the worktree; in this case:

- `10.2.0` at `projects/js-themes-toolkit/packages/liferay-theme-tasks`; and:
- `10.1.0` at `projects/npm-tools/packages/npm-scripts/node_modules/liferay-theme-tasks`

We don't want that kind of thing in general because it makes us look fat, and in this specific instance because it means the improvement added in 10.2.0 (#365) might get overlooked and not actually make it into liferay-portal in a timely fashion.

> I guess this is fixed by

Yep, 5145c8d087f8a1ae129f090cb06d3b870d08d6f1 just updates the internal reference and the redundancy goes away.

wincent:
Rebased and force-pushed this to drop the now-unnecessary commit [discussed here](https://github.com/liferay/liferay-frontend-projects/pull/367#discussion_r549418126).

wincent:
CI is happy with this, so _moiging_.

merge

yarn.lock
izaera: Strange. I think I also ran `yarn why` (it's the first I instinctively do when this things happen) but didn't report any duplicates. However it was some time ago, so maybe I didn't do it... :shrug:

PR #368, opened 2020-12-23
feat(js-themes-toolkit): add build option for using dart-sass instead of lib-sass

jbalsas:
Quick and dirty POC so @bryceosterhaus can take a look at this.

Getting relative import errors:

> Can't find stylesheet to import.
>     @import 'clay/base';


projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/build/sass.js
```diff
@@ -0,0 +1,58 @@
...
+	if (!file.contents.length) {
+	  file.path = replaceExtension(file.path, '.css');
+	  return cb(null, file);
+	}
+
+	try {
+		const {css} = sass.renderSync({
+			data: file.contents.toString(),
+			file: file.path,
+			includePaths: [path.dirname(file.path), file.path]
```
bryceosterhaus, line 33: I think this should actually be taken from the options config.

`includePaths: options.includePaths,`

bryceosterhaus (commented):
I can't seem to figure out why this is behaving like it is. Part of me thinks its the test that is flakey, the setup for the test is really intense and trying to figure out all the steps is a little confusing.

Manually inspecting the test folder structure, I don't see any place where clay exists, so it'd make sense to me that it can't find that import.

I can dig into this a bit more next week if we still can't get it to work properly

projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/build/sass.js
jbalsas: Yeah, at this point I was just trying to reduce the problem as much as possible, so just hardcoded this to make sure I had the proper paths.

This for sure shouldn't be merged as is ;)

jbalsas:
>I can't seem to figure out why this is behaving like it is. Part of me thinks its the test that is flakey, the setup for the test is really intense and trying to figure out all the steps is a little confusing.
>
>Manually inspecting the test folder structure, I don't see any place where clay exists, so it'd make sense to me that it can't find that import.
>
>I can dig into this a bit more next week if we still can't get it to work properly

Yeah, it occurred to me that the test was flaky, but it was working before. The setup part of the theme will invoke the generator and overlay the `unstyled` and `styled` themes on top of the test one in a temp directory. Querying that directory for the `clay` folder before invoking `sass` shows that the files are indeed there.

I'll do some tests directly on real themes to see if they build and we can pin this on the test itself and report back later.

jbalsas:
Okay, so I can 'kinda'  confirm that this actually builds a theme in DXP, so the suspicion goes back to the test itself... @bryceosterhaus, could you take a look, see if you can figure out something else?

/cc @pat270, testing this in DXP, the new compiler complains about 2 things:

#### Unexpected `!default`

> Message:
>     build/_css/compat/_atlas_variables.scss
> Error: Expected "important".
> 
> 35 │ $panel-title-text-transform:  !default;

This comes from [_atlas_variables](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-theme/frontend-theme-styled/src/main/resources/META-INF/resources/_styled/css/compat/_atlas_variables.scss#L35). I see to remember we dealt with similar issues... what should be the fix there?

#### compound selectors may no longer be extended

> Message:
>     build/_css/compat/components/_dropdowns.scss
> Error: compound selectors may no longer be extended.
> Consider `@extend .dropdown-item, .disabled` instead.
> See http://bit.ly/ExtendCompound for details.
> 
> 34 │         @extend .dropdown-item.disabled;

This comes from [compat/components/_dropdowns](https://github.com/liferay/liferay-portal/blob/master/portal-web/test/functional/com/liferay/portalweb/dependencies/test-theme.war/css/compat/components/_dropdowns.scss#L24). The fix seems pretty self-explanatory and http://bit.ly/ExtendCompound explains the reasoning behind it, so I guess this should be safe to modify too.

jbalsas:
The not-so-good news is that in this case, the build seems to be ~50% slower than with `node-sass` with 15s compared to 10s previously.

bryceosterhaus:
> Okay, so I can 'kinda' confirm that this actually builds a theme in DXP, so the suspicion goes back to the test itself... @bryceosterhaus, could you take a look, see if you can figure out something else?

I spent some time trying to debug this and I can't seem to wrap my head around why its failing... I inspected the generated file structure and everything seems to work fine. I looked through the internals of gulp-sass and it seems pretty straight forward, not like they are doing anything special. I'll continue to take a look, but otherwise I am pretty stumped at knowing why sass compiling is breaking in the test.

jbalsas:
> I spent some time trying to debug this and I can't seem to wrap my head around why its failing... I inspected the generated file structure and everything seems to work fine. I looked through the internals of gulp-sass and it seems pretty straight forward, not like they are doing anything special. I'll continue to take a look, but otherwise I am pretty stumped at knowing why sass compiling is breaking in the test.

Yeah, same here from when I last checked. I'd say we could try to use the opportunity to figure out a more rational test :)

@pat270, could you please take a look at https://github.com/liferay/liferay-frontend-projects/pull/368#issuecomment-751801122 so this is unblocked once the tooling part is done?

pat270:
Sorry I missed this, hard to keep track with all the emails in my inbox.

@jbalsas These should be non issues in Clay CSS.

> $panel-title-text-transform: !default;

I was taking advantage of bug in libsass where I can reset a `!default`  variable back to `null`. It stopped working when we updated libsass/node-sass in Clay. There should be no instances of this in 3.x.

> Error: compound selectors may no longer be extended

I only extend Sass placeholders in Clay CSS. We don't extend CSS selectors. I can't say for Portal though.

bryceosterhaus:
@jbalsas I was able to get the test working. It was breaking due to Jest's default env of jsdom. Still a POC and need to clean up the error handling

Now just facing errors like this.
Screen Shot 2021-01-19 at 4 13 47 PM

Next steps would probably be to address these issues in our code before merging and releasing this code?

jbalsas:
> Now just facing errors like this.

Where are those occurring? We should've removed all `compat` code by now, so that shouldn't be an issue in current `master` in DXP. I see still one case but it should be in a test fixture so it shouldn't fail the build.

> Next steps would probably be to address these issues in our code before merging and releasing this code?

Now that I come to think of it, I guess we're pulling an old version of `styled` for the tests, so we'll need to push a new one out and start using it to get the tests green again.

bryceosterhaus:
> Now that I come to think of it, I guess we're pulling an old version of styled for the tests, so we'll need to push a new one out and start using it to get the tests green again.
> 
> Maybe a better option would be to take a step further and try to simplify the themes so we don't rely on real artefacts such as theme_styled and rather have our own test themes that we can better control.


Yeah looks like its due to an old version of styled. I didn't realize we were publishing these actually, any idea where and who is responsible for this? The published versions on npm are all over the place
Screen Shot 2021-01-20 at 2 01 35 PM

bryceosterhaus:
> Maybe a better option would be to take a step further and try to simplify the themes so we don't rely on real artefacts such as theme_styled and rather have our own test themes that we can better control.

One thought on this is that the real artifacts are part of the functionality and thus testing it with the actual publish artifacts is probably important. One option would be to move the themes here, but that seems like overkill IMO.

jbalsas:
> One thought on this is that the real artifacts are part of the functionality and thus testing it with the actual publish artifacts is probably important. 

I was thinking about this in terms that the functionality is really the overlay of base themes and the successful compilation. People can use other bases other than ours.

That being said, catching that this fails with currently published versions of `styled` is a good thing since we know we can't just merge this or we'll risk breaking theme development in many places, so we might as well keep it that way.

> One option would be to move the themes here, but that seems like overkill IMO.

I'm pretty sure that would create tons of different issues ;)

> any idea where and who is responsible for this? 

Themes are published automatically whenever maintenance branches `.x` are updated/published. That is new versions are expected for each FixPac. Versions don't get published from `master` until a real release is out so we won't see one coming from there anytime soon unless we push for it (which we can).

Based on this, I think this kinda signals that we might need to:
- Keep `node-sass` around and provide a dual-build so old themes can continue building with it without disruption
- Kick off a new version of the theme tooling and publish a new version of the themes from `master` to go with it as we did last time

- A dual build allows us to stay on this version so we don't have to worry about applying different fixes across branches but doesn't bring us an install speed benefit since we won't be free of `node-sass`.
- A new version gives us some freedom to start exploring other approaches (drop gulp?) but it adds yet another different codebase to maintain

Thoughs?

/cc @izaera, @wincent, @dsanz, @david-truong

wincent:
> * A dual build allows us to stay on this version so we don't have to worry about applying different fixes across branches but doesn't bring us an install speed benefit since we won't be free of `node-sass`.
> * A new version gives us some freedom to start exploring other approaches (drop gulp?) but it adds yet another different codebase to maintain

I don't even think install speed is the main reason to want to get rid of `node-sass`; for me, it's that installation _may not work at all_ in the presence of `node-sass` due to its incredible sensitivity to the version of Node installed on the system, which has led to much confusion and many support questions in different forums over the years.

Also:

you-had-me-at-drop-gulp

jbalsas:
> you-had-me-at-drop-gulp

helpmehelpyou

bryceosterhaus:
Alright I went with trying to get it to work with adding a env variable for choosing whether its dart-sass or node-sass. In order to properly run the test though I think the base theme needs to be in node_modules. I seemed to have tried every which way to use a local base theme but just couldn't get it working. Debugging the gulp tasks is just a nightmare since if anything fails it emits `Error [ERR_UNHANDLED_ERROR]: Unhandled error. ([object Object])`.

So if we could get a release of latest styled and unstyled theme on a beta tag and then we can test the dart-sass flag against that.

And dropping gulp seems like a great idea, would love to see it re-worked without gulp.

bryceosterhaus:
I'm having quite the headache here... I released new versions off of master for both unstyled and style `v6.0.1-test.fa6ecdb8ad0e`. However, now the original test breaks with this new version of themes and using dart-sass also breaks too.

Worst of all, trying to debug the test failures has gotten me no where and I can't quite pinpoint whats even failing. The way the test is arranged, if any part of the gulp process of test setup fails it just results in `Error [ERR_UNHANDLED_ERROR]: Unhandled error. ([object Object])`.

I was thinking of just re-writing the whole test, but then I thought it might just be better to start to gulp removal and write new tests in the process of doing that. Any thoughts on direction from here? Or if there are any gulp wizards out there, I am happy to get their help. Latest code of mine is on this PR, although its nothing very helpful.

wincent:
> I was thinking of just re-writing the whole test, but then I thought it might just be better to start to gulp removal and write new tests in the process of doing that. Any thoughts on direction from here? Or if there are any gulp wizards out there, I am happy to get their help. Latest code of mine is on this PR, although its nothing very helpful.

I'll have a look at the failure this morning on the off chance that I can surface any more info that might help.

Rewriting tests is always tricky (in terms of understanding the behavior of the old tests and making sure the new ones have similar/better/equivalent coverage), but if you are hard-blocked, it may be the only way forward... I haven't looked at the tests since forever, but I am imagining that they might not be the prettiest things ever seen... 😬 

About removing gulp, that would be awesome but it would be a pretty big task; given that part of our public API is [the ability to extend the gulp tasks via the hook mechanism described here](https://help.liferay.com/hc/en-us/articles/360018166731-Overwriting-and-Extending-Liferay-Theme-Tasks), we need to figure out whether we make a breaking change (which as you know, is pretty frowned upon in the Liferay Code of Chivalry and Etiquette — there would need to be a commensurately huge upside in order to justify the change) or introduce some kind of compatibility layer which would enable us to maintain something very like the old extension mechanism without actually depending on Gulp (shuddering at the thought of the complexity).

wincent:
Ok, I think I've got it figured out, but it is going to take me a while to sort it all out. Will report back "shortly".

wincent:
Ok, so I pushed [a commit](https://github.com/liferay/liferay-frontend-projects/pull/368/commits/c2fe98983939de5bc531a52fd14eaf86e8fa9311) to make debugging these changes easier. That surfaces the error messages that would otherwise get obscured/obfuscated and reveals a number of files that are broken in the newly published themes.

Let's look at what is different between 5.0.0 and 6.0.1. Testing this out in a clean project, the theme packages have bloated considerably:

- `yarn add liferay-theme-styled@5.0.0`: 788KB on disk (16 directories, 173 files).
- `yarn add liferay-theme-unstyled@5.0.0`: 20MB on disk (84 directories, 2658 files).
- `yarn add liferay-theme-styled@6.0.1-test.fa6ecdb8ad0e`: 1.8MB on disk (57 directories, 402 files).
- `yarn add liferay-theme-unstyled@6.0.1-test.fa6ecdb8ad0e`: 30MB on disk (203 directories, 3851 files).

Other than the bloat, has the structure changed at all? Yes, it seems it has. Here is the top level of "unstyled" at 5.0.0:

```
css
images
js
package.json
templates
```

vs at 6.0.1:

```
bnd.bnd
build
build.gradle
classes
package.json
src
```

So that in itself looks broken, and "styled" is the same, but in the interest of continuing troubleshooting, with the new logging in place from the commit we can temporarily unbreak things by moving the files:

```
cd node_modules
mv liferay-frontend-theme-styled/src/main/resources/META-INF/resources/_styled/* liferay-frontend-theme-styled/
mv liferay-frontend-theme-unstyled/src/main/resources/META-INF/resources/_unstyled/* liferay-frontend-theme-unstyled
```

So now our test assertions are no longer swallowed but gulp-sass(?) is still throwing an error. We need to hackily edit `node_modules/undertaker/lib/helpers/createExtensions.js` to see what it actually is:

```
build/_css/clay.scss
Error: File to import not found or unreadable: clay/base.
        on line 1 of build/_css/clay.scss
>> @import 'clay/base';
```

Something fishy may be going on there; right now I am not sure whether dart-sass or gulp-sass is running because the errors for _both_ tests show this in their metadata:

```
      plugin: 'gulp-sass',
```

And the error text doesn't match what Chema posted in his original description. So I am not really sure which plug-in is running (I'm wondering whether something about plug-in registration is at issue here, where we _can't_ actually override during the test suite because some kind of global state has been closed over by the time we do it), and I'm also aware that by `mv`-ing the files around like I did I might have messed with the search paths in some way that prevent it from working. I'll dig a little bit more and report back if I find anything out.

wincent:
Actually, test passes if I actually move the built files instead of the source ones:

```
cd node_modules
mv liferay-frontend-theme-unstyled/classes/META-INF/resources/_unstyled/* liferay-frontend-theme-unstyled/
mv liferay-frontend-theme-styled/classes/META-INF/resources/_styled/* liferay-frontend-theme-styled/
```

"Proof":

```
yarn test packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js
yarn run v1.22.5
$ liferay-workspace-scripts test packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js
Running: yarn --cwd /Users/greghurrell/code/liferay-frontend-projects run jest packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js
$ /Users/greghurrell/code/liferay-frontend-projects/node_modules/.bin/jest packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js
[17:55:12] liferay-frontend-theme-unstyled using custom path: /Users/greghurrell/code/liferay-frontend-projects/node_modules/liferay-frontend-theme-unstyled
[17:55:12] liferay-frontend-theme-unstyled using custom path: /Users/greghurrell/code/liferay-frontend-projects/node_modules/liferay-frontend-theme-unstyled
[17:55:12] liferay-frontend-theme-styled using custom path: /Users/greghurrell/code/liferay-frontend-projects/node_modules/liferay-frontend-theme-styled
[17:55:14] gulp-inject 1 file into _custom.scss.
[17:55:14] gulp-inject 1 file into portal_normal.ftl.
[17:55:20] liferay-frontend-theme-unstyled using custom path: /Users/greghurrell/code/liferay-frontend-projects/node_modules/liferay-frontend-theme-unstyled
[17:55:20] liferay-frontend-theme-unstyled using custom path: /Users/greghurrell/code/liferay-frontend-projects/node_modules/liferay-frontend-theme-unstyled
[17:55:20] liferay-frontend-theme-styled using custom path: /Users/greghurrell/code/liferay-frontend-projects/node_modules/liferay-frontend-theme-styled
[17:55:22] gulp-inject 1 file into _custom.scss.
[17:55:22] gulp-inject 1 file into portal_normal.ftl.
 PASS  projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js (17.876 s)
  using lib_sass
    ✓ build task should correctly compile theme (8457 ms)
    ✓ build task should correctly compile theme with dart-sass (7344 ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        17.957 s
Ran all test suites matching /packages\/liferay-theme-tasks\/theme\/tasks\/__tests__\/build.test.js/i.
✨  Done in 19.35s.
```

bryceosterhaus:
Thanks for helping me on this, this should help me quite a bit.

> given that part of our public API is the ability to extend the gulp tasks via the hook mechanism described here

I was thinking about that as well and my guess is that there really isn't enough upside for the consumer to move away from using gulp. Major upside would be Chema's happiness and that may just be more important than any customers we have 😂

bryceosterhaus:
> Here is the top level of "unstyled" at 5.0.0:
> 
> ```
> css
> images
> js
> package.json
> templates
> ```
> 
> vs at 6.0.1:
> 
> ```
> bnd.bnd
> build
> build.gradle
> classes
> package.json
> src
> ```

So with this change, are our current theme tasks even supposed to support this structure? Forgive my ignorance, but I also am fairly clueless when it comes to themes in DXP. Last time I touched anything relating to themes was back in 6.2.x

cc @jbalsas maybe you have a better idea of whats supported

jbalsas:
> So with this change, are our current theme tasks even supposed to support this structure? Forgive my ignorance, but I also am fairly clueless when it comes to themes in DXP. Last time I touched anything relating to themes was back in 6.2.x

I think the issue here is with the publication process. I don't fully remember, but my guess is that we publish the build folder so there are some steps involved:
- Run gulp build
- Move to the build folder
- Copy the package.json to `build` (if it isn't automatically done)
- Run `npm publish` from there

Some part of that logic lives inside [PublishNodeModuleTask](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/sdk/gradle-plugins-node/src/main/java/com/liferay/gradle/plugins/node/tasks/PublishNodeModuleTask.java) which is active for the modules [frontend-css-common](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/frontend-css/frontend-css-common/build.gradle#L10), [frontend-theme-styled](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/frontend-theme/frontend-theme-styled/build.gradle#L3-L8), [frontend-theme-unstyled](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/frontend-theme/frontend-theme-unstyled/build.gradle#L16)

So, I think we probably want to figure out the publication process rather than updating the way we handle it.

bryceosterhaus:
> I think the issue here is with the publication process

ah good looking out on that, I hadn't thought there would be an intermediary steps for publishing them.

Looks like they do...
- move the root package.json into `src/main/resources/META-INF/resources/_styled`
- copy some meta-data to the package.json
- version & publish
- remove package.json from folder


Okay so I updated the code and was able to get both tests to pass and I think it works as intended now. One part I am unsure about is how I am using the two versions of themes. I am using [package aliases](https://github.com/npm/rfcs/blob/latest/implemented/0001-package-aliases.md#package-aliases) which is only available as of npm v6.9.0. Any better ways we have handled this?

wincent:
> Okay so I updated the code and was able to get both tests to pass and I think it works as intended now. One part I am unsure about is how I am using the two versions of themes. I am using [package aliases](https://github.com/npm/rfcs/blob/latest/implemented/0001-package-aliases.md#package-aliases) which is only available as of npm v6.9.0. Any better ways we have handled this?

I don't quite get why you need two versions of the themes. I [saw all tests passing](https://github.com/liferay/liferay-frontend-projects/pull/368#issuecomment-767680335) with a single version of the themes (6.0.1-test.fa6ecdb8ad0e), once I fixed the damage inside `node_modules`. What am I missing?

bryceosterhaus:
@wincent my thought is that we should continue to test node-sass against the old version of themes to make sure we don't break anything for those using node-sass. Maybe we don't need to do that though?

wincent:
> @wincent my thought is that we should continue to test node-sass against the old version of themes to make sure we don't break anything for those using node-sass. Maybe we don't need to do that though?

I don't think we do, unless I am missing something. It seems that seeing them pass against the new versions is enough, isn't it? Unless things were totally deceiving me the other day, I saw the tests pass against the new themes with both node-sass and sass:

```
    ✓ build task should correctly compile theme (8457 ms)
    ✓ build task should correctly compile theme with dart-sass (7344 ms)
```

And we've never historically tested more than a single version at a time.

bryceosterhaus:
> It seems that seeing them pass against the new versions is enough, isn't it?

Yeah thats probably true. I just updated the code and cleaned up the commit history. Let me know if anything stands out.

One thing I noticed is some test errors with `Cannot find module '@liferay/js-toolkit-core' or its corresponding type declarations.` but these aren't related to theme tasks at all.

wincent:
> One thing I noticed is some test errors with `Cannot find module '@liferay/js-toolkit-core' or its corresponding type declarations.` but these aren't related to theme tasks at all.

I blame @izaera for those (just kidding, but he might have some insight about them). I gather you're looking into the other ones though, right? Those do look to be related (specifically, [some ones in here](https://github.com/liferay/liferay-frontend-projects/pull/368/checks?check_run_id=1793561857)).

bryceosterhaus:
@wincent I actually didn't realize the sass tests were failing till you mentioned it. I had some stale state that made my local tests pass. Looks like the new theme I added caused more issues than it helped, so I removed that and just use the 7.2 theme for both tests. Looks to work locally, we will see for ci now...

bryceosterhaus:
Not sure about these failing windows tests, gonna investigate although I don't have a windows machine to fully debug

bryceosterhaus:
Looks like tests are all good now. We good with moving forward with this or any last thoughts, fears, or hesitations?

wincent:
> Looks like tests are all good now. We good with moving forward with this or any last thoughts, fears, or hesitations?

Yeah. I see you're still using package aliases to install two versions of each theme at once. I thought we concluded that that wouldn't be necessary?

bryceosterhaus:
@wincent not sure how that stuck around, I remember specifically removing that. 🤷‍♂️  oh well, I removed it again and just updated the PR.

bryceosterhaus:
These tests on windows are rather flakey. I don't have access to windows machine, but I can't seem to reproduce any failures on mac.

izaera:
>> One thing I noticed is some test errors with Cannot find module '@liferay/js-toolkit-core' or its corresponding type declarations. but these aren't related to theme tasks at all.

> I blame @izaera for those (just kidding, but he might have some insight about them). I gather you're looking into the other ones though, right? Those do look to be related (specifically, some ones in here).

Test errors may appear whenever you use the JS Toolkit without having built it. I don't know if it is the case... Is there a link to the errors?

The "problem" is that our `test` target doesn't invoke `build` because it would make it slower for very little benefit.

Also, I will migrate the JS Toolkit modules we are using from `npm-scripts` to `npm-scripts` itself. That should alleviate some of the problems, but not the ones with Themes Toolkit (that I believe uses the JS Toolkit too). But we can give it a try if you want, since JS Toolkit is going to disappear/freeze some day...

projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/test/fixtures/themes/7.2/base-theme-7-2/package.json
```diff
@@ -36,7 +36,7 @@
 		"liferay-theme-tasks": "^9.5.0",
 		"compass-mixins": "0.12.10",
 		"liferay-frontend-common-css": "^1.0.4",
-		"liferay-frontend-theme-styled": "^4.0.0-alpha.1552930087997",
-		"liferay-frontend-theme-unstyled": "^4.0.0-alpha.1552930030671"
+		"liferay-frontend-theme-styled": "^6.0.1-test.1c962d5e12be",
+		"liferay-frontend-theme-unstyled": "^6.0.1-test.1c962d5e12be"
```
wincent, line 40: I don't think we want to change this one, right? (I was sure we commented on this in some previous iteration somewhere, but I can't find the comment right now in GitHub or in Gmail...)

When you make a theme for 7.2, you're supposed to use v4 of the themes. Note that this change doesn't affect the lockfile, and it shouldn't affect the tests either.

package.json
```diff
@@ -29,8 +29,8 @@
 		"gulp-coveralls": "0.1.4",
 		"http-server": "^0.11.1",
 		"jest": "26.4.2",
-		"liferay-frontend-theme-styled": "5.0.0",
-		"liferay-frontend-theme-unstyled": "5.0.0",
+		"liferay-frontend-theme-styled": "6.0.1-test.1c962d5e12be",
+		"liferay-frontend-theme-unstyled": "6.0.1-test.1c962d5e12be",
```
wincent, line 33: Before we go final with this, do we want to publish non-`test` releases? Or do we have to wait for 7.4 release to do that? (ie. because we want to make sure that `@latest` continues to point at the v5 versions?)

wincent (commented):
> These tests on windows are rather flakey. I don't have access to windows machine, but I can't seem to reproduce any failures on mac.

Sadly, I use a VM for that when I need to test or troubleshoot on Windows. Sometimes I just guess what the problem might be and try fixing it and re-push. In this case, the 30000ms timeout might be a "flakey" in the sense that it might pass sometimes but not others.

As least some of the others, though, may be indicative of real issues. For example `Error: ENOENT: no such file or directory, open 'C:\Users\RUNNER~1\AppData\Local\Temp\liferay-theme-tasksWUKJWa\dart_sass_build_task\7.2\base-theme-7-2\build\css\main.css'` might be a real bug (because some path handling happening elsewhere isn't being done in a Windows-friendly way).

```
FAIL projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js (39.148 s)
  ● using lib_sass › build task should correctly compile theme

    : Timeout - Async callback was not invoked within the 30000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 30000 ms timeout specified by jest.setTimeout.

       98 | describe('using lib_sass', () => {
       99 | 	let buildPath;
    > 100 | 	let sassOptionsCalled;
          | 	 ^
      101 | 	let tempTheme;
      102 | 
      103 | 	afterEach(() => {

      at new Spec (../../node_modules/jest-cli/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:100:3)
      at Object.<anonymous> (packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:93:1)

  ● using lib_sass › build task should correctly compile theme with dart-sass

    Unhandled error. ({
      uid: 50,
      name: '_assertFixAtDirectives',
      branch: false,
      error: Error: ENOENT: no such file or directory, open 'C:\Users\RUNNER~1\AppData\Local\Temp\liferay-theme-tasks-WUKJWa\dart_sass_build_task\7.2\base-theme-7-2\build\css\main.css'

      235 | 		assertExists(cssPath);
      236 | 		assertMatches(
    > 237 | 			fs.readFileSync(mainCssPath, 'utf8'),
          | 			   ^
      238 | 			/@import\surl\(file\.css\?t=[0-9]+\);/
      239 | 		);
      240 | 

      at _assertFixAtDirectives (packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:237:7)
      at asyncRunner (../../node_modules/async-done/index.js:55:18)
      at processTicksAndRejections (../../internal/process/task_queues.js:79:11) {
          errno: -4058,
          syscall: 'open',
          code: 'ENOENT',
          path: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\liferay-theme-tasks-WUKJWa\\dart_sass_build_task\\7.2\\base-theme-7-2\\build\\css\\main.css',
          domainThrown: true
        },
        duration: [ 0, 1743900 ],
        time: 1612386697646
      })Error [ERR_UNHANDLED_ERROR]: Unhandled error. ({
        uid: 50,
        name: '_assertFixAtDirectives',
        branch: false,
        error: Error: ENOENT: no such file or directory, open 'C:\Users\RUNNER~1\AppData\Local\Temp\liferay-theme-tasks-WUKJWa\dart_sass_build_task\7.2\base-theme-7-2\build\css\main.css'
      at _assertFixAtDirectives (packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:237:7)
      at asyncRunner (../../node_modules/async-done/index.js:55:18)
      at processTicksAndRejections (../../internal/process/task_queues.js:79:11) {
          errno: -4058,
          syscall: 'open',
          code: 'ENOENT',
          path: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\liferay-theme-tasks-WUKJWa\\dart_sass_build_task\\7.2\\base-theme-7-2\\build\\css\\main.css',
          domainThrown: true
        },
        duration: [ 0, 1743900 ],
        time: 1612386697646
      })
      at Object.error (../../node_modules/undertaker/lib/helpers/createExtensions.js:61:10)
      at handler (../../node_modules/now-and-later/lib/mapSeries.js:47:14)
      at f (../../node_modules/once/once.js:25:25)
      at f (../../node_modules/once/once.js:25:25)
      at tryCatch (../../node_modules/async-done/index.js:24:15)
      at done (../../node_modules/async-done/index.js:40:12)
      at Domain.onError (../../node_modules/async-done/index.js:51:5)

FAIL projects/npm-tools/packages/changelog-generator/test/index.js (9.92 s)
  ● getVersion() › from "2.1.4-pre.0" with version "premajor" and prefix "v" returns "v3.0.0-pre.0"

    Unhandled error. ({
      uid: 54,
      name: '_assertSrc',
      branch: false,
      error: Error: Expected file does not exist at: C:\Users\RUNNER~1\AppData\Local\Temp\liferay-theme-tasks-WUKJWa\dart_sass_build_task\7.2\base-theme-7-2\build\WEB-INF

      80 | 
      81 | 	throw new Error(message);
    > 82 | }
         |  ^
      83 | 
      84 | beforeAll(() => {
      85 | 	process.env.LIFERAY_THEME_STYLED_PATH = path.dirname(

      at fail (projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:82:9)
      at assertExists (projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:50:5)
      at _assertSrc (projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:234:5)
      at asyncRunner (node_modules/async-done/index.js:55:18)
         {
          domainThrown: true
        },
        duration: [ 0, 4792300 ],
        time: 1612386717689
      })Error [ERR_UNHANDLED_ERROR]: Unhandled error. ({
        uid: 54,
        name: '_assertSrc',
        branch: false,
        error: Error: Expected file does not exist at: C:\Users\RUNNER~1\AppData\Local\Temp\liferay-theme-tasks-WUKJWa\dart_sass_build_task\7.2\base-theme-7-2\build\WEB-INF
      at fail (projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:82:9)
      at assertExists (projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:50:5)
      at _assertSrc (projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/__tests__/build.test.js:234:5)
      at asyncRunner (node_modules/async-done/index.js:55:18)
         {
          domainThrown: true
        },
        duration: [ 0, 4792300 ],
        time: 1612386717689
      })
      at Object.error (node_modules/undertaker/lib/helpers/createExtensions.js:61:10)
      at handler (node_modules/now-and-later/lib/mapSeries.js:47:14)
      at f (node_modules/once/once.js:25:25)
      at f (node_modules/once/once.js:25:25)
      at tryCatch (node_modules/async-done/index.js:24:15)
      at done (node_modules/async-done/index.js:40:12)
      at Domain.onError (node_modules/async-done/index.js:51:5)


Test Suites: 2 failed, 128 passed, 130 total
Tests:       3 failed, 5 skipped, 4 todo, 1626 passed, 1638 total
Snapshots:   64 passed, 64 total
Time:        180.206 s
```

package.json
bryceosterhaus: From my understanding, we shouldn't release anything under the `latest` tag manually. I was under the impression that we wait for 7.4 to release those.

It is a little odd to be testing off of a `beta` release, but since we are the only ones using these tests(infra team) I don't think it would be a major issue.

projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/test/fixtures/themes/7.2/base-theme-7-2/package.json
bryceosterhaus: Since it doesn't matter either way for the tests since they reference the root version, is your thought that its just best to keep it the old version for people who might look through these tests?

I'm not opposed to keeping it at the older version, just curious of your thought on it.

wincent: > Since it doesn't matter either way for the tests since they reference the root version, is your thought that its just best to keep it the old version for people who might look through these tests?

Yep. We have too much confusion around version numbers and dependencies as it is. So if the "truth" is that we expect a specific version of the theme to be used with a specific version of DXP, then we should make sure our tests don't contradict that.

bryceosterhaus:
I can't seem to pinpoint these windows failures. I spent some time debugging in a windows vm, but turns out that the tests past locally there as well. @wincent could there be some sort of dirty cache folder? I noticed the CI is using the `--cache-folder` flag and wasn't sure where we are exactly pulling that from.

wincent:
> @wincent could there be some sort of dirty cache folder? I noticed the CI is using the `--cache-folder` flag and wasn't sure where we are exactly pulling that from.

Anything is possible, but I don't think the cache should be an issue. It's caching [the result of the `yarn` install](https://github.com/liferay/liferay-frontend-projects/blob/fe5e842d2109d301d6d8337513ecb57e2b48c8c3/.github/workflows/global.yml#L115-L122) and nothing else, so unless there is a bug in Yarn or some subtle non-determinism, it is unlikely to be the cause. But like I said, anything _is_ possible...

You can try out your hypothesis by bumping the cache-breaker though to start with a clean cache; just change `2` to `3` [on this line](https://github.com/liferay/liferay-frontend-projects/blob/fe5e842d2109d301d6d8337513ecb57e2b48c8c3/.github/workflows/global.yml#L13).

bryceosterhaus:
Alright, turned out to just be a timing issue is seems. Running the tests locally passed fine with the default timeout, but it seems the github windows env is much slower and so increasing the timeout helped.

wincent:
Just pushed [a commit](https://github.com/liferay/liferay-frontend-projects/pull/368/commits/b435c78e980179746fe3dff96eeda76754f45c14) deduplicating the yarn.lock, but if CI likes this, I don't see any reason why not to merge. 👍

bryceosterhaus:
Ci seems to like it, going to merge

wincent:
> Ci seems to like it, going to merge

frodo waking up

PR #374, opened 2020-12-28
refactor(guidelines): use a better name for CI job

wincent:
So it will show up as "guidelines/check-links" instead of "guidelines/build" (which is a bit misleading. All the other jobs have more accurate/descriptive names, like "check-lockfile", "format-check" and so on, so this just brings this workflow into line with the others.

### DIFF
--- a/.github/workflows/guidelines.yml
+++ b/.github/workflows/guidelines.yml
@@ -16,7 +16,7 @@ env:
     yarn-cache-path: .yarn
 
 jobs:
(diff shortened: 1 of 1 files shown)


jbalsas:
Screen Shot 2020-12-28 at 17 04 14

I think I've got this... hope I don't break anything :D

wincent:
> hope I don't break anything

push-and-pray

PR #375, opened 2020-12-29
fix: stabilize liferay-portal build until a new webpack release arrives

izaera:
This is like reverting to version 36.2.0 but without removing the code, so that we leave `liferay-portal` build in a stable situation for further development.

The code being removed was necessary for [LPS-124287](https://issues.liferay.com/browse/LPS-124287) but given that it won't make it into portal until a new webpack release is made, we are removing this code until then.

The serve side counterpart of this PR is [here](https://github.com/liferay-frontend/liferay-portal/pull/660)

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/utils/createBridges.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/createBridges.js
@@ -24,12 +24,15 @@ const writeBridge = require('./writeBridge');
  * @return {void}
  */
 module.exports = function (bridges, dir) {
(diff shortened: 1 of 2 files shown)


izaera:
I'm testing this locally (for the last time) before releasing. I want to make sure it will work.

izaera:
This works in a local build with some minor changes. I'll prepare the PR to link it so that we can review both sides of the "feature".

izaera:
I'm merging this because it's green and I need to release.

PR #376, opened 2020-12-30
feat: upgrade webpack and add more portal projects to federated build

izaera:
See commits for more info.

This is the final step needed for [LPS-24287](https://issues.liferay.com/browse/LPS-124287)

I will merge this and release a new version so that I can make a server side PR to be tested in CI.

The server counterpart of this PR is in a private [branch](https://github.com/izaera/liferay-portal/tree/LPS-124287-followup) until this gets merged and released.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
@@ -21,9 +21,24 @@ const writeWebpackFederationEntryPoint = require('./writeWebpackFederationEntryP
  * the relation is not used for anything. However, it is better to do it this
  * way, because if we mix all together we end up with a very entangled
  * configuration.
+ *
+ * Note also that when you want to include a project in the federated build to
+ * consume other packages (even if it doesn't publish any package) you must
+ * declare it here.
  */
 const FEDERATED_PACKAGES = {
-	'frontend-js-react-web': ['react'],
+	'frontend-js-react-web': [
+		'classnames',
+		'formik',
+		'prop-types',
+		'react',
+		'react-dnd',
+		'react-dnd-html5-backend',
+		'react-dom',
+	],
+	'frontend-js-web': [],
+	'frontend-taglib-clay': ['@clayui/icon'],
+	'portal-template-react-renderer-impl': [],
 };
 
 /**
--- a/yarn.lock
+++ b/yarn.lock
@@ -15729,36 +15729,6 @@ webpack@^5.11.1:
     watchpack "^2.0.0"
     webpack-sources "^2.1.1"
 
-webpack@^5.4.0:
-  version "5.11.0"
-  resolved "https://registry.yarnpkg.com/webpack/-/webpack-5.11.0.tgz#1647abc060441d86d01d8835b8f0fc1dae2fe76f"
-  integrity sha512-ubWv7iP54RqAC/VjixgpnLLogCFbAfSOREcSWnnOlZEU8GICC5eKmJSu6YEnph2N2amKqY9rvxSwgyHxVqpaRw==
-  dependencies:
-    "@types/eslint-scope" "^3.7.0"
-    "@types/estree" "^0.0.45"
-    "@webassemblyjs/ast" "1.9.1"
-    "@webassemblyjs/helper-module-context" "1.9.1"
-    "@webassemblyjs/wasm-edit" "1.9.1"
-    "@webassemblyjs/wasm-parser" "1.9.1"
-    acorn "^8.0.4"
-    browserslist "^4.14.5"
-    chrome-trace-event "^1.0.2"
-    enhanced-resolve "^5.3.1"
-    eslint-scope "^5.1.1"
-    events "^3.2.0"
-    glob-to-regexp "^0.4.1"
-    graceful-fs "^4.2.4"
-    json-parse-better-errors "^1.0.2"
-    loader-runner "^4.1.0"
-    mime-types "^2.1.27"
-    neo-async "^2.6.2"
-    pkg-dir "^5.0.0"
-    schema-utils "^3.0.0"
-    tapable "^2.1.1"
-    terser-webpack-plugin "^5.0.3"
-    watchpack "^2.0.0"
-    webpack-sources "^2.1.1"
-
 websocket-driver@0.6.5:
   version "0.6.5"
   resolved "https://registry.yarnpkg.com/websocket-driver/-/websocket-driver-0.6.5.tgz#5cb2556ceb85f4373c6d8238aa691c8454e13a36"
--- a/yarn.lock
+++ b/yarn.lock
@@ -15699,6 +15699,36 @@ webpack-sources@^2.1.1:
     source-list-map "^2.0.1"
     source-map "^0.6.1"
 
+webpack@^5.11.1:
+  version "5.11.1"
+  resolved "https://registry.yarnpkg.com/webpack/-/webpack-5.11.1.tgz#39b2b9daeb5c6c620e03b7556ec674eaed4016b4"
+  integrity sha512-tNUIdAmYJv+nupRs/U/gqmADm6fgrf5xE+rSlSsf2PgsGO7j2WG7ccU6AWNlOJlHFl+HnmXlBmHIkiLf+XA9mQ==
+  dependencies:
+    "@types/eslint-scope" "^3.7.0"
+    "@types/estree" "^0.0.45"
+    "@webassemblyjs/ast" "1.9.1"
+    "@webassemblyjs/helper-module-context" "1.9.1"
+    "@webassemblyjs/wasm-edit" "1.9.1"
+    "@webassemblyjs/wasm-parser" "1.9.1"
+    acorn "^8.0.4"
+    browserslist "^4.14.5"
+    chrome-trace-event "^1.0.2"
+    enhanced-resolve "^5.3.1"
+    eslint-scope "^5.1.1"
+    events "^3.2.0"
+    glob-to-regexp "^0.4.1"
+    graceful-fs "^4.2.4"
+    json-parse-better-errors "^1.0.2"
+    loader-runner "^4.1.0"
+    mime-types "^2.1.27"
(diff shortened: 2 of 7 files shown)


izaera:
Oops, looks like I have to update the bundler too :-(.

projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
```diff
@@ -21,9 +21,24 @@ const writeWebpackFederationEntryPoint = require('./writeWebpackFederationEntryP
  * the relation is not used for anything. However, it is better to do it this
  * way, because if we mix all together we end up with a very entangled
  * configuration.
+ *
+ * Note also that when you want to include a project in the federated build to
+ * consume other packages (even if it doesn't publish any package) you must
+ * declare it here.
```
jbalsas, line 27: When you say "here" does it mean we can declare it dynamically from somewhere, or here here?

As in... `layout-taglib` uses `frontend-js-web` we need to ship a new `npm-scripts` version?

izaera:
I thought we were using `@types/webpack` but it is only pulled from `html-webpack-plugin` and because of that `tsc` is being fooled:

imagen

izaera:
Apparently nobody is using `html-webpack-plugin`. That's because we commented out its usage for the webpack 5 migration (in the first PR) given that it didn't work in v5. So I will remove this dependency and see what happens.

jbalsas:
> Apparently nobody is using `html-webpack-plugin`. That's because we commented out its usage for the webpack 5 migration (in the first PR) given that it didn't work in v5. So I will remove this dependency and see what happens.

You will break headless again, won't you? 

https://github.com/liferay/liferay-portal/blob/master/modules/apps/headless/headless-discovery/headless-discovery-web/webpack.config.js#L15

izaera:
> You will break headless again, won't you?

Mmm, in that case we may add the `html-webpack-plugin` dependency to that project (given that it's the only one using it).

What I don't understand is..., how does that work? What webpack are they using? :thinking: 

I'll check now locally. Meanwhile I will set this to draft to avoid merging.

projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
izaera: Yes, for now we need to release `npm-scripts` versions every time we change this, which is bad. And a source of problems as we have seen with previous PRs, because we need to release `npm-scripts` before we merge the changes in `liferay-portal`.

However I think (but we'll see) that we won't have the same problem again because adding more things to this object doesn't trigger any effect until we activate the federation in `liferay-portal`'s projects (at least that's the theory).

We could argue that this should be extracted ASAP, but on the other hand, we had the preset for `liferay-npm-bundler` which had the "same" problem (we needed to release a new version of it, and then a new version of npm-scripts; much worse, in fact) and nothing happened so, until we advance more in the epic and make this clearer, I think it's OK to leave it here (unless we see a big problem with it in the future).

izaera: Think of this object as the `imports` section of the default preset we use for the bundler. In fact it's that, because it serves the very same purpose.

izaera:
I can confirm the portal build fails as expected. Gonna investigate why it works and decide what to do based on that.

izaera:
OK, found why the portal build works now with the same plugin:

I removed [this](https://github.com/liferay/liferay-portal/commit/58be1bd73230f6638ce960dab18aec1e50a443b1). 

And then [this](https://github.com/liferay/liferay-portal/commit/75b493760650c8eeaca37e4a6974e6552e590c65) was added.

So it looks like the problematic plugins were `script-ext-html-webpack-plugin` and `style-ext-html-webpack-plugin`, not `html-webpack-plugin` or `mini-css-extract-plugin` per se.

Now, we have two options:

1. Move  `html-webpack-plugin` and `mini-css-extract-plugin` to `headless-discovery-web` given that its build is special and it's really the only project using those plugins.
2. Keep those two dependencies in `npm-scripts` and configure `tsc` to ignore the offending `@types/webpack`.

I would say 1 is cleaner and more logical given that `html-webpack-plugin` and `mini-css-extract-plugin` are not part of our standard build, but moving them to the project is going to decentralize those deps (though I don't think it's so horrible given that we can track them via `yarn.lock` in case someone else wants to use them in the future).

Doing 2 may bring some temporary relief, but we can have trouble in the future, because `html-webpack-plugin` is pulling typescript definitions for webpack 4 but we are using webpack 5.

Opinions @jbalsas, @wincent ?

izaera:
@jbalsas says we are going with 2 (in slack conversation). I will check how to configure `tsc`, then.

izaera:
The last commit should fix the issue. I don't know the implications for the IDE (if it is going to miss some types) but given that the bundler project is going to be replaced by webpack 5 federation and `npm-scripts` it's OK as long as it fixes the build.

izaera:
Nope :-(

wincent:
> Now, we have two options:
> 
> 1. Move  `html-webpack-plugin` and `mini-css-extract-plugin` to `headless-discovery-web` given that its build is special and it's really the only project using those plugins.
> 2. Keep those two dependencies in `npm-scripts` and configure `tsc` to ignore the offending `@types/webpack`.
> 
> I would say 1 is cleaner and more logical given that `html-webpack-plugin` and `mini-css-extract-plugin` are not part of our standard build, but moving them to the project is going to decentralize those deps (though I don't think it's so horrible given that we can track them via `yarn.lock` in case someone else wants to use them in the future).

(I see you already said that "2" is the way, but just to provide some historical linkage...)

We previously made the Headless folks [go through "The Process"](https://github.com/liferay/liferay-frontend-guidelines/issues/156) for requesting that those dev deps be added to npm-scripts because [we want a centralized mechanism for bringing dev deps into liferay-portal](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/dxp/dev_dependencies.md). Even making a small number of exceptions to the rules here takes us in a pretty scary direction, because dev dependencies often have large transitive graphs, and even an apparently small change on the surface can lead to rather large and confusing inconsistencies in the build process, under the hood. Kind of like tip-of-iceberg vs mass-of-iceberg-under-the-water.

Sadly, [two of the plugins are now unmaintained](https://github.com/liferay/liferay-frontend-guidelines/issues/156#issuecomment-716782058) and incompatible with webpack 5.

So "1" would be backtracking quite a bit. Independently of that, I am disappointed to see that https://github.com/liferay/liferay-portal/commit/75b493760650c8eeaca37e4a6974e6552e590c65 went in without us noticing it. 😢 I haven't found the Slack thread that you mentioned, but I get the feeling that, _if we consider this use case to be important_ (and it seems that Headless does), we should make our own plug-ins that work with webpack 5 (the abandoned plugins are small) and remove the liability that comes with have these third-party deps. If we wanted to keep the risk local to Headless, the plugins could be developed and live inside their project instead of in npm-scripts.

yarn.lock
```diff
@@ -15699,6 +15699,36 @@ webpack-sources@^2.1.1:
...
+    loader-runner "^4.1.0"
+    mime-types "^2.1.27"
+    neo-async "^2.6.2"
+    pkg-dir "^5.0.0"
+    schema-utils "^3.0.0"
+    tapable "^2.1.1"
+    terser-webpack-plugin "^5.0.3"
+    watchpack "^2.0.0"
+    webpack-sources "^2.1.1"
+
```
wincent, line 15731: I gather we can get rid of the duplication reported by `yarn why` by updating webpack dep in the bundler too:

```
=> Found "webpack@5.11.0"
info Has been hoisted to "webpack"
info Reasons this module exists
   - "workspace-aggregator-63f55e4a-73ca-4134-8d4a-bb0252cc0639" depends on it
   - Specified in "devDependencies"
   - Hoisted from "_project_#@liferay#npm-bundler#webpack"
   - Hoisted from "_project_#webpack"
   - Hoisted from "_project_#@liferay#js-toolkit-core#webpack"
info Disk size without dependencies: "5.97MB"
info Disk size with unique dependencies: "10MB"
info Disk size with transitive dependencies: "22.53MB"
info Number of shared dependencies: 52
=> Found "@liferay/npm-scripts#webpack@5.11.1"
info This module exists because "_project_#@liferay#npm-scripts" depends on it.
info Disk size without dependencies: "4.84MB"
info Disk size with unique dependencies: "8.88MB"
info Disk size with transitive dependencies: "21.41MB"
info Number of shared dependencies: 52
```

projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
wincent: > We could argue that this should be extracted ASAP

I think it's fine for this to be "in the wrong place", awkward, or inconvenient to update during this phase of development. The design may still change and the most important thing now is proving the viability of webpack federation by rolling it out more broadly. Once it's proven, we can worry about the ergonomics and the update flow.

yarn.lock
izaera: Yes, I [got that one](https://github.com/liferay/liferay-frontend-projects/pull/376#issuecomment-752504567) but first I tried to fix the `tsc` issue.

izaera:
This is fixed. I'm going to test it locally in the portal's build.

izaera:
Works locally. Going to test if master can be built too, to avoid problems like the ones we had in the past because the changes in `npm-scripts` were not compatible with `liferay-portal`'s master branch and required some changes to be merged. 

I will do this from now on, so that we are sure that the latest `npm-scripts` version doesn't need anything to be present in `liferay-portal` to work. This way we will honor the decoupled lifecycles of `liferay-portal` and `npm-scripts`.

izaera:
Strange: I tested locally with `liferay-portal`'s master and this version of `npm-scripts` and it failed after an `ant all`. Then built the modules by hand and it got fixed. Then ran `ant all` and it's not failing any more :shrug: .

I'm changing this from draft to prepared and will merge it if nobody has anything else to say (I have addressed all the comments so I guess it's OK to do it).

Issue #369, opened 2020-12-24
Backport Senna fix for LPS-125077 and cut v2.7.10


wincent:
As per [this Slack thread](https://liferay.slack.com/archives/CNBG06JS3/p1608248642459400), the [fix](https://github.com/brianchandotcom/liferay-portal/pull/97087) for [LPS-125077](https://issues.liferay.com/browse/LPS-125077) needs to be backported as [we described here](https://liferay.slack.com/archives/CNBG06JS3/p1606895628358800?thread_ts=1606839326.352100&cid=CNBG06JS3).

The [senna.js](https://github.com/liferay/senna.js) repo hasn't been imported here yet (ie. to `maintenance/`), but we probably should do that first, like the thread says. Then apply the fix. Then cut v2.7.10. And hopefully not touch it again, but we'll see... 🤞

wincent:
Re-opening until v2.7.10 is actually published (with luck, later today).

wincent:
Closing again, because v2.7.10 is now out: https://github.com/liferay/liferay-frontend-projects/releases/tag/senna%2Fv2.7.10

cc @holatuwol

PR #379, opened 2021-01-04
chore: merge senna.js history into maintenance/projects/senna/

wincent:
So this is related to https://github.com/liferay/liferay-frontend-projects/issues/369 which is about backporting a fix from `frontend-js-spa-web` to the Senna npm package, so that the fix can be sent to older liferay-portal branches.

This is a draft though because more work is needed (and seeing as I have a few more days of PTO, I'd probably rebase this rather than merging a stale version, so as to not make the history too confusing).

The issue is that Senna isn't really buildable on new NodeJS versions either here or in the old repo (I tried it in both places, with Node v6, v8, and v10, and also with `npm` instead of `yarn`, without success). The tricky bit is explained in https://github.com/liferay/liferay-frontend-projects/commit/61ad89d9cbdb1907c21ef9f2656467115f49e8b6 and https://github.com/liferay/liferay-frontend-projects/commit/a9973de99037005a3e766b401d747d0a8f260ef3, which I will quote here because those SHA-1s might go away if we do end up rebasing:

---

**refactor(senna): switch to Yarn**

This is the result of running `yarn` and deleting the package-lock.json file. Note that the build is horribly broken both before and after this change due to the usual node-sass shenanigans (which we partake in by virtue of using gulp-metal, which pulls in gulp-sass):

    Error: Node Sass does not yet support your current environment: OS X
    64-bit with Unsupported runtime (64)

I think I'll be able to fix this by pinning us to a later release of node-sass, but that in itself might be tricky:

- Senna depends on `gulp-metal@^1.0.0`; the latest release of gulp-meta   is 2.2.2 (from mid-2017), but Senna is still using 1.12.2 (from around the same time; I'm not sure why we didn't update back then).
- gulp-metal 1.12.2 depends on `gulp-sass@^2.0.1`, which resolves to 2.3.2 (from mid-2016), and which in turn depends on `node-sass@^3.4.2` (resolves to v3.13.1, from late-2016).
- Additionally, 2.2.2 of gulp-metal still depends on an old version of gulp-sass (`^3.1.0`). If we update to `gulp-metal@2.2.2`, we get `gulp-sass@3.2.1` (early-2018), which depends on `node-sass@^4.8.3`, but somehow we still get `node-sass@4.5.3`, which will issue the same error seen above. (No idea why that might be, but I am 100% sure it is happening, having tested it twice now after a `git clean -fdx .`)

Given that this is a dev-only dep, it would be convenient to sketchily force it to use a v4 release of node-sass via Yarn resolutions, but unfortunately resolutions aren't powerful enough to override a hard limit within a dependency range, only to select a version within such a range (ie. we can't turn a `^3.4.2` dep into a `^4.14.1` one), so we'll have to see if there is a way around this without having to do anything unseemly (like cutting a release of gulp-metal, which is hard because most of the owners on the npm package have left the company or are on vacation, or inlining that project in here etc...)

---

**chore(senna): update gulp-metal dependency to latest (2.2.2)**

As I said in the previous commit, even if I try updating gulp-metal from the one we were using (1.12.2) to the latest (2.2.2) in order to get a newer version of node-sass (eg. 4.14.1), I still get a duplicate old one (4.5.3) nested under `node_modules/gulp-metal/node_modules/node-sass/`.

It took me a long time to figure this out, but the reason is for [an npm feature — `bundledDependencies` — I didn't even remember existed](https://stackoverflow.com/questions/11207638/advantages-of-bundleddependencies-over-normal-dependencies-in-npm), and which we are abusing horribly here, to ship basically the entire dependency graph of gulp-metal as an invisible, unoverridable blob.

Yep, we really ship [all of these packages](https://github.com/metal/gulp-metal/blob/17218b70ace3f0a733b6f49ef95d581c1e5c26bf/package.json#L73-L103) *inside* gulp-metal. Observe how a `wget https://registry.npmjs.org/gulp-metal/-/gulp-metal-2.2.2.tgz` has us downloading a _50 MB archive_:

    Resolving registry.npmjs.org (registry.npmjs.org)... 104.16.22.35, 104.16.25.35, 104.16.26.35, ...
    Connecting to registry.npmjs.org (registry.npmjs.org)|104.16.22.35|:443... connected.
    HTTP request sent, awaiting response... 200 OK
    Length: 52343983 (50M) [application/x-tar]

So this is why `yarn why node-sass` swears there is only one version on the disk:

    => Found "node-sass@4.14.1"
    info Reasons this module exists
       - "gulp-metal#gulp-sass" depends on it
       - Hoisted from "gulp-metal#gulp-sass#node-sass"

And why `yarn why gulp-sass` says there's one version of that:

    => Found "gulp-sass@3.2.1"
    info Reasons this module exists
       - "gulp-metal" depends on it
       - Hoisted from "gulp-metal#gulp-sass"

When in reality we have:

- node_modules/gulp-sass at 3.2.1, asking for node-sass@^4.8.3
- node_modules/node-sass at 4.14.1
- node_modules/gulp-metal "asking for" gulp-sass@^3.1.0
- node_modules/gulp-metal/node_modules/gulp-sass at 3.1.0, "asking for" node-sass@^4.2.0
- node_modules/gulp-metal/node_modules/node-sass at 4.5.3

In addition to being invisible to `yarn why`, these hidden nested versions cannot be overridden with a `"resolutions"` field in the package.json.

So I'm putting this up here for educational purposes only; the only way to fix this for real will be to cut a gulp-metal release without this insanity (will have to wait for Chema to come back for that), or to copy the project source here and modify it before using it inline, or to rip out and replace the whole Metal-based build toolchain (sadly, that would be expensive because it is complicated and that means we'd have to pull it apart carefully to figure out which bits of it actually matter; I think Senna's build requirements are actually pretty simple underneath all the layers of abstraction and packages like `metal-tools-build-*` and so on, but it would take time to figure that out).

---

So, in short, the bit-rot is strong with this one. Can't build the project or run the tests, and we have build tooling spread across a number of packages, at least one key one of which we don't have broadly shared publishing permissions, which makes cutting a release pretty difficult.


wincent:
For anybody watching this, giving this another shot today (and maybe tomorrow 😂).

matuzalemsteles:
🤔 I would say good luck with that!!! 🤪 but why not keep those fixes still in senna.js repo as it can help or will we file there too and move here?

wincent:
> but why not keep those fixes still in senna.js repo as it can help or will we file there too and move here?

Master plan = move as much as possible into the mono-repo (there is already a deprecation notice on the Senna repo).

But more immediately, I also have trouble building `HEAD` of `master` in the Senna repo because of bit-rot in the dependency graph. So by means fair or foul, I'm going to find a way to cut a release.

matuzalemsteles:
> Master plan = move as much as possible into the mono-repo (there is already a deprecation notice on the Senna repo).

Oh okay, I saw it now, I probably missed a lot during my time in migrations...

wincent:
Ok, after much wailing and gnashing of teeth, I got local build working and `yarn ci` is green (on my machine). Let's see now if it is green in GitHub actions too. You can expected the Semantic Pull Request bot to fail, because of the old commit messages, but we can ignore that one.

wincent:
There will be more follow-up/clean-up to do tomorrow, but we can "merge" this one (really, push it directly) and continue from there. I don't want to have to redo the subtree merge each time I rebase, so I want to get this in sooner rather than later.

Stuff I want to do tomorrow:

- [ ] Go through ["Importing an external project into the monorepo"](https://github.com/liferay/liferay-frontend-projects/blob/master/CONTRIBUTING/importing-a-project.md) and see what I've missed.
- [ ] See if we can use `yarn version` to release this project like we do for the others (eg. create a `.yarnrc` etc).
- [ ] See if we can use the changelog generator on this project.
- [ ] Actually backport the bugfix that is motivating all of this work in the first place.
- [ ] Cut a release.

wincent:
To `master` she goes: https://github.com/liferay/liferay-frontend-projects/commit/12e6e346d1aa38dbd7f2aa07fe01f7547c9893df

jbalsas:
Not bad for a Tuesday!

awesome

PR #382, opened 2021-01-13
feat: provide a mechanism to extend the `remotes` and `shared` fields of webpack config

izaera:
See [IFI-2310](https://issues.liferay.com/browse/IFI-2310).

The changes here are needed to build and deploy [this branch](https://github.com/izaera/liferay-portal/tree/LPS-125746).

I will release a new version of `npm-scripts` after this is merged which, BTW, should be the last that we need unless we add new features or fix bugs, because I have added all information contained in the old default preset (see [this commit](https://github.com/liferay/liferay-frontend-projects/commit/c380ba0e1fb1e4f8c3b9e168a7d3b1b667eb75cd)) given that it doesn't fail the build. In addition, it has been made extensible, so..., that should be enough.

As always this has been tested locally with the linked private branch of `liferay-portal`.

I will send the portal's PR as soon as we release a new version of npm-scripts.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/package.json
+++ b/projects/npm-tools/packages/npm-scripts/package.json
@@ -51,6 +51,7 @@
 		"metal-tools-soy": "4.3.2",
 		"mini-css-extract-plugin": "0.11.2",
 		"minimist": "^1.2.0",
+		"path-browserify": "^1.0.1",
 		"prettier": "^2.1.2",
 		"react": "*",
 		"react-dom": "*",
--- a/projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
@@ -127,5 +127,10 @@ module.exports = async function () {
 					}, {}),
 			}),
 		],
+		resolve: {
+			fallback: {
+				path: require.resolve('path-browserify'),
+			},
+		},
 	};
 };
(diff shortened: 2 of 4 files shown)


projects/npm-tools/packages/npm-scripts/package.json
```diff
@@ -51,6 +51,7 @@
 		"metal-tools-soy": "4.3.2",
 		"mini-css-extract-plugin": "0.11.2",
 		"minimist": "^1.2.0",
+		"path-browserify": "^1.0.1",
```
izaera, line 54: I'm sure @wincent will love this :heart_eyes:

projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
```diff
@@ -9,37 +9,153 @@ const {
...
 const createTempFile = require('./createTempFile');
+const getMergedConfig = require('./getMergedConfig');
 const parseBnd = require('./parseBnd');
 const writeWebpackFederationEntryPoint = require('./writeWebpackFederationEntryPoint');
 
 /**
- * This object must represent the current configuration status of the portal's
- * project if we don't want to break the build. It contains, for each
- * federation-enabled project, the packages that are federated.
+ * The following two arrays are roughly equivalent to the old default preset
```
izaera, line 17: Turns out that I was wrong in the previous PR and it didn't make too much sense to mix imports and federation because the model is a bit different.

yarn.lock
```diff
@@ -11804,6 +11804,11 @@ passwd-user@^3.0.0:
   dependencies:
     execa "^1.0.0"
 
+path-browserify@^1.0.1:
```
izaera, line 11807: Mmmm, I love the smell of new dependencies in the morning....

:helicopter: :fire:

izaera:
As agreed with @jbalsas in a slack conversation I'm merging this to avoid blocking the server side PR. If we want to add changes later we can do it without any problem.

PR #387, opened 2021-01-25
feat: allow extension of webpack `exposes`

izaera:
This is needed for LPS-126503 because we need to export JS endpoint to be consumed from `<react:component>`.

I've tested this locally.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
@@ -179,8 +179,13 @@ module.exports = async function () {
 
 	await writeWebpackFederationEntryPoint(mainFilePath);
 
-	let {remotes, shared} = getMergedConfig('npmscripts', 'federation');
+	const config = getMergedConfig('npmscripts');
 
+	const {build, federation} = config;
+
+	let {exposes, remotes, shared} = federation;
+
+	exposes = exposes || [];
 	remotes = remotes || [];
 	shared = shared || [];
 
@@ -220,6 +225,7 @@ module.exports = async function () {
 		plugins: [
 			new ModuleFederationPlugin({
 				exposes: {
+					...prefixExposesPaths(exposes, `./${build.input}/`),
 					'.': mainFilePath,
 				},
 				filename: 'container.js',
@@ -256,3 +262,15 @@ module.exports = async function () {
 		},
 	};
 };
+
+function prefixExposesPaths(exposes, prefix) {
+	return exposes.reduce((exposes, filePath) => {
+		const exposeName = path.posix
+			.relative('.', filePath)
+			.replace(/\.js$/i, '');
+
+		exposes[exposeName] = `${prefix}${filePath}`;
+
+		return exposes;
+	}, {});
+}
--- a/projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
@@ -263,13 +263,19 @@ module.exports = async function () {
 	};
 };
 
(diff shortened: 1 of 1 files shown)


projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
```diff
@@ -256,3 +262,15 @@ module.exports = async function () {
 		},
 	};
 };
+
+function prefixExposesPaths(exposes, prefix) {
```
wincent, line 266: A more generic name might be appropriate here, like `transformExposesPaths` or `normalizeExposesPaths`, because `prefix` makes it sound like it's simply going to _add_ to the front of each path, but what it actually does is more more complicated than that (ie. it returns an object, it strips the `.js` extension from the keys, strips the leading `./` from the keys if it is present, and then create a value out of `prefix` and the `filePath`).

Should we throw an error if user passes in a `./` path? I'm thinking of the kind of confusion we talked about [in Slack](https://liferay.slack.com/archives/G3JBR21HA/p1611569317122800) where users can write their `resources/`-relative paths as either `a/b/c` or `./a/b/c`... (forcing them to write it always the same way might be a good thing).

projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
```diff
@@ -220,6 +225,7 @@ module.exports = async function () {
 		plugins: [
 			new ModuleFederationPlugin({
 				exposes: {
+					...prefixExposesPaths(exposes, `./${build.input}/`),
```
wincent, line 228: Do we need the `./` here? I believe it won't change the behavior of `path.relative`.

izaera: That is for webpack not us :-). We invoke `path.relative` below to make sure we strip any leading `./` (that's `path.relative` convention for its return value) and then we prepend `./` for webpack (see [webpack docs](https://webpack.js.org/concepts/module-federation/#troubleshooting) for more detail; there you can see the leading `./` convention in file paths).

projects/npm-tools/packages/npm-scripts/src/utils/createFederationConfig.js
izaera: I didn't like the name either, but couldn't think of anything better (because of _webpack fatigue_ :sweat_smile: ) so I will use your `normalizeExposesPaths` proposal.

Regarding throwing an error I don't like it too much, because it looks like an anti-feature. I mean, in a external environment (different from`liferay-portal`) it sounds quite reasonable to support both syntaxes (think of a customers that _loves prefixing everything with dots_ using `npm-scripts`). 

So I don't know. Maybe we could move it to the linting layer?

wincent: I don't think I'd move it to the linting layer. Either we decide it is "wrong" (in which case we should reject it outright):

reject

Or we decide we don't care (in which case we allow it):

i-ll-allow-it

(Note that we treat almost all our lint problems as hard errors rather than warnings, because mere warnings/recommendations tend to get ignored and accumulate over time.)

izaera: Finally renamed it `transform...` :sweat_smile: .

Regarding the check for `./` I will make it mandatory to use `<inputDir>` before the path to make it more explicit. The reason is that, given that we are composing the path from the `build.input`+ `the path the user is configuring`, having a leading `./`is very strange, because it looks like you are referring to the project folder, not `build.input`.

As decided in Slack, we will slowly shift to this paradigm in future PRs so that all the configuration is easily understood by any untrained user.

PR #389, opened 2021-01-29
feat(eslint-config): add Cavanaugh 7.4.x to list of possible deprecations

kresimir-coko:
Fixes #388 

It adds `Cavanaugh (7.4.x)` to the list of possible deprecation versions, adds it to the docs, and adds it to the test.
I ran `yarn ci` and it was green, as all CI's should be.

I didn't mess around with the versioning thing, as I saw some other changes were in the list, if I needed to do that LMK so I can do it.

PR #391, opened 2021-02-03
docs: link to general guidelines and list recommended scopes

wincent:
Based on [conversation here](https://github.com/liferay/liferay-frontend-projects/pull/389#issuecomment-770646774) which was about not finding contributing guidelines for how to format commit messages.

We could still use some more specific docs about what scopes we'd like to see when working in the monorepo (eg. using scopes to indicate project), so I added that in the second commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,3 +21,23 @@ These apply to any work at Liferay, including in this monorepo:
 -   [Creating a project](CONTRIBUTING/creating-a-project.md)
 -   [Importing a project](CONTRIBUTING/importing-a-project.md)
 -   [Migrating an npm package to the `@liferay` named scope](CONTRIBUTING/migrating-an-npm-package-to-the-liferay-named-scope.md)
+
+## Conventional commit scopes
+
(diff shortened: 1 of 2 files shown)


wincent:
@kresimir-coko: in an ideal world, we'd mention the scope in every single project-specific `CONTRIBUTING.md`, but I am too lazy for that (when we moved projects into the monorepo, each project came in with its own `CONTRIBUTING.md` and we only lightly edited it to reflect the new reality... updating paths and such).

And note that I didn't include a scope in this PR because it isn't specific to any project, and it isn't what I'd call "cross-cutting" either (ie. touching multiple projects). It's just a tiny doc tweak, and it doesn't touch any paths that would cause it to show up in project-specific changelogs, so it doesn't need a scope.

kresimir-coko:
@wincent  Awesome! One more thing I noticed right now, in the [Liferay Frontend Guidelines](https://github.com/liferay/liferay-frontend-projects/tree/master/guidelines), the first 2 links (bolded) are pointing to a 404: 
> This is a live (changing) repository containing **general** and Liferay **DXP-specific** guidelines for doing Frontend Development at Liferay Inc.

wincent:
Thanks for catching the 404s! Will update.

PR #392, opened 2021-02-05
docs(guidelines): provide more info about narrowing suppression scope

wincent:
Seeing as [this came up in code review over here](https://github.com/liferay-frontend/liferay-portal/pull/746#discussion_r570879556).

Funnily enough, we may want to make a lint for lint suppressions 😂 — not sure if ESLint can already be configured to have a kind of "strict mode" like this, or we'd have to cook something up ourselves. 🤔

cc @kresimir-coko


wincent:
Trivial so just going to merge it.

Issue #394, opened 2021-02-08
"Maintain" a forked copy of jquery-form in this repo


wincent:
See discussion over on [this PR](https://github.com/liferay-frontend/liferay-portal/pull/699) and [in Slack](https://liferay.slack.com/archives/C3JBR21HA/p1612519869209500) (private link).

Basically, we need to patch [jquery-form](https://github.com/jquery-form/form) and forking a copy into our monorepo seems the least horrible way to do it. Other options considered:

- [Original PR](https://github.com/brianchandotcom/liferay-portal/pull/97915) to Brian just patched the code inline in liferay-portal but was rejected.
- [Draft PR from me](https://github.com/wincent/liferay-portal/pull/372) checking in a patch and applying it via Gradle; all too fiddly.
- Doing a literal GitHub fork, or a "fork-plus-more-stuff" in a separate repo (like we have been doing in [liferay-ckeditor](https://github.com/liferay/liferay-frontend-projects)); rejected because we don't like the diaspora affect of having yet another scattered repo.
- Patching a JAR using [the Gradle `PatchTask`](https://github.com/liferay/liferay-portal/blob/19d993d6e3c0dd6865924bf23f75992f52fc355e/modules/sdk/gradle-plugins-patcher/src/main/java/com/liferay/gradle/plugins/patcher/PatchTask.java); rejected because shipping npm packages via JARs seems, er, "suboptimal".

So, basic plan is to dump a copy into the repo here undo a "third-party" directory. ie. We won't apply the usual linting/formatting/testing; these projects will live off in their own separate world, even more isolated than the stuff in the `maintenance/` directory (which at least uses shared linting/formatting etc), and it will be for stuff that we don't intend to update much... If that changes for any project, and we realize we need to update a project more invasively or frequently, we have the option of turning it into a "real" project and opting in to linting/formatting/testing etc. At that point it really becomes a hard fork. We can cross that bridge when we come to it.

wincent:
Ok, so I [cut a release](https://github.com/liferay/liferay-frontend-projects/releases/tag/jquery-form%2Fv3.51.0-liferay.1) and will presently send an PR to liferay-portal updating to use it. As the work in this repo is done now, closing this.

wincent:
Upstream PR got merged: https://github.com/brianchandotcom/liferay-portal/pull/98619

PR #395, opened 2021-02-08
chore: merge jquery-form history into third-party/projects/

wincent:
Just a draft PR to see if CI likes the merge here.

This is for https://github.com/liferay/liferay-frontend-projects/issues/394 — if CI is good, I'll push directly to `master` (no merge commit) and then apply the patch on top of there.

Details (from third-to-last commit) follow:

---

#### chore: merge jquery-form history into third-party/projects/

In service of [#394](https://github.com/liferay/liferay-frontend-projects/issues/394).

Basically, we want to pull in the code from this third-party project (which is mostly on life-support at this point) and apply a small patch on top. We don't anticipate having to change it much, if ever, so we are sticking it under "third-party/projects/" and we won't cover it with our normal linting/formatting/testing.

Note that because everything else in the repo pulls in history via subtree merge, doing it here as well. That way if we ever need to `git blame` in here we'll have something useful to look at.

Note also that in liferay-portal we are using v3.51 of this plugin and have no plans to switch to the latest (v4.3.0) at this time, but I am still pulling in and re-writing all of the tags so that we can easily merge the newer stuff in in the future if we want to.

Steps follow.

Grab the code:

    git remote add -f jquery-form https://github.com/jquery-form/form.git

Make directory for it. Note that it is going to live at "third-party/projects/jquery-form/" and not "third-party/jquery-form/"; we want to reserve the top-level inside of "third-party/" for any future tooling that we might want to share across projects, or any common READMEs or other docs:

    mkdir -p third-party/projects

Create annotated tags seeing as all the tags in the project are unannotated:

    for TAG in \
      3.25.0-2013.01.18 \
      3.32.0-2013.04.09 \
      3.36 \
      3.37.0 \
      3.45.0 \
      3.46.0 \
      3.49 \
      3.50 \
      3.51; do
      git tag -f -a -m jquery-form/v$TAG $TAG $TAG
    done

    for TAG in \
      v3.19 \
      v4.0.0 \
      v4.0.1 \
      v4.1.0 \
      v4.2.0 \
      v4.2.1 \
      v4.2.2 \
      v4.3.0; do
      git tag -f -a -m jquery-form/$TAG $TAG $TAG
    done

Rewrite tags to have a prefix. First tags starting with "v" and then those without a leading "v":

    support/filter-tags.sh jquery-form/master 'sed "s#^v#jquery-form/v#"'
    support/filter-tags.sh jquery-form/master 'sed "s#^\([0-9]\)#jquery-form/v\1#"'

Do the actual subtree merge:

    git merge -s ours --no-commit --allow-unrelated-histories jquery-form/v3.51
    git read-tree --prefix=third-party/projects/jquery-form -u jquery-form/v3.51
    git commit -m 'chore: merge jquery-form history into third-party/projects/'

Make sure shape of DAG is sane:

    git log --oneline --decorate --graph --all

Push rewritten tags so that entire history will be reachable/explorable:

    git push upstream-rw --tags --dry-run
    git push upstream-rw --tags

Kill temporary remote that was added earlier. We don't want this lying around because a careless `fetch` could pull down all the tags that we don't want (because they aren't prefixed, or annotated, or even consistent), and it is all too easy to accidentally contaminate the upstream with these tags by a push. If we ever need to resync with upstream history again, we can just re-add the remote:

    git remote remove jquery-form

Amend commit to make these notes that you're reading now:

    git commit --amend

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -26,4 +26,5 @@
 # Third-party code
 #
 
+/third-party/projects
 node_modules
--- a/.prettierignore
+++ b/.prettierignore
@@ -24,3 +24,9 @@
 /copyright.js
 /maintenance/projects/**/copyright.js
 /projects/**/copyright.js
+
+#
+# Third-party.
+#
+
+/third-party/projects
--- a/LICENSES/GPL-2.0.txt
+++ b/LICENSES/GPL-2.0.txt
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.  This
+General Public License applies to most of the Free Software
+Foundation's software and to any other program whose authors commit to
+using it.  (Some other Free Software Foundation software is covered by
+the GNU Lesser General Public License instead.)  You can apply it to
+your programs, too.
+
+  When we speak of free software, we are referring to freedom, not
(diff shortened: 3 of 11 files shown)


wincent:
Semantic PR failure is expected (because of old commit messages), but all the others are green, so yay. Going to push a couple small README/doc fixes on top of this, but I think we're good. Closing this draft for now.

Issue #14, opened 2020-09-24
Enforce use of "strict equality"


wincent:
For safety, as [I recommended in this PR](https://github.com/liferay/liferay-js-toolkit/pull/542/commits/b88b29470bb037e05a7c8407e79f72d2f6695423#r390321880), we should just use `===`/`!==` everywhere instead of `==`/`!=`. The single exception I'd make to this rule is that `x == null` is ok (convenient short-hand for `x == null || x == undefined`, and it should come up pretty rarely).

PR #396, opened 2021-02-08
feat: add soy resolver and scanExposes utility target

izaera:
We need to stabilize webpack based builds:

- The soy resolver is needed to find compiled soy templates from webpack
- The `scanExposes` is an initial implementation of a tool to find JS modules being used from JSPs so that developers may configure them in the `exposes` configuration field of webpack federation plugin.
- Finally, I have optimized the webpack federation entry point to avoid the need for JS code transformation which was failing when some features of ES were used.

I've tested it locally with today's master branch of `liferay-portal` with no apparent failures.

In any case, I won't release any version yet until I really need it for `liferay-portal`.


projects/npm-tools/packages/npm-scripts/src/index.js
```diff
@@ -32,6 +32,10 @@ module.exports = async function () {
 			await require('./scripts/prettier')(...ARGS_ARRAY.slice(1));
 		},
 
+		async scanExposes() {
```
jbalsas, line 35: Is there a more common name for this in our already existing Gradle plugins or tasks? `scanExposes` Something with `dependencies` or something like that?

izaera: There's nothing like this anywhere (that I know of).

But we can still call it something more generic (like `scanDependencies` as you propose) in case we want to add more things in the future. For example, if we decide that this will not only scan needed `expose`s but also needed `remote`s, we can add that feature here.

jbalsas: Yeah, I was thinking this is also kind of the things SF (or bndtools) does, so it could make sense that this would actually be integrated in the "check"/"format" scripts rather than being an independent tool.

izaera: :thinking: Not sure about that, because the algorithm is too simple to assume that it will always be correct.

It will -for sure- miss some necessary cases (for example: dynamic strings containing module names, modules used in Java, etc.) but then, it can also show some false positives for any taglib that uses an attribute called `module` or `propsTransformer`.

It is possible that we may fix false positives by looking if the name is a valid `require`able module but we would need to check. As it is now, I consider it too fragile to be included in a SF.

But yeah, would be great to be able to have a SF covering as much use cases as possible.

izaera: I'm thinking now that we may also define some kind of SF exclusion for cases where false positive shows up.

izaera: As I suspected, this is not as easy as we may think. If we want to scan JSPs and fail the SF, we need to make sure that we correctly parse the taglibs and that's not an easy task because a tag invocation may have `<` and `>` signs inside the attribute values so we must parse attributes too. 

Think of this code:

```JSP
<my:taglib module="<%= myModule %>" message="Use < and > to move inside the UI">
     var module = "<>";
     ...
</my:taglib>
```

There are many places where a simple parser would fail but we need to make sure we don't fail in those cases.

So I will delay this task for a following PR because we need the other changes to continue with the epic.

We have [IFI-2327](https://issues.liferay.com/browse/IFI-2327) for this task, so I will revert it to open and remove the commit from this PR to be able to merge this.

izaera: Also, for the record, @wincent suggested in a Slack conversation that this could also be done in the Java SF given that currently we only lint JS code inside the JSP files, not the files theirselves.

However, given that this will probably be needed for builds other than `liferay-portal` (GS, customers, etc.) I think we will need to move it here anyway.

wincent: > this is not as easy as we may think

Indeed. We face the same problem with linting and formatting JS inside JSP — you can have arbitrary ~garbage~ JSP tags/expressions/scriptlets and EL expressions mixed in _anywhere_. Most of the complexity in the JS-in-JSP handling stems from dealing with that.

izaera:
Set to draft because I'm finally working on the scanExposes feature (given that we had a lexer for JSPs that can be leveraged to do the job).

izaera:
This is ready for review.

I've tested it in `document-library` and it works. It may be possible that there are some corner cases I'm not covering, but since this will be used in future portal PRs, we will polish it then, as those corner cases appear. I think that will be way more efficient than trying to find all of them now.

@wincent Can you please review the lexer parts? I've extracted the JSP DSL to a file so that we can reuse it from several lexers.

projects/npm-tools/packages/npm-scripts/src/jsp/checkExposesConfiguration.js
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		const modulePath = `${moduleName}.js`;
+
+		if (!fs.existsSync(path.join(config.build.input, modulePath))) {
+			continue;
+		}
+
+		if (!exposes.includes(`<inputDir>/${modulePath}`)) {
+			messages.push({
```
izaera, line 80: I'm mimicking eslint messages here because our linter assumes that everything is linted by eslint.

We can probably turn the check-exposes code into an eslint plugin, but I have no experience with them so I wanted to first settle what we have and then we may enhance it if we think it is necessary.

However, I don't know what others may think... Feedback welcome.

wincent: > We can probably turn the check-exposes code into an eslint plugin, but I have no experience with them so I wanted to first settle what we have and then we may enhance it if we think it is necessary.

I think mimicking the message format is fine. You don't want a real ESLint plugin here because you're not linting ES.

(That said, I haven't read the code yet. I'm about to go and do that now.)

projects/npm-tools/packages/npm-scripts/src/jsp/checkExposesConfiguration.js
```diff
@@ -0,0 +1,93 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/* eslint-disable no-console */
```
wincent, line 6: Lint suppressions should be made as locally as possible (see [guidelines](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/dxp/linting.md#lint-suppressions)), because it's too easy for an overly broad suppression to end up allowing in actually-bad stuff unintentionally.

In this case, you don't even have any `console` usages in this file, so it's not even a question of scoping this more narrowly; just remove it.

projects/npm-tools/packages/npm-scripts/src/jsp/checkExposesConfiguration.js
```diff
@@ -0,0 +1,93 @@
...
+	const invocations = getTaglibInvocations(source, ...Object.keys(TAGLIBS));
+
+	const moduleNames = new Set();
+
+	for (const invocation of invocations) {
+		const {attributes: validAttributes} = TAGLIBS[invocation.tag];
+		const {attributes} = invocation;
+
+		for (const validAttribute of validAttributes) {
+			const require = attributes[validAttribute];
```
wincent, line 49: Shadowing `require` here is a bit 🤮  — can you rename this? Maybe just `attribute`?

projects/npm-tools/packages/npm-scripts/src/jsp/checkExposesConfiguration.js
```diff
@@ -0,0 +1,93 @@
...
+
+			if (!require) {
+				continue;
+			}
+
+			require
+				.split(',')
+				.map((part) => part.split(' as ')[0])
```
wincent, line 57: I don't know how sensitive these taglibs are to whitespace, but we could pretty easily be whitespace-agnostic at this layer at least. That way, even if somebody hasn't run SF yet we can detect more cases.

```js
.split(/\s*,\s*/)
.map((part) => part.split(/\s+as\s+/)[0])
```

projects/npm-tools/packages/npm-scripts/src/jsp/checkExposesConfiguration.js
```diff
@@ -0,0 +1,93 @@
...
+	const messages = [];
+
+	const federation =
+		typeof config.federation === 'object' ? config.federation : {};
+	const exposes = federation.exposes || [];
+
+	for (const moduleName of moduleNames) {
+		if (moduleName.includes('<%=')) {
+			continue;
+		}
```
wincent, line 71: 🤦‍♂️ — fun times

projects/npm-tools/packages/npm-scripts/src/jsp/getTaglibInvocations.js
```diff
@@ -0,0 +1,110 @@
...
+
+			if (to === -1) {
+				to = jsp.indexOf('/>', from + 1 + tag.length);
+
+				if (to === -1) {
+					break;
+				}
+			}
+			else {
+				to += 3 + tag.length;
```
wincent, line 77: So the `3` here corresponds to what exactly? Is there a way you can write this that makes it explicit?

projects/npm-tools/packages/npm-scripts/src/jsp/getTaglibInvocations.js
```diff
@@ -0,0 +1,110 @@
...
+	return [...lexer.lex(source)];
+}
+
+module.exports = function (jsp, ...tags) {
+	const invocations = [];
+
+	for (const tag of tags) {
+		let from = 0;
+
+		while ((from = jsp.indexOf(`<${tag}`, from)) != -1) {
```
wincent, line 66: Do `!==` here instead of `!=` because the latter [will be a lint error](https://github.com/liferay/liferay-frontend-projects/issues/14) once we turn on the rule.

projects/npm-tools/packages/npm-scripts/src/jsp/getTaglibInvocations.js
```diff
@@ -0,0 +1,110 @@
...
+			let to = jsp.indexOf(`</${tag}>`, from + 1 + tag.length);
```
wincent, line 67: Here again it would be nice if we could be whitespace-agnostic (ie. `</ blah>` vs `</blah>`), but it's up to you how robust you want to be.

projects/npm-tools/packages/npm-scripts/src/jsp/getTaglibInvocations.js
```diff
@@ -0,0 +1,110 @@
...
+
+				if (to === -1) {
+					break;
+				}
+			}
+			else {
+				to += 3 + tag.length;
+			}
+
+			const jspFragment = jsp.substring(from, to);
```
wincent, line 80: So here I'm obviously thinking about nested tags. Now given the tags that you're passing in from `TAGLIBS`, none of those can ever be nested recursively, can they? I don't think they can, but just wanted to check with you.

projects/npm-tools/packages/npm-scripts/src/jsp/getTaglibInvocations.js
```diff
@@ -0,0 +1,110 @@
...
+			const attributeTokens = tokens.filter(
+				(token) => token.name === 'ATTRIBUTE'
+			);
+
+			for (const token of attributeTokens) {
+				const {contents} = token;
+
+				const i = contents.indexOf('=');
+				const name = contents.substring(0, i);
+				const value = contents.substring(i + 2, contents.length - 1);
```
wincent, line 95: Why the `2` here? Is that `=` plus a quote? If so, is it guaranteed that there'll always be a quote (and nothing else)?

projects/npm-tools/packages/npm-scripts/src/jsp/jspDSL.js
```diff
@@ -0,0 +1,625 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+module.exports = function (api) {
```
wincent, line 6: I don't think `jspDSL` is the right name for this file. A DSL is a Domain-Specific Language. This file exports something like "matchers for JSP tokens". So maybe it should be called something like `getJSPTokenMatchers.js`?

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
```diff
@@ -0,0 +1,109 @@
...
+	const update = args.includes('-u');
+	const verbose = args.includes('-v');
+
+	const exposes = scanJSPs(input, verbose);
+
+	if (verbose) {
+		console.log('');
+	}
+
+	console.log(exposes);
```
wincent, line 29: Why log this, and why isn't it under the `verbose` guard? As far as I can tell, this is an example of a log that shouldn't exist (ie. because you needed it only during development to convince yourself that it worked; but in normal usage, there's no reason for any developer to activate this), and we have an (overly broad) lint suppression that allows it to remain. I think you should delete it all.

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
```diff
@@ -0,0 +1,109 @@
...
+ * This library is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation; either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ */
```
wincent, line 66: I don't think this is going to work. You have to use the local `copyright.js` file because if this ever runs under `modules/dxp/`, then this is going to be the wrong header (and ESLint will come along and then add the right header _as well_).

Annoying, I know, but it's not going to be too hard to find the right file and read it.

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
```diff
@@ -0,0 +1,109 @@
...
+
+			config +=
+				'module.exports.federation = module.exports.federation || {};\n';
+			config += 'module.exports.federation.exposes = [\n';
+
+			exposes.forEach((expose) => {
+				config += `  '${expose}',\n`;
+			});
+
+			config += '];';
```
wincent, line 43: Hm. This seems like it could mangle an existing config file pretty badly (or least, in a way that no human would ever write it). For example, if I already have a `federation` config, this will overwrite it, leaving the other one behind. We have a few examples of AST-based autofixes in this repo that you could use as a model if you wanted to do it "The Right Way™".

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
```diff
@@ -0,0 +1,109 @@
...
+
+function scanJSPs(dir, verbose) {
+	const exposes = [];
+
+	const items = fs.readdirSync(dir);
+
+	items.forEach((item) => {
+		const itemPath = path.join(dir, item);
+
+		if (fs.statSync(itemPath).isDirectory()) {
```
wincent, line 89: You probably shouldn't make a separate `statSync` call for each item. When you call `readdirSync` you can pass [the `withFileTypes` option](https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options) and get the info "for free".

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
```diff
@@ -0,0 +1,109 @@
...
+			exposes.push(...scanJSPs(itemPath));
+		}
+		else if (itemPath.toLowerCase().endsWith('.jsp')) {
```
wincent, line 92: Probably need to deal with `'.jspf'` as well; that's what we do with all the other formatting/linting stuff.

projects/npm-tools/packages/npm-scripts/src/utils/createWebpackSoyResolver.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * A webpack resolver plugin to instruct webpack to get compiled Soy templates
+ * from our output folder.
+ *
+ * The compiled templates are there because our compileSoy() target is run first
+ * in the build.
+ *
+ * We cannot use standard Soy loaders because:
+ *
+ * 1) There's only from Atlassian currently maintained
```
wincent, line 15: Grammar fix needed here.

projects/npm-tools/packages/npm-scripts/src/jsp/checkExposesConfiguration.js
izaera: :-( I forgot to remove old code because this was a standalone script before the refactor. I will remove it. This applies to some other revisions below as well.

Sorry.

projects/npm-tools/packages/npm-scripts/src/jsp/checkExposesConfiguration.js
izaera: :roll_eyes:

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
izaera: This file was intended to be removed. I'll push another commit to do it.

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
izaera: This file was intended to be removed. I'll push another commit to do it.

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
izaera: This file was intended to be removed. I'll push another commit to do it.

projects/npm-tools/packages/npm-scripts/src/scripts/scanExposes.js
izaera: This file was intended to be removed. I'll push another commit to do it.

projects/npm-tools/packages/npm-scripts/src/utils/createWebpackSoyResolver.js
izaera: Apparently :sweat_smile:

projects/npm-tools/packages/npm-scripts/src/jsp/getTaglibInvocations.js
izaera: It is `</` and  `>`. I've pushed a commit with the explanation in a comment.

projects/npm-tools/packages/npm-scripts/src/jsp/getTaglibInvocations.js
izaera: No. They cannot be nested.

Also take into account that this linter won't be able to cover all cases, so it's just a "best effort" type of check. We are leaving out JS modules inside Java and dynamic JS modules, for example. So, I don't think it makes sense to put excessive effort in this feature as the benefit will be insignificant compared to the complexity of the task. I would say handling the happy path is enough...

izaera: Also, I prefer to miss some valid cases rather than showing false positives :grimacing: . It's not this case, but just wanted to explain which where my assumptions when writing this feature...

projects/npm-tools/packages/npm-scripts/src/jsp/getTaglibInvocations.js
izaera: Yes, it's for the quote. I added a commentary in the code.

It is guaranteed that there will always be a quote because the lexer removes whitespace already (it returns it as a token that this method simply ignores).

izaera:
I'm setting this to draft because I'm unsure if merging and releasing this for `liferay-portal` may break anything. It was safe in the beginning, but with the latest changes this situation may have changed :grimacing: .

izaera:
I'm closing this because of the reasons explained in the [JIRA issue](https://issues.liferay.com/browse/IFI-2327), in the description section.

I will extract the commits not related to the linter and resend a new PR to continue with the webpack migration epic.

PR #397, opened 2021-02-08
chore(jquery-form): prepare to release jquery-form under named scope

wincent:
Our first release is going to be `@liferay/jquery-form` v3.51.0-liferay.1, and is going to correspond to the same code as upstream, but with the patch originally proposed in https://github.com/liferay-frontend/liferay-portal/pull/699 applied to it. I'll send a separate PR for that in a minute. Subsequent releases, if any, will be `@liferay/jquery-form` v3.51.0-liferay.2, v3.51.0-liferay.3 etc. ie. the same versioning scheme we employ in [liferay-ckeditor](https://github.com/liferay/liferay-ckeditor) — the idea is that you can see at a glance what upstream version we're using as a base, and on top of that you can tell how many patches/tweaks we've applied on top.

### DIFF
--- a/third-party/projects/jquery-form/package.json
+++ b/third-party/projects/jquery-form/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "jquery-form",
+  "name": "@liferay/jquery-form",
   "title": "jQuery Form Plugin",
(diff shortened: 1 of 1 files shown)


wincent:
Simple enough to merge. To make sure all this works I need to try releasing.

PR #400, opened 2021-02-10
feat(npm-scripts): add support for copying css filed and RTL conversion

bryceosterhaus:
@jbalsas you mentioned you a ticket about this earlier but wasn't able to locate it. Let me know if you want me to create one.

This PR should add support
- copying plain css files, we used to only copy scss
- RTL conversion for both scss and css files. This will auto-generate `*_rtl.css` files which the RTL servlet will pick up instead of converting at run time.


projects/npm-tools/packages/npm-scripts/src/sass/build.js
```diff
@@ -4,6 +4,7 @@
  */
 
 const fs = require('fs');
+const {r2} = require('liferay-theme-tasks');
```
izaera, line 7: I got scared. I thought this was the main `build.js` when I started looking at it :sweat_smile:

wincent (commented):
> @jbalsas you mentioned you a ticket about this earlier but wasn't able to locate it.

This is the only one I know of: https://github.com/liferay/liferay-frontend-projects/issues/249

All the path manipulation makes my brain hurt, so I guess we just have to trust that your tests prove this works. 😂

I'd be ok with merging this, _but_ are you sure you need those lockfile changes? I wouldn't expect to see any changes at all seeing as you didn't touch any `package.json` anywhere.

bryceosterhaus:
@wincent good catch on the yarn.lock. I'm not entirely sure why it was generating... I checked out locally and tried to fix it by running `yarn && npx yarn-deduplicate yarn.lock` but still produced the same output. So I ended up just removing the changes to the yarn.lock in that commit manually. Any idea why yarn.lock  would've updated in the first place? I've been jumping back and forth node versions for a different issue, wasn't sure if using an different version could cause it to get out of wack.

_Sidenote, I need to get better at looking at at total list of files changed before sending pr's, seems like I occasionally slip in a yarn.lock change when I didn't intend to._

bryceosterhaus:
Updated commit without the yarn.lock changes

wincent (approved):
> Any idea why yarn.lock would've updated in the first place? I've been jumping back and forth node versions for a different issue, wasn't sure if using an different version could cause it to get out of wack.

It's hard to say, but I would think that in decreasing order of variables likely to trigger an "aberrant" lockfile change we'd have: Yarn version > platform differences (Windows vs Linux/macOS) > Node version.

To avoid headaches, I try to switch things around as little as possible, but it's not always possible...

> _Sidenote, I need to get better at looking at at total list of files changed before sending pr's, seems like I occasionally slip in a yarn.lock change when I didn't intend to._

Sure would be nice if GitHub didn't _literally hide the darn file_ from the diff view. 😂

Anyway, this lg2m.

fire-at-will-commander

PR #411, opened 2021-02-19
fix(npm-scripts): fix incorrect versions in package.json

izaera:


wincent:
Screenshot 2021-02-19 -142358-l1lZNuA1@2x

What's not to like about this?

`#RetrospectiveApproval`

approved

izaera:
dance

Issue #409, opened 2021-02-19
Write guidance on how to write functions that take options with defaults


wincent:
Via [this thread](https://github.com/liferay/liferay-frontend-projects/pull/408#discussion_r579026927):

> We don't have this documented in the guidelines anywhere yet, but a pattern we often use for dealing with options is [the one seen here](https://github.com/liferay/liferay-portal/blob/e4f606604ab9882875282c3167c4f56792697d1f/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/util/format_storage.es.js#L15-L43).
> 
> The problem with the `options = options || {...}` approach is twofold. One thing is that it would be better written using a default parameter:
> 
> ```js
> function createTempFile(filename, contents, options = {autoDelete: true}) {
>     // ...
> }
> ```
> 
> The other is that both the form in this PR and the form that uses a default parameter make it impossible to pass "partial" options. As in, just say I did `createTempFile(filename, content, {hello: 'friend'})`, `options.autoDelete` would end up being `undefined` (ie. falsey) even though the intent is clear that it should default to true.
> 
> The pattern that avoids this is:
> 
> ```js
> const DEFAULT_OPTIONS = {
>     autoDelete: true,
> };
> 
> function createTempFile(filename, content, options = {}) {
>     const {autoDelete} = {...DEFAULT_OPTIONS, ...options};
> }
> ```
> 
> Obviously not a deal-breaker for this PR, but I'm going to create an issue to remind us to document this pattern at some point.

Ah, and another option that I forgot to mention that is probably even better in a lot of cases is to rely on destructing with default values:

```js
function createTempFile(filename, content, {autoDelete = true} = {}) {
    // ...
}
```

It may be that that becomes unwieldy in some instances, but certainly in the simple case it's pretty good.

PR #412, opened 2021-02-19
docs(guidelines): add guidance on options objects

wincent:
I admit this guidance is a bit weakly worded because it says "you should (blah)" but it also lists "an acceptable alternative approach", because we have multiple examples of the latter already in liferay-portal, and there is no real urgency to change them (the pattern is fine).

Closes: https://github.com/liferay/liferay-frontend-projects/issues/409

### DIFF
--- a/guidelines/general/options.md
+++ b/guidelines/general/options.md
@@ -1,0 +1,78 @@
+# Optional function parameters
+
+## How to use an `options` parameter
+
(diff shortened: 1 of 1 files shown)


wincent:
Screenshot 2021-02-22 -091557-Wsimvnln@2x

That's good enough for me 😂

PR #415, opened 2021-02-22
refactor: remove explicit dependency on clayui/css

wincent:
This came in via 18e9c76f12 but I don't think we actually need this as a top-level `devDependency`. It will come in anyway transitively with liferay-frontend-theme-styled and liferay-frontend-theme-unstyled, which were added in f5da76606f.

Originally looked at this because of [this dependabot PR](https://github.com/liferay/liferay-frontend-projects/pull/414).

I don't think this will stop the bot from reminding us to update our own internal dependencies (that we aren't even directly using here anyway), so I might also tell the bot to ignore this dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
 	"description": "Liferay's Frontend Infrastructure Team monorepo",
 	"devDependencies": {
 		"@babel/preset-env": "^7.4.2",
-		"@clayui/css": "^3.x",
 		"@cnakazawa/watch": "^1.0.4",
 		"@types/ejs": "3.0.4",
 		"@types/escodegen": "0.0.6",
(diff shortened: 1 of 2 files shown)


wincent:
Seems trivial, CI is green, so merging.

bryceosterhaus:
I don't think your change should break anything, but in case it does, we do reference `@clayui/css` at https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/packages/npm-scripts/src/presets/standard/index.js#L24

This _shouldn't_ matter, but wanted to reference this specific line just in case it causes any issues when testing in a dev environment.

wincent:
Hm... good catch. I overlooked that one in my grep.

I agree with your assessment and think it'll be ok. Looking in liferay-portal right now, a single/unique copy of `@clayui/css` is getting hoisted up to `modules/node_modules/` (as you'd hope), and, as far as I can tell, that will continue to happen after this change (because when the presence or absence of a package in `devDependencies` of this monorepo can't possibly have any effect on what `yarn` does over in liferay-portal), so the `require` will keep on working (we still have the transitive dep, via the theme, and even if we didn't, other modules depend on `@clayui/css` so we know it'll always be there).

PR #418, opened 2021-02-24
feat(npm-scripts): bump liferay-theme-tasks dependency to 10.3.0

bryceosterhaus:
Now that we have 10.3.0, we need to update that here.

Once this is merged and a new version is released, we should be able to use the `dartSass` option for building themes


wincent (approved):
Ok, so fingers crossed this doesn't break anything. Seeing as, in theory, nothing changes in the theme tasks until we actually flip the switch. What could possibly go wrong? 😱

dont-feel-very-good

PR #420, opened 2021-02-25
feat: externalize liferay-portal's specific configuration for webpack / bundler

izaera:
This PR makes it possible to extract webpack's remotes/shared and bundler's imports to liferay-portal.

This is necessary since:

1. Those values really belong to the liferay-portal project, not npm-scripts, because they depend on the configuration of that project. For instance: they may differ between different portal versions.
2. Having circular dependencies between npm-scripts and liferay-portal has historically led to unstabilities when a new version of npm-scripts needs to be released and integrated in portal.

This PR has been tested locally with [this other PR](https://github.com/liferay-frontend/liferay-portal/pull/839).


izaera:
ci:stop

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
```diff
@@ -153,17 +160,31 @@ function getMergedConfig(type, property) {
...
-			if (Object.keys(userConfig).length === 0) {
-				userConfig = require('../config/npmscripts.config');
+				rootDefaults = {};
 			}
 
+			let userConfig = deepMerge([
+				require('../config/npmscripts.config.js'),
+				rootDefaults,
+				getUserConfig('npmscripts'),
+			]);
```
wincent, line 184: I am not a huge fan of the override model here and it would be good to be really _sure_ that this is the one we want before locking it in. I haven't been able to think about this as deeply as you, so bear with me because there might be some mistakes or inaccuracies in what I am about to say.

I think I get _why_ you did it this way: ie. because there are 23 `npmscripts.config.js` files in liferay-portal (including the root one), and you don't want to do a blindly squish them together with the root because the overrides may produce unpredictable results. So, instead of that, you do this special merge that treats the `defaults` key from the top-level as special, and you carefully merge that in with the user config.

But the counter-argument against this design is that it is atypical (compared to other JS tools that do  config file inheritance, most canonically represented by ESLint's approach) and therefore quite subtle. As in, instead of simple/generic merging strategy, we have a "complex"/specific one. 😢

When I inspect the existing `npmscripts.config.js`, I don't see anything in there that looks unsafe to merge. These two are about as complicated as it gets (I'm just copy-pasting these from `master`, not from your other in-flight PR):

`modules/apps/document-library/document-library-web/npmscripts.config.js`:

```js
const preset = require('@liferay/npm-scripts/src/presets/standard');

module.exports = {
	build: {
		dependencies: [...preset.build.dependencies, 'asset-taglib'],
	},
	preset: '@liferay/npm-scripts/src/presets/standard',
};
```

`modules/apps/asset/asset-categories-admin-web/npmscripts.config.js`:

```js
module.exports = {
	federation: {
		exposes: [
			'<inputDir>/js/ActionsComponentPropsTransformer.js',
			'<inputDir>/js/AssetCategoriesManagementToolbarDefaultEventHandler.es.js',
			'<inputDir>/js/AssetCategoriesSelectorTag.es.js',
		],
		remotes: ['asset-taglib'],
		runLegacyBuild: false,
	},
	preset: '@liferay/npm-scripts/src/presets/standard',
};
```

We would just need to define the semantics to be (_I think_) "more local config overwrites more general config", which is something that `deepMerge` can do (ie. `deepMerge.MODE.OVERWRITE_ARRAYS`, and if that isn't exactly what we need, we can easily enough make another). Your [other PR](https://github.com/liferay-frontend/liferay-portal/blob/ee7d34a8190300fe146d89a5355578fe4eca0d49/modules/npmscripts.config.js) only defines `mode`, `remotes` and `shared` keys under `federation`. As in, a more local config would safely overwrite `mode` and I gather that we just need to decide whether arrays should be overwritten or appended (because `remotes` seems to be the common key that is going to come up). I'm sure you've analyzed this in detail and can know exactly what my clash and why it would be a problem if we didn't merge it right.

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
```diff
@@ -153,17 +160,31 @@ function getMergedConfig(type, property) {
...
 
-			let userConfig = getUserConfig('npmscripts');
+			let rootDefaults;
 
-			// Use default config if no user config exists
+			try {
+				rootDefaults = require(path.join(
+					rootDir,
+					'npmscripts.config.js'
+				)).defaults;
```
wincent, line 171: If `defaults` were ever `undefined` here, what would `deepMerge` do with it below? Gracefully ignore and continue?

wincent (commented):
I would have kept the `npm-bundler-preset-liferay-dev` package around (marked as deprecated until we see all this working), but I admire you boldness. 👏 

I left some comments/questions inline, but overall my major doubt is just what the end goal is here. I see you effectively inline `npm-bundler-preset-liferay-dev` inside the default bundler config that is bundled with `npm-scripts`, but we also have:

- The webpack federation config that you are moving into liferay-portal in [liferay-frontend#839](https://github.com/liferay-frontend/liferay-portal/pull/839); and:
- The sole `npmscripts` preset that continues to [live inside this package](https://github.com/liferay/liferay-frontend-projects/blob/8ea2293220c701c03643e70e1fbac89294282fa7/projects/npm-tools/packages/npm-scripts/src/config/npmscripts.config.js).

It kind of feels like we have three different ways of doing things here...

1. Bundler (v2) configured by an optional JSON file (.npmbundlerrc) in the consuming repo (liferay-portal), with a base set of defaults bundled inside `npm-scripts` (defaults which, as of this PR, don't seem to be compatible with all branches, and I had thought we were going to move them into the consuming repo like you're doing for the webpack federation config).
2. `npm-scripts` itself, configured by optional `npmscripts.config.js` files in the consuming repo (+ after this PR, a set of overrides that are only of interest to webpack defined inside another `npmscripts.config.js` at the `modules/` root; see below) + a base [set of defaults in this repo](https://github.com/liferay/liferay-frontend-projects/blob/6f948439e6a103a4845fc56ae832c2ca037ae432/projects/npm-tools/packages/npm-scripts/src/config/npmscripts.config.js), which itself is largely based on an internally bundled "preset".
3. webpack ("Bundler v4") configured by piggy-backing inside one of the files mentioned in "2" above, according to special merge rules.

...even though we control all the tooling, and if we are going to take the opportunity to do this cleanup, it would be good to get as close to an "optimal", uniform way of doing it as part of this work. It feels like "2" might be too complicated in itself, and if you zoom out and look at the thing as a whole it _definitely_ feels too complicated. I'm worried that any time we have to come back to and touch this code or fix bugs after an absence it is going to seem very hard to maintain.

So all that said I guess my two questions are: what is your desired end state here? and what opportunities to you see for simplification and consolidation as part of all this?

projects/npm-tools/packages/npm-scripts/src/presets/standard/index.js
```diff
@@ -22,15 +22,13 @@ const CHECK_AND_FIX_GLOBS = [
 const getClayPaths = () => {
 	try {
 		return require('@clayui/css').includePaths;
-	}
-	catch {
+	} catch {
```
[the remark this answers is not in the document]
izaera: Yeah, it's not SFed yet.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -4,6 +4,7 @@
  */
 
 const fs = require('fs');
+const {run} = require('jest');
```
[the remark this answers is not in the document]
izaera: :open_mouth:  That was the `CoC` plugin. I didn't know it was so smart.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -81,7 +81,7 @@ module.exports = async function (...args) {
 	}
 
 	const runLegacyBuild =
-		!FEDERATION_CONFIG || FEDERATION_CONFIG.runLegacyBuild !== false;
+		!federation || ['compatible', 'disabled'].includes(federation.mode);
```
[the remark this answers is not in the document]
izaera: Good naming :sweat_smile: . I wrote `incompatible` because the other option is `compatible`, but being the default we may as well name it `default` :+1: .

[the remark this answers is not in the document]
izaera: And `compatible` is because it is compatible with the previous architecture (i.e.: bundler + AMD).

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
izaera: Well, the reason was simpler than that...

We have definitions in the root `npmscripts.config.js` to do specific "global" tasks:

```javascript
const CHECK_AND_FIX_GLOBS = [
     '!settings.json',
     '!tsconfig.json',
     '/{,dxp/}*.{js,ts}',
     '/{,dxp/}apps/*/*/*.{js,json,ts,tsx}',
     '/{,dxp/}apps/*/*/*/*.{js,json,ts,tsx}',
     '/{,dxp/}apps/*/*/*/{src,test}/**/*.{js,scss,ts,tsx}',
     '/{,dxp/}apps/*/*/*/{src}/**/*.{jsp,jspf}',
     '/{,dxp/}apps/*/*/{src,test}/**/*.{js,scss,ts,tsx}',
     '/{,dxp/}apps/*/*/{src}/**/*.{jsp,jspf}',
];
```

In particular, these globs :point_up: are broader than the ones used in the default preset. Also, because we were already using that file, I didn't want to change that. 

But as you say, maybe this is the right time to do that change. Now that the file hasn't grown too big and we can still do it. I wasn't a big fan of the `defaults` field either for the same reasons you give.

The only caveat is, what would be the behavior of the root level (`liferay-portal/modules`) npm scripts... 

Currently we have:

```json
     "scripts": {
         "checkFormat": "liferay-npm-scripts check",
         "format": "liferay-npm-scripts fix"
     }
```

If we put defaults in the sibling `npmscripts.config.js` file it looks a bit incorrect to put these scripts here, because, being at the root level, they are intended for global purposes and, in that case, they may need some way to override defaults (as is the case here). So, a bit of conflict here, I would say...

izaera: Regarding the merging strategy I've come across that in my local WIP and it seems we won't be able to use `deepMerge.MODE.OVERWRITE_ARRAYS` because of the `remotes` and `shared` arrays, unless we want to replicate them entirely each time we want to add a dependency. You could argue that you may as well need to remove a dependency and then you cannot use an extend-only strategy, but for now that doesn't happen and I think we can live with that, because if you need to remove something from the defaults, why did you put it in the defaults in first place? :sweat_smile:

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
izaera: Probably fail .

me coding

izaera:
>  what is your desired end state here? 

The reason everything is not yet done is because I wanted to do this change with intermediate steps that can be tested (by running ant all and checking that the build and execution is stable). The steps are:

1. Externalize webpack federation `remotes` and `shared` to `liferay-portal` (two first commits)
2. Remove npm-bundler preset by internalizing its contents into the default `npm-bundler.json` config file (next two commits).
3. Remove npm-scripts presets feature by moving the npm-scripts preset contents to the default `npmscripts.config.js` file in `npm-scripts`). I'll push this today.
4. Externalize sensible defaults from the `npmscripts.config.js` and `npm-bundler.json` files to `liferay-portal`.

Step 4 is yet to be decided because I had thought of two options:

a) Create a base `.npmbundlerrc` file at `portal/modules` with defaults.
b) Put those defaults inside the root `npmscripts.config.js` file 

I'm more inclined to do `b` because:

1) There's no concept of base `.npmbundlerrc` in the bundler so, putting one base file in `portal/modules` may give a false impression to people not knowing the tools in depth.
2) Given that we already have the `npm-bundler.json` file inside `npm-scripts`, it is logical to externalize its content to a file that configures `npm-scripts`, thus `npmscripts.config.js`.
3) Given that in master we will eventually end up not having bundler configuration at all, it makes sense not to create another new config file to delete it afterwards.

That said, I haven't digged into 4 yet, so I may encounter some problems/nuances that make me reconsider something/everything.

> and what opportunities to you see for simplification and consolidation as part of all this?

Right now I'm only paying attention to externalize imports/federation configuration which is the most obvious part that needs to be pulled off npm-scripts. However, once we finish that, we need to carefully inspect, for every defaults file, what is part of npm-scripts itself, and what is specific to the portal build.

For example, we may define that npm-scripts supports ES6+ syntax. In that case, we need to put babel configuration for that to happen inside npm-scripts defaults not in portal, since it is a feature provided by npm-scripts and it would be provided for people using it ouside of portal too.

But then, what happens with the output directories, for example? I would say they are something that belong more to portal than to npm-scripts (though it's clearly a definition problem, not something objective).

And what about RTL, scss, linting, etc...? Luckily we don't need to tackle all this now, but we will need to do this exercise in the mid term if we want npm-scripts to be used outside of portal. And, of course, it's not only a technical exercise, but it depends on the product definition too so we may defer it for now.

What I'm trying to say is that we may not have the complete image now to define what the final state of this would be...

izaera:
Commit https://github.com/liferay/liferay-frontend-projects/pull/420/commits/da642bf7896b763eaeecb0799ab16e9352aec716 corresponds to commit https://github.com/liferay-frontend/liferay-portal/pull/839/commits/f19e94f1b578153dcee23ec2253d65a840108f67 in the other PR.

These two implement step 3 of comment https://github.com/liferay/liferay-frontend-projects/pull/420#issuecomment-786011535 : 

> Remove npm-scripts presets feature by moving the npm-scripts preset contents to the default npmscripts.config.js file in npm-scripts). I'll push this today.

projects/npm-tools/packages/npm-scripts/src/config/npmscripts.config.js
```diff
@@ -3,8 +3,143 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-var path = require('path');
+const path = require('path');
+
+const CHECK_AND_FIX_GLOBS = [
```
izaera, line 8: These are the globs I was referrring to in [this comment](https://github.com/liferay/liferay-frontend-projects/pull/420#discussion_r582933129)

projects/npm-tools/packages/npm-scripts/src/config/npmscripts.config.js
```diff
@@ -3,8 +3,143 @@
...
+	}
+};
 
 module.exports = {
-	preset: path.join(__dirname, '../presets/standard/index'),
+	build: {
+		// Passed to:
+		// - `metalsoy` executable (via `generateSoyDependencies()`).
+
+		dependencies: [
```
izaera, line 31: This will be externalized to the portal in further commits.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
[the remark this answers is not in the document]
wincent: > but being the default we may as well name it `default` 👍 .

Sounds a bit easier to sell than `incompatible`. 😁

wincent:
> * Externalize webpack federation `remotes` and `shared` to `liferay-portal` (two first commits)
> * Remove npm-bundler preset by internalizing its contents into the default `npm-bundler.json` config file (next two commits).
> * Remove npm-scripts presets feature by moving the npm-scripts preset contents to the default `npmscripts.config.js` file in `npm-scripts`). I'll push this today.
> * Externalize sensible defaults from the `npmscripts.config.js` and `npm-bundler.json` files to `liferay-portal`.

good-good

Well, that's good to hear. And sounds about right to me.

> And what about RTL, scss, linting, etc...?

It's clear that we want to leverage existing standard config files whenever we can (eg. `.eslintrc.js`, `.babel.config.js` etc), so we can definitely cross "linting" off that list. We've long used `webpack.config.js` in those places where we use webpack explicitly (I think the federation stuff doesn't count because we've decided to hide webpack as an implementation detail). So for "everything else", yeah, we need to find somewhere to shove it, and given that "Bundler" is kind of disappearing at this point, it does seem that `npmscripts.config.js` is the natural home for a lot of this stuff because `npm-scripts` is our "choke point" tool through which we try to funnel all of our build-time/test-time activity, and as a vehicle for controlling/abstracting dev dependencies.

Let's talk further about the details tomorrow. I'm too tired now to say sensible things.

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
izaera: A good way to fix the global lint configuration "problem" is to provide an CLI argument for npm-scripts to specify a different config file (like some other tools do). That way we may use the npmscripts.config.js file in portal/modules as base config and provide another file for the global tasks (if needed).

wincent: So we could have `modules/npmscripts.config.js` and `modules/npmscripts.config2.js`?

(Just kidding about the names). One of the reasons you wanted to put this stuff into `npmscripts.config.js` in the first place was to avoid the need to add another configuration file... So I think you need to make a stronger argument for how this would work and why it would be a good idea (ie. details of what the file would be called, where it would live, how it would be clearly separated from the other `npmscripts.config.js` at a conceptual level etc.)

izaera: The other alternative is to refactor the whole `check` and `fix` for the workspace folder so that they simply scan subfolders and invoke the task per project (with different configurations for each project). However, we can achieve the same result with gradle (in the portal).

Shall we do the refactor? That would be the best option IMO, because it is the most standard behavior. However it is out of the scope of this PR, I think. Maybe we want to defer it. I don't know...

izaera: I have finally added a `global` section to invert the relation between defaults and configuration used to check and lint the whole workspace.

wincent:
Still needs source formatting (maybe didn't include changes in your last commit?):

```
projects/npm-tools/packages/npm-scripts/src/config/npmscripts.config.js
projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
```

And don't forget to retitle the PR to satisfy the Semantic PR bot. I'm going to look at the code now.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -80,8 +86,7 @@ module.exports = async function (...args) {
 		buildSoy();
 	}
 
-	const runLegacyBuild =
-		!FEDERATION_CONFIG || FEDERATION_CONFIG.runLegacyBuild !== false;
+	const runLegacyBuild = !federation || federation.mode != 'default';
```
wincent, line 89: `!==`

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
```diff
@@ -123,12 +147,20 @@ function getMergedConfig(type, property) {
...
 
-		case 'bundler':
-			mergedConfig = deepMerge([
-				require('../config/npm-bundler'),
-				getUserConfig('npmbundler'),
-			]);
+		case 'bundler': {
+			const {bundler: bundlerDefaults} = getMergedConfig(
+				'npmscripts',
+				'build'
```
wincent, line 153: Just checking: so this will `throw` if there is no `build` key; and that's ok to enforce because we provide one by default?

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -64,6 +58,18 @@ if (!BUILD_CONFIG) {
  * `minify()` is run unless `NODE_ENV` is `development`.
  */
 module.exports = async function (...args) {
+	const config = getMergedConfig('npmscripts');
+
+	createTempFile('npmscripts.config.json', JSON.stringify(config, null, 2), {
+		autoDelete: false,
```
wincent, line 64: I gather the plan is to remove this once we're confident it works, right? In the past we worked pretty hard to keep untracked/dirty files out of the worktree.

wincent:
Left some comments in-line but it all seems pretty reasonable to me.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: This is not generated in the source tree, but in the temporary directory, which is configurable and (in our case) is placed inside the build folder. Remember we had to move everything out of the source tree because of the `watch` feature.

The reason we need these files is that, for any diagnostic, it is quite useful to know what the tools are really seeing. Think of it as the report file in the bundler, which helped a lot with that. Also, for learning purposes, some users may find these files useful.

Otherwise, with all the complex inheritance we have in place, it's nearly impossible to know what is being fed to webpack, the federation plugin, and so on, without spending a good amount of time looking at different places.

That said, I'm sure we will end up doing this some other way, once we settle all this stuff... Perhaps a report file or some kind of advanced diagnostic tool that can be used by developers to know what's going on in their builds. Maybe a log level parameter could work too...

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
izaera: I think it is impossible that you arrive to this point without having a `build` config. Otherwise the bundler wouldn't be invoked.

However, I'll rewrite it just in case, because it may be assuming too much...

izaera:
The semantic PR bot still complains. What's wrong? :thinking:

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
wincent: > This is not generated in the source tree, but in the temporary directory, which is configurable and (in our case) is placed inside the build folder.

Bah, and you probably told me this the last time I asked about it too. 🤦‍♂️

sorryz

wincent:
> The semantic PR bot still complains. What's wrong? 🤔

You have `feat!(npm-scripts):` but it should be `feat(npm-scripts)!:`

izaera:
Oops, that has changed, then...

I did a grep inside git history; see this commit, which is green :point_right: https://github.com/liferay/liferay-frontend-projects/commit/62e1efc300302895bc727e02edb3d27c949722b4 )

izaera:
Apparently it was that, yes. :shrug:

wincent:
> It could be that the linked commit is old and the bot wasn't in place.

I reckon it was, so no idea why it didn't complain. It's not in the list of checks.

PR #425, opened 2021-03-01
feat(npm-scripts): add some basic support for building with TypeScript

wincent:
So I need to add some code to liferay-portal in service of [LPS-127081](https://issues.liferay.com/browse/LPS-127081) ("Implement and integrate Global JS Store mechanism in DXP") and I think it would be very useful to "future proof" it by making sure it is TypeScript-ready. We probably don't want to have our store just be un untyped "bag of data", especially if we are going to have to live with and support this API for many years and it is going to become a central channel for cross-app communication with DXP. Example libraries which share this point of view are [Undux](https://github.com/bcherny/undux) and [Recoil](https://recoiljs.org), but they are both React-specific and we want something generic.

So, we can either:

1. Build and ship the API in TypeScript in a way that lends itself to strongly typing the data; or:
2. At least _prototype_ the API in TypeScript to show that it supports strongly-typed usage, even if we end up stripping the type annotations and shipping the thing in JS at the end (ie. leave the door open to an eventual conversion to TypeScript).

There are some obstacles to doing either though, because "turning on TypeScript" in DXP is not going to be as simple as flipping a switch. The only module where we're using TypeScript currently is doing it in a sneaky way, [via webpack and the ts-loader](https://github.com/liferay/liferay-portal/blob/9837c4d81f2de200ad2bcd7acc83578fb25e975c/modules/apps/remote-app/remote-app-client-js/webpack.config.js#L28).
    
To make it possible to _actually_ use TypeScript in a well-integrated fashion in DXP we're going to have to do a bit more work (that means making sure all the linting, formatting, testing pathways works, that our build process doesn't regress in terms of performance, that cross-module type definitions are easily consumed and available, and so on...).

This PR is a draft showing what the minimum first steps would be to get us going and be able to do one of "1" or "2" mentioned above.

There are three main ways to process TypeScript at this time:

1. `webpack` + `ts-loader`: this handles both building and type-checking.
2. Babel's `@babel/preset-typescript`, which handles _only_ stripping out of TypeScript syntax (eg. annotations); it does not do any actual type-checking, so is of limited value.
3. Using the TypeScript compiler itself (`tsc`), which can both type-check and emit code.

So given that "2" on its own is not valuable (no type-checking) and going all-in on "3" is a huge/risky change (ie. from Babel to `tsc`, especially fraught as we're in the middle of moving towards a "federated" build style that is much more webpack-centric than Babel-centric), I'm going to go with a mixture: we'll run `tsc --noEmit` to do type-checking only, and then Babel with the preset to do the build while stripping out TS type annotations.

Further details explained in the commit messages, but the executive summary is that we need to:

1. Add some type definitions to duct tape over an old transitive dependency that we have and allow us to run `tsc --noEmit` during the build.
2. Actually run `tsc --noEmit` during the build.
3. Configure Babel (awkwardly) to not ignore ".ts" files.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/package.json
+++ b/projects/npm-tools/packages/npm-scripts/package.json
@@ -23,6 +23,7 @@
 		"@testing-library/react-hooks": "^3.4.2",
 		"@testing-library/user-event": "^4.2.4",
 		"@types/jest": "^26.0.14",
+		"@types/vfile-message": "2.0.0",
 		"babel-eslint": "^10.0.3",
 		"babel-jest": "^25.1.0",
 		"babel-loader": "^8.1.0",
--- a/projects/npm-tools/packages/npm-scripts/src/config/babel.json
+++ b/projects/npm-tools/packages/npm-scripts/src/config/babel.json
@@ -11,7 +11,8 @@
 				"targets": "Chrome 65, Firefox ESR, Safari >= 12, last 1 Chrome version, last 1 ChromeAndroid version, last 1 Edge version, last 1 Firefox version, last 1 iOS version"
 			}
 		],
-		"@babel/preset-react"
+		"@babel/preset-react",
+		"@babel/preset-typescript"
 	],
 	"plugins": [
 		"@babel/proposal-class-properties",
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
@@ -17,6 +17,7 @@ let runBridge = require('../utils/runBridge');
 let runBundler = require('../utils/runBundler');
 const setEnv = require('../utils/setEnv');
 let {buildSoy, cleanSoy, soyExists, translateSoy} = require('../utils/soy');
+const spawnSync = require('../utils/spawnSync');
 const validateConfig = require('../utils/validateConfig');
 let webpack = require('./webpack');
 
@@ -89,6 +90,10 @@ module.exports = async function (...args) {
 	const runLegacyBuild = !federation || federation.mode !== 'default';
 
 	if (inputPathExists && runLegacyBuild) {
+		if (fs.existsSync('tsconfig.json')) {
+			spawnSync('tsc', ['--noEmit']);
+		}
+
 		runBabel(
 			BUILD_CONFIG.input,
 			'--out-dir',
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
@@ -93,7 +93,9 @@ module.exports = async function (...args) {
 			BUILD_CONFIG.input,
 			'--out-dir',
 			BUILD_CONFIG.output,
-			'--source-maps'
+			'--source-maps',
+			'--extensions',
+			'.cjs,.es,.es6,.js,.jsx,.mjs,.ts,.tsx'
 		);
 	}
 
--- a/projects/npm-tools/packages/npm-scripts/test/utils/getMergedConfig.js
+++ b/projects/npm-tools/packages/npm-scripts/test/utils/getMergedConfig.js
@@ -91,6 +91,7 @@ describe('getMergedConfig()', () => {
 							targets: expect.stringContaining('Chrome version'),
 						},
 					],
+					'@babel/preset-typescript',
 				]);
 
 				expect(config.overrides[0].presets).toMatchObject([
--- a/yarn.lock
+++ b/yarn.lock
@@ -1885,6 +1885,13 @@
   resolved "https://registry.yarnpkg.com/@types/unist/-/unist-2.0.3.tgz#9c088679876f374eb5983f150d4787aa6fb32d7e"
   integrity sha512-FvUupuM3rlRsRtCN+fDudtmytGO6iHJuuRKS1Ss0pG5z8oX0diNEw94UEL7hgDbpN94rgaK5R7sWm6RrSkZuAQ==
 
+"@types/vfile-message@2.0.0":
+  version "2.0.0"
+  resolved "https://registry.yarnpkg.com/@types/vfile-message/-/vfile-message-2.0.0.tgz#690e46af0fdfc1f9faae00cd049cc888957927d5"
+  integrity sha512-GpTIuDpb9u4zIO165fUy9+fXcULdD8HFRNli04GehoMVbeNq7D6OBnqSmg3lxZnC+UvgUhEWKxdKiwYUkGltIw==
+  dependencies:
+    vfile-message "*"
+
 "@types/vinyl@*":
   version "2.0.4"
   resolved "https://registry.yarnpkg.com/@types/vinyl/-/vinyl-2.0.4.tgz#9a7a8071c8d14d3a95d41ebe7135babe4ad5995a"
@@ -15428,7 +15435,7 @@ vfile-location@^3.0.0:
   resolved "https://registry.yarnpkg.com/vfile-location/-/vfile-location-3.1.0.tgz#81cd8a04b0ac935185f4fce16f270503fc2f692f"
   integrity sha512-FCZ4AN9xMcjFIG1oGmZKo61PjwJHRVA+0/tPUP2ul4uIwjGGndIxavEMRpWn5p4xwm/ZsdXp9YNygf1ZyE4x8g==
 
-vfile-message@^2.0.0:
+vfile-message@*, vfile-message@^2.0.0:
   version "2.0.4"
   resolved "https://registry.yarnpkg.com/vfile-message/-/vfile-message-2.0.4.tgz#5b43b88171d409eae58477d13f23dd41d52c371a"
   integrity sha512-DjssxRGkMvifUOJre00juHoP9DPWuzjxKuMDrhNbk2TdaYYBNMStsNhEOt3idrtI12VQYM/1+iM0KOzXi4pxwQ==


wincent:
> For whatever reason, when the error is present, Yarn is installing a nested copy of vfile-message (v2.0.2) inside `@types/vfile-message` because elsewhere via a lengthy dependency chain we're getting v1.1.1 (via npm-scripts -> stylelint -> postcss-markdown -> remark -> unified -> vfile, which asks for "^1.0.0"):
>
> Adding the dependency explicitly in DXP itself, the nested copy goes away, and along with it, the error. We'll see if adding it here does the trick too.

Confirming that it does indeed go away. Proving once again — _DUCT TAPE: It Just Works™_.

jbalsas:
Well... this is an unexpected side effect :)

I wonder, why not go with `1` since we're going to be building a new module and can leverage directly the new building system?

wincent:
There's not actually that much difference between "1" and "2".

"1" means that we do "2" and then chicken out and dial it back, and "2" just means that we do "2" and stick with it.

bryceosterhaus:
> I'm going to go with a mixture: we'll run tsc --noEmit to do type-checking only, and then Babel with the preset to do the build while stripping out TS type annotations.

This seems solid to me. Pretty sure we actually do something similar in Clay by using babel to build the packages and then tsc --noEmit to generate the types. It's worked pretty well for Clay and haven't had too many issues.

wincent:
Yeah, I tested it with an `ant all` and nothing breaks. The main concern that I have in the long term is cross-module type-checking. The way you would normally do this in a large project would probably be with [project references](https://www.typescriptlang.org/docs/handbook/project-references.html), but the problem we have in liferay-portal is that each OSGi module is its own separate little world that gets built with a call from Gradle to `liferay-npm-scripts build` in an order determined by Gradle, based on _Gradle's_ knowledge of the dependency graph.

If we could run TypeScript once at the top, then the build would be slow, but it would always be correct because the TypeScript compiler would "see" every TS/JS module in the repo in an up-to-date state as per what's on the disk. But I doubt we'll can have a "from the top" build like that (too much of an impedance mismatch with the Gradle/OSGi model), and that forces us in the direction of totally independent builds (where the type checker would only catch problems _internal_ to a given module) or project references (where we effectively have an entire dependency graph outside of, and invisible to, Gradle).

As [the docs](https://www.typescriptlang.org/docs/handbook/project-references.html#caveats-for-project-references) say:

> Because dependent projects make use of `.d.ts` files that are built from their dependencies, you’ll either have to check in certain build outputs _or_ build a project after cloning it before you can navigate the project in an editor without seeing spurious errors. We’re working on a behind-the-scenes `.d.ts` generation process that should be able to mitigate this, but for now we recommend informing developers that they should build after cloning.

So it might get easier in the future, but at least for now it is a thorny problem due the fact that we're living inside an Ant/Gradle based build system. When TS is only used in 1 or 2 modules, it doesn't really matter, but by the time we're in 10, 20, 50, 100 modules, we're going to have to tackle the problem in a definitive way or we'll be leaving a fair chunk of the value of the type checker on the table.

jbalsas:
> There's not actually that much difference between "1" and "2".
>
> "1" means that we do "2" and then chicken out and dial it back, and "2" just means that we do "2" and stick with it.

Sorry, probably too many 1s and 2s in there for my comment to make sense. Just to clarify, I was referring to:

>1. webpack + ts-loader: this handles both building and type-checking.
>2. Babel's @babel/preset-typescript, which handles only stripping out of TypeScript syntax (eg. annotations); it does not do any actual type-checking, so is of limited value.
>3. Using the TypeScript compiler itself (tsc), which can both type-check and emit code.

I think I meant why not do the webpack+ts-loader already instead of an intermediate mixture.

jbalsas:
Nevertheless, looks shippable to me!

batmanrobin

wincent:
> I think I meant why not do the webpack+ts-loader already instead of an intermediate mixture.

This was one of the things I was considering (in a private thread with Iván yesterday) but didn't want to touch anything webpack related while those changes are still happening. Some time between now (when a handful of modules are being built with the new "bundler-not-a-bundler-really-just-webpack-with-a-wrapper") and the future (when all modules are being built with webpack and no modules are using Babel) we'll have to flip a switch. My main concern about the eventual shape of the build tooling, no matter what form it takes, is the cross-module stuff I was mentioning in [my previous comment](https://github.com/liferay/liferay-frontend-projects/pull/425#issuecomment-788709653).

jbalsas:
> This was one of the things I was considering (in a private thread with Iván yesterday) but didn't want to touch anything webpack related while those changes are still happening. Some time between now (when a handful of modules are being built with the new "bundler-not-a-bundler-really-just-webpack-with-a-wrapper") and the future (when all modules are being built with webpack and no modules are using Babel) we'll have to flip a switch. My main concern about the eventual shape of the build tooling, no matter what form it takes, is the cross-module stuff I was mentioning in my previous comment.

Sounds good

shalliproceed

izaera:
LGTM. 

I see you are only integrating it for the legacy build, but that's OK for now. I suppose that, in the future we will want to move the tsc configuration to npmscripts.config.js itself and things like that, but until we get to that point I wouldn't spend time thinking about it.

wincent:
> I see you are only integrating it for the legacy build, but that's OK for now.

Yep, the idea was to keep it minimal for now just to see it working (with the idea that, as we move off the legacy build, this will probably become `ts-loader`-based anyway)...

wincent:
Rebased this and added [one more commit](https://github.com/liferay/liferay-frontend-projects/pull/425/commits/6fd8c0f460bb505a5a03e84ed3c7a54659344045):

> feat(npm-scripts): activate TypeScript preset by default
>
> While rebasing [my WIP liferay-portal PR](https://github.com/wincent/liferay-portal/pull/373) I had to resolve conflicts because we've had a concurrent update to `@liferay/npm-scripts` in the meantime. In order to continue developing that PR, I am going to need a release of the scripts.
> 
> Given that [this draft PR](https://github.com/liferay/liferay-frontend-projects/pull/425) got some "lg2m" comments, I may as well merge it so that it will wind up in the next scripts release, and I won't need to keep releasing new previews just to keep up.
>
> This commit is the one change that I think will be needed in order to make this changes "safe" for liferay-portal. Without this preset, the build will choke in layout-content-page-editor-web because it contains ".d.ts" files that Babel would normally ignore, but will now try to parse; activating the preset prevents the build from choking on the syntax. The preset is pretty benign; all it does is teach Babel to grok TypeScript syntax and strip it out during the build (no type-checking actually occurs during the Babel run).
>
> Test plan: I will cut one more test release after this PR is pushed, and
if an `ant all` works with that, then I'd say we'll be safe to merge it.

wincent:
The only thing that might make a release with this PR unsafe to integrate in liferay-portal might be the necessary changes that I made to `remote-app-client-js` in [the WIP](https://github.com/wincent/liferay-portal/pull/373/files#diff-c391c9855a37bdddca4f06f0b090941b969f4dde7f49f2cdf976022ac8ff6ea7). Not sure if that link will survive the next rebase/push, so this is the diff I am referring to:

```diff
diff --git a/modules/apps/remote-app/remote-app-client-js/tsconfig.json b/modules/apps/remote-app/remote-app-client-js/tsconfig.json
index bb4af4c059d6..2bb5acb44d26 100644
--- a/modules/apps/remote-app/remote-app-client-js/tsconfig.json
+++ b/modules/apps/remote-app/remote-app-client-js/tsconfig.json
@@ -1,6 +1,7 @@
 {
 	"compilerOptions": {
 		"module": "es6",
+		"moduleResolution": "node",
 		"outDir": "./build/node/packageRunBuild/dist/",
 		"sourceMap": true,
 		"strict": true,
@@ -10,5 +11,6 @@
 			"./node_modules/@types",
 			"./src/main/resources/META-INF/resources/js/types"
 		]
-	}
+	},
+	"include": ["src/**/*", "tests/**/*"]
 }
\ No newline at end of file
```

I'm going to confirm that doesn't break liferay-portal if sent independently and send a PR for that if I can.

jbalsas:
> Without this preset, the build will choke in layout-content-page-editor-web because it contains ".d.ts" files that Babel would normally ignore, but will now try to parse; activating the preset prevents the build from choking on the syntax.

Huh... that's interesting! I think I remember we discussed this and ended telling them to create the `.d.ts` files for the future, but maybe that was not the right call?

How would this look in the end? Shouldn't these be auto-generated once the project is migrated to TS?

wincent:
> I think I remember we discussed this and ended telling them to create the .d.ts files for the future, but maybe that was not the right call?

I don't remember. I don't think I would have told someone to do this, but I can imagine (not remember) that if somebody came to me with a question like, "Given that we want to go to TypeScript, should I stop writing `propTypes` definitions and/or start publishing a `.d.ts` file so that my IDE can give me good autocomplete?" I would have said, "Sure, why not?".

> How would this look in the end? Shouldn't these be auto-generated once the project is migrated to TS?

Yeah, once a project is TS, you can get TS itself to emit type definition files. The main use case is for integration between "projects that are not in TS" with "projects that are in TS".

jbalsas:
> I don't remember. I don't think I would have told someone to do this, but I can imagine (not remember) that if somebody came to me with a question like, "Given that we want to go to TypeScript, should I stop writing propTypes definitions and/or start publishing a .d.ts file so that my IDE can give me good autocomplete?" I would have said, "Sure, why not?".

I think it was an internal-short conversation in the context of [Inline content editing mode](https://issues.liferay.com/browse/LPS-106718) that produced https://github.com/sandrodw3/liferay-portal/pull/187

wincent:
> I think it was an internal-short conversation

count-on-you

wincent:
> I'm going to confirm that doesn't break liferay-portal if sent independently and send a PR for that if I can.

Confirmed. PR [here](https://github.com/liferay-frontend/liferay-portal/pull/871).

wincent:
> I will cut one more test release after this PR is pushed, and if an ant all works with that, then I'd say we'll be safe to merge it.

`ant all` works and test driving product works, so I think this is safe to merge given previous endorsements:

> [This seems solid to me.](https://github.com/liferay/liferay-frontend-projects/pull/425#issuecomment-788203940)

big-clap

> [Nevertheless, looks shippable to me!](https://github.com/liferay/liferay-frontend-projects/pull/425#issuecomment-788724926)

slow-clap-yeah

> [LGTM.](https://github.com/liferay/liferay-frontend-projects/pull/425#issuecomment-788835020)

lo-clap

PR #426, opened 2021-03-04
feat: add a resolver for soy templates to webpack federation

izaera:
This is a partial resend of https://github.com/liferay/liferay-frontend-projects/pull/396 leaving out the `exposes` linter.

This implements the webpack resolver for soy modules (https://issues.liferay.com/browse/IFI-2326), which is necessary to continue migrating modules to webpack in liferay-portal .

(see https://github.com/liferay/liferay-frontend-projects/pull/426#issuecomment-790742329 below for an explanation of the strikethrough :point_down: )
~I won't be releasing any npm-scripts version after merge until it is needed, but I've tested this PR with [Brian's master branch from today](https://github.com/brianchandotcom/liferay-portal/commit/c724841cf6f0b0baff32e5aa3fca17b799583427) locally and everything seems fine so I consider it safe to merge.~ 

~The code was already reviewed in the [former PR](https://github.com/liferay/liferay-frontend-projects/pull/396) but if anyone wants to give it another look since it has passed so much time, it's OK.~


izaera:
I'm going to add one more commit because I will need it to release a new version and integrate it in portal.

izaera:
The commit holds the explanation and the rationale for the new change, which is needed by [LPS-128669](https://issues.liferay.com/browse/LPS-128669).

wincent (approved):
> Once we remove the bundler from npm-scripts we may get rid of this option.

No biggy... but for temporary stuff that we know we want to delete we could be giving them names that start with "do not use or you will be fired" (or `unstable`, `experimental`, `__` etc).

Whatever. I looked at this again and it looks to be in the right language.

izaera:
> for temporary stuff that we know we want to delete

I knew I wanted to delete `mode` and it has been promoted to permanent :joy: 

Once we finish the migration and, if we remove `npm-bundler` from `npm-scripts` I will clean up all those.

However, I'm not 100% sure we will remove the bundler, because maybe we want to provide backwards compatibility (note this is a PM decision). So, in that case, instead of removing them, I will have to refactor the thing a bit to put things in coherent places.

Right now it's a bit of a mess :cry: . For example, we put `federation` in the top level of npmscripts.config.js, but the bundler configuration is inside build. And there are more incoherences I think. So, instead of stopping the epic and fixing them, I preffer to continue and do a final round to make it logic (once we know what is temporary, final, etc) ... :thinking:

jbalsas:
> No biggy... but for temporary stuff that we know we want to delete we could be giving them names that start with "do not use or you will be fired" (or unstable, experimental, __ etc).

I've always wanted to name something `incompatible`... makes me really trust the thing 😝

izaera:
> I've always wanted to name something incompatible

So much noise... :roll_eyes: 

noise

Issue #405, opened 2021-02-15
Turn off no-for-of-loops rule by default


wincent:
As [noted here](https://github.com/liferay-frontend/liferay-portal/pull/789#discussion_r5762704010), this rule is still on by default and it shouldn't be. We can either turn the rule off in `master`, or turn it off by default and then opt-in on the older branches. I like the latter approach better, even though it is more work, because it sets us up for the long term.

bryceosterhaus:
> turn it on by default and then opt-out on the older branches

^ I like this option

wincent:
Edited for clarity because I am a 🤡 and had the options inverted. Originally had:

> We can either turn the rule off in master, or turn it **on** by default and then **opt-out** on the older branches.

But it's supposed to be:

> We can either turn the rule off in master, or turn it **off** by default and then **opt-in** on the older branches.

bryceosterhaus:
I must've misread because I agree'd with your edited version. So it all works out

wincent:
Closed this issue, but linking to the upstream PRs that turn the rule back on where it's "needed":

- https://github.com/brianchandotcom/liferay-portal-ee/pull/33540
- https://github.com/brianchandotcom/liferay-portal-ee/pull/33541
- https://github.com/brianchandotcom/liferay-portal-ee/pull/33542
- https://github.com/liferay/alloy-editor/pull/1469

Issue #354, opened 2019-07-02
Standardize how we iterate over objects (`for`/`of`)


markocikos:
tl;dr: See below for discussion, but [plan](https://github.com/liferay/liferay-frontend-projects/issues/354#issuecomment-762117930) is:

- Turn _on_ `no-for-of-loops/no-for-of-loops` rule on older branches (7.3.x and prior) which must support IE, via explicit setting in `.eslintrc.js`.
- Turn the rule _off_ in the default config provided by `@liferay/eslint-config` (note that we'll probably continue to _bundle_ the `no-for-of-loops` plug-in with the config in order to not break existing set-ups that may be relying on it transitively).
- Update `@liferay/eslint-config` everywhere, freeing the people and bringing peace to the galaxy (we don't need to do an immediate migration; we can write a guideline for this and let usage drift towards it organically — if expected progress doesn't materialize, we can make a lint).

---

I've seen all below used in portal, and there are probably others.

1. 
```
for (const key in obj) {
    if (hasOwnProperty(key)) {
        // do stuff
    }
}
```
2. 
```
for (const key in obj) {
    // do stuff, we don't care about fringe inheritance case
}
```
3.
```
Object.keys(obj).forEach(key => {
    // do stuff
});
```
4.
```
for (const key of Object.keys(obj)) {
    // do stuff
}
```
5.
```
for (const [key, value] of Object.entries(obj)) {
    // do stuff
}
```
6.
```
for (let i = 0, l = Object.keys(obj).length; i < l; i++) {
    // do stuff
}
```

Usage of `for..on` is currently flagged as warning in eslint, although (5) is my favorite iteration  method. What should be the recommended method?

victorg1991:
I guess that now that we don't support ie11, we can remove the lint rule forbidding for..of loops, right?

wincent:
Yes, @victorg1991, but we have to do it in a way that doesn't break things when the change gets backported to older branches (which still support IE). There are a few options for that, but the best one is probably turning the rule _on_ explicitly in an `.eslintrc.js` override on each of the older branches and _then_ turning the rule off in the default config provided by `@liferay/eslint-config`.

victorg1991:
That sounds reasonable @wincent :D. Should I open a ticket somewhere or is this being tracked somewhere?

wincent:
I think _this_ is the ticket; I'll update the title and description to reflect our "conclusion".

kresimir-coko:
[...]

PR #429, opened 2021-03-08
feat(eslint-config)!: turn off no-for-of-loops rule by default

wincent:
Now that we no longer support IE on the DXP master branch, we can turn this rule off by default, and then opt explicitly back into it on the maintenance branches (eg. 7.3.x and older), or in products which go to the maintenance branches (eg. alloy-editor etc).

The truth is that even in places where we need to support IE, this rule was always very conservative; we polyfill/transpile in IE, and [the only reason that that couldn't be relied upon to work swimmingly](https://github.com/liferay/liferay-frontend-projects/issues/354) is that Babel used `Symbol.iterator` in its output and we weren't 100% confident in our polyfill coverage.

In any case, I'll be following up with some pulls to other projects/branches to make turn the rule back on where it is still needed.

One final note: I am turning the rule `'off'` rather than just removing it so as to reflect that we still do use this rule in some places, it's just that we don't have it on by default.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/405

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,7 +26,6 @@ module.exports = {
 		},
 	],
 	rules: {
-		'no-for-of-loops/no-for-of-loops': 'off',
 		'notice/notice': [
 			'error',
 			{
--- a/maintenance/projects/js-toolkit/.eslintrc.js
+++ b/maintenance/projects/js-toolkit/.eslintrc.js
@@ -13,7 +13,6 @@ module.exports = {
 	rules: {
 		'@liferay/liferay/no-dynamic-require': 'off',
 		'no-console': 'off',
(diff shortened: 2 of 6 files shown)


wincent:
Oh, forgot to mention. The reason I am doing this now is because I was using `for`/`of` a bit [over here](https://github.com/liferay-frontend/liferay-portal/pull/874) and don't want to add more suppressions (or rewrite the code).

wincent:
Everybody voted for this over on [the issue](https://github.com/liferay/liferay-frontend-projects/issues/405), and it's trivial, so I'm just going to merge it.

Screenshot 2021-03-08 -175435-TVVGrkv9@2x

Issue #9, opened 2020-09-24
Migrate to eslint-plugin-liferay


wincent:
This is something that I have been putting off for a while, because renaming packages is a pain, but I think we'll eventually have to do it: move eslint-config-liferay to eslint-plugin-liferay.

As I explained in 5850f26d0f9e4c03872a2:

> ESLint surprisingly doesn't make it easy for you to bundle rules with a config, requiring you to publish a separate package. That is, if you have "eslint-config-foo", you have to stick your rules in "eslint-plugin-foo". The inverse doesn't appear to be true: if you have "eslint-plugin-foo", you can bundle configs with it without having to create "eslint-config-foo".
>
> We however, have "eslint-config-liferay", and I don't want to create yet another package, so we use a horrifying but well-contained hack that is hidden away in `utils/local.js` to enable us to bundle local rules with our config.

The trouble with this hack is that:

- It is hideously ugly, requiring us to monkey-patch ESLint at runtime.
- It is sensitive to upstream changes in ESLint (eg. we have to maintain two versions of the hack, one for v5 and one for v6; see 7c67c4c89d7a37bef8).
- Because we have to monkey-patch ESLint, it is sensitive to multiple versions of ESLint being present in the environment, lest it patch the wrong one (see 944deeb32b0b0b and de4388a6b9a672 for examples of how we have to guard carefully against that).

Overall, things are working for now, but this feels like a liability. It would be much better to be able to freely update ESLint without worrying about whether the hack still works, or having to [deal with edge cases stemming from people using the config in different ways](https://issues.liferay.com/browse/IFI-1454).

Tentative plan of action:

- [x] Create an eslint-plugin-liferay placeholder package on NPM to reserve the name — we should do this immediately, then do the following steps whenever we consider optimal.
- [ ] Restructure eslint-config-liferay to be a plug-in, publish it as eslint-plugin-liferay; as part of this, remove [the monkey-patching hack](https://github.com/liferay/eslint-config-liferay/blob/f95833c2aafd2461eb669be7911d6ce4e3831182/utils/local.js).
[...]

PR #430, opened 2021-03-08
fix(eslint-config): update monkey patching to work with newer ESLint

wincent:
While following up on [#429](https://github.com/liferay/liferay-frontend-projects/pull/429) I went to update alloy-editor's lint set-up (which is very old) and noticed that ESLint has been doing some refactoring that means that newer versions don't work properly with our monkey-patching technique, and need to be updated again.

Now, the monkey patching has _always_ sucked, and we know how to restructure our code so that we won't need any more, but it will require quite a bit of work (see [#9](https://github.com/liferay/liferay-frontend-projects/issues/9)); in short, it means creating a plugin package and using that instead of the existing config package, because configs cannot contain plugins but plugins can contain configs.

Anyway, I don't want to do that now because it is probably a full day's work and right now I just want to get unblocked, so this commit goes ahead and updates the monkey patch.

Test plan: Copy the modified file into Alloy Editor's `node_modules` folder after updating to ESlint's latest version (7.21.0), and see linting work/pass with just this tweak to the config:

```diff
diff --git a/.eslintrc.js b/.eslintrc.js
index 40dcef71..f26f795b 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,7 @@ module.exports = {
 		jest: true,
 		node: true,
 	},
-	extends: ['liferay/react'],
+	extends: ['@liferay/eslint-config/react'],
 	globals: {
 		AlloyEditor: true,
 		CKEDITOR: true,
```

### DIFF
--- a/projects/eslint-config/utils/local.js
+++ b/projects/eslint-config/utils/local.js
@@ -31,6 +31,12 @@ let originalResolve;
  * longer uses `require` since:
  *
  * https://github.com/eslint/eslint/commit/6ae21a4bfe5a
+ *
+ * For ESLint 7.8.0+, the config file handling began to be refactored into a
+ * separate package, so we actually patch both just in case...
+ *
+ * https://github.com/eslint/eslint/commit/2bee6d256ae0
+ * https://github.com/eslint/eslintrc/commit/9fb9c5da9b95
  */
 function patch() {
 	const eslint = require.resolve('eslint/package.json', {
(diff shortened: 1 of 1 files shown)


jbalsas:
Just go on and get this merged, looked as ugly to me as you promised it would be but all is green and you know better, so... 👍

wincent:
whats-the-worst-that-could-happen

PR #431, opened 2021-03-08
feat(eslint-config): teach no-duplicate-imports rule about type imports

wincent:
It's legit to import from the same module twice, as long as exactly one of the imports is a type-only import:

```ts
import Thing from './Thing';
import type {ThingT} from './Thing';
```

In the old days, Flow had this syntax and TypeScript didn't. You would instead have to import the type as though it were a value like this, and trust the TypeScript compiler to figure out that you only wanted it for the type definition:

```ts
import Thing, {ThingT} from './Thing';
```

This made it tricky to do things like circular type-only imports (ie. modules A and B wanting to share type definitions with each other). But TypeScript has support for this now and [the way we're currently using this it in liferay-portal](https://github.com/liferay-frontend/liferay-portal/pull/874), we don't even have a choice really, because we're using Babel to strip out type information (Babel knows nothing about the types unless they are flagged syntactically, so if it sees a mixed type/non-type import it doesn't know that it has to strip out part of it).

Presently a draft because tomorrow I need to check or update our other rules to make sure they behave properly with type-only imports (eg. `group-imports`, sort-imports` etc... we have about 7 rules that look at imports in some way, so I want to check them all; I suspect that most of them will be just fine, but one or two may need tweaking).

### DIFF
--- a/projects/eslint-config/plugins/liferay/docs/rules/no-duplicate-imports.md
+++ b/projects/eslint-config/plugins/liferay/docs/rules/no-duplicate-imports.md
@@ -18,3 +18,10 @@ Examples of **correct** code for this rule:
 import {a, g, z} from 'one';
 import x from './x';
 ```
+
+Note the one exception to this rule, that in TypeScript files it is permitted (and sometimes necessary) to have two imports from a module, one that imports _values_ and another that imports _types_:
+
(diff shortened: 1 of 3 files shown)


wincent:

PR #433, opened 2021-03-09
feat(eslint-config)!: teach import-extensions rule about .ts, .tsx

wincent:
This is a companion to [#431](https://github.com/liferay/liferay-frontend-projects/pull/431) (which teaches the no-duplicate-imports rule about type-only imports). I'm going through all the other import-related rules and making sure they do the right thing with TypeScript.

So, in this commit, we add tests to show that both type-only and value-only imports/exports from ".ts" and ".tsx" files work with this rule, which requires us to add those extensions to the list of "banned" extensions.

Technically, a "breaking" change because we'll be potentially catching more lint error than we were before.


wincent:

PR #434, opened 2021-03-09
test(eslint-config): show that various rules already support TS

wincent:
These are just additions to the test suite to show that these import-related rules already support TypeScript without breaking.

Other pulls pushed so far are:

- https://github.com/liferay/liferay-frontend-projects/pull/431 ("teach no-duplicate-imports rule about type imports")
- https://github.com/liferay/liferay-frontend-projects/pull/433 ("teach import-extensions rule about .ts, .tsx")

Both of those required actual changes to the code, unlike this one.

Only remaining one after this will be the `group-imports` rule, but that one is complex enough to warrant its own changelog entry and therefore go in via a separate PR.


wincent:

PR #435, opened 2021-03-09
feat(eslint-config)!: teach sort-imports, group-imports to separate type-only inputs

wincent:
Via [Slack thread](https://liferay.slack.com/archives/C3JBR21HA/p1615278868119400).

We discussed two possible end states:

1.  Mix type-only imports among other imports:

        import X from './X';
        import type {T} from './X';
        import Z from './Z';

2.  Put type-only imports at the end, in a section of their own:

        import X from './X';
        import Z from './Z';

        import type {T} from './X';

We agreed that, while "1" is easier to implement, we like the conceptual separation of "2". Type-only inputs really are totally different beasts (for example, they get elided during compilation, for example).

Changing the group-imports rule alone doesn't produce the desired effect; it merely makes sure that type-only imports start a new group. eg.

```ts
import a from 'a';
import b from 'b';
import type {T} from 'a';
import type {U} from 'b';
import c from 'c';
import d from 'd';
```

will get autofixed to:

```ts
import a from 'a';
import b from 'b';

import type {T} from 'a';
import type {U} from 'b';

import c from 'c';
import d from 'd';
```

but when combined with the effect of the `sort-imports` rule, we move all the type-only imports together to the end and the final result is:

```ts
import a from 'a';
import b from 'b';
import c from 'c';
import d from 'd';

import type {T} from 'a';
import type {U} from 'b';
```

It turns out that this requires a pretty invasive change to the sort-imports rule. Previously, we only sorted within groups, but now that we want to move type-only imports into a separate group at the end, we need to combine the two strategies (sorting within groups and sorting type-only imports to the end).

Note that the error output is a bit "meh" here. In most cases it looks simple:

    imports must be sorted by module name (expected: "./a" << "./b")

But given that you can have a type-only and a value-only import from the same module, you could wind up with confusing output like:

    imports must be sorted by module name (expected: "./a" << "./a")

So, we distinguish that by adding a "(type)" annotation:

    imports must be sorted by module name (expected: "./a" << (type) "./a")

It's a bit "meh" because the text doesn't really tell the whole story, but we need our lint output to be relatively terse (that's the way ESLint rules are; mostly intended to fit on one line so they be displayed inline in IDEs etc), so spelling it all out literally isn't really an option:

    imports must be sorted by module name, with type imports coming last
    (expected: import a from "./a" << import type {A} from "./a")

As such, we settle for the "(type)" disambiguator as a reasonable but not perfect compromise solution. Note that in the case of duplicate imports we already don't have very good output, because we rely on the `no-duplicate-imports` rule to deal with those problems for us; eg:

```ts
import a from './a';
import {b} from './a';
```

would already print a confusing error like:

    imports must be sorted by module name (expected: "./a" << "./a")

I'm not going to stress to much about this, because we have no-duplicate-imports doing its thing and handling type-only imports appropriately since https://github.com/liferay/liferay-frontend-projects/pull/431 and we also haven an autofix for sort-imports, so it's probably all going to be fine despite the suboptimal error messages.

Note how this requires us to re-jig the error reporting in one additional sense. Previously, if we had two "groups":

```
a
c
b

l
m
n

z
y
x
```

We would report minimal differences within each group (ie. we'd say that the order needs to be `"b << c"` + `"z << y << z"`).

Now that we are potentially moving items across groups, we report the misordered items globally (eg. `"b << c << l << m << n << x << y << z"`). This is the price of progress, I guess.

To see how this interacts with the group-imports rule to give us the desired effect, I added some integration tests that run all the rules in the plug-in at once (all of our existing tests run only one rule at a time).

The idea here is to focus on rules that act together to produce a combined effect (all of our rules already have separate tests that verify their behavior when operating individually).

Bonus refactoring included: extracted common `isTypeImport()` helper function that can be used in a few places.

### DIFF
--- a/projects/eslint-config/plugins/liferay/lib/rules/sort-imports.js
+++ b/projects/eslint-config/plugins/liferay/lib/rules/sort-imports.js
@@ -11,6 +11,7 @@ const {
 	hasSideEffects,
 	isAbsolute,
 	isRelative,
+	isTypeImport,
 	withScope,
 } = require('../common/imports');
 
@@ -21,8 +22,8 @@ const DESCRIPTION = 'imports must be sorted';
  * their relative ordering.
  */
 function compare(aKey, bKey) {
-	const [aName, aTieBreaker] = aKey.split(':');
-	const [bName, bTieBreaker] = bKey.split(':');
+	const [aPrefix, aName, aTieBreaker] = aKey.split(':');
+	const [bPrefix, bName, bTieBreaker] = bKey.split(':');
 
 	const cmp = (a, b) => {
 		return a < b ? -1 : a > b ? 1 : 0;
@@ -31,6 +32,7 @@ function compare(aKey, bKey) {
 	//
 
 	return (
+		cmp(aPrefix, bPrefix) ||
 		cmp(ranking(aName), ranking(bName)) ||
 		cmp(aName, bName) ||
 		cmp(aTieBreaker, bTieBreaker)
@@ -38,13 +40,25 @@ function compare(aKey, bKey) {
 }
 
 /**
- * Ideally we'd just sort by source, but in case there is a pathological case
- * where we import from the same module twice, we need to include information
- * about the import specifiers in the sort key.
+ * Ideally we'd just sort by source, but we need to handle two "edge" cases:
+ *
+ * - When we have type-only imports, those always come last.
+ * - When we import from the same module twice (not always an error; can
+ *   legitimately happen if we have both a type-only and a value-only import
+ *   from the same module).
+ *
+ * So, the sort key must include information about both the import kind
+ * (this serves as a prefix in the sort key) and any import specifiers
+ * that may be present (these serve to construct a trailing "tiebreaker"
+ * suffix in the sort key).
  */
 function getSortKey(node) {
 	const source = getSource(node);
 	let tieBreaker;
+
+	// Type-only imports always go last.
+
+	const prefix = isTypeImport(node) ? 'type-import' : 'normal-import';
 
 	if (node.type === 'ImportDeclaration') {
 		const specifiers = node.specifiers.map((specifier) => {
@@ -102,7 +116,7 @@ function getSortKey(node) {
 		tieBreaker = '';
 	}
 
-	return `${source}:${tieBreaker}`;
+	return `${prefix}:${source}:${tieBreaker}`;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


wincent:
Have to go pick up the kids now, but when I get back I want to:

1. Add integration test like I mentioned above.
2. Double check that this does the right thing with respect to various edge-casey things (like side-effect-only imports and such).

wincent:
As feared, had to do some deeper surgery to get this to work, as explained in third commit:

---

**feat(eslint-config)!: teach sort-imports to sort type-only imports to the end**

So, it turns out that this requires a bit more of an invasive change to the sort-imports rule. Previously, we only sorted within groups, but now that we want to move type-only imports into a separate group at the end, we need to combine the two strategies (sorting within groups and sorting type-only imports to the end).

Note how this requires us to re-jig the error reporting. Previously, if we had three "groups":

```
a
c
b

l
m
n

z
y
x
```

We would report minimal differences _within_ each group (ie. we'd say that the order needs to be `"b << c"` + `"z << y << z"`).

Now that we are potentially moving items across groups, we report the misordered items globally (eg. `"b << c << l << m << n << x << y << z"`). This is the price of progress, I guess.

To see how this interacts with the group-imports rule to give us the desired effect, I added some integration tests that run all the rules in the plug-in at once (all of our existing tests run only one rule at a time).

I'll follow up with some more integration tests before I merge this pull; for now consider this a sample.

---

Will add some more tests tomorrow (that's why I am leaving it as draft), but this is basically ready for review now. I might end up squishing commits "1" and "3" together seeing as "1" alone wasn't enough to get the job done.

projects/eslint-config/plugins/liferay/lib/rules/sort-imports.js
```diff
@@ -31,21 +32,34 @@ function compare(aKey, bKey) {
...
+ * that may be present (these serve to construct a trailing "tiebreaker"
+ * suffix in the sort key).
  */
 function getSortKey(node) {
 	const source = getSource(node);
 	let tieBreaker;
 
+	// Type-only imports always go last.
+
+	const prefix = isTypeImport(node) ? 'type-import' : 'normal-import';
```
wincent, line 61: Strictly speaking, we don't need to bake this indicator of "type-import-ness" into the key itself (because we are handling these specially further down anyway). I think I could delete this and all the tests would still pass. But, it doesn't really hurt too much for the ordering info to be in the key anyway, so 🤷

jbalsas:
> I added some more integration tests, so going to take this out of draft and if it is green, merge it.

DO IT!!

PR #436, opened 2021-03-09
fix: add languageId parameter to module requests

izaera:
This PR implements the build part of [LPS-127736](https://issues.liferay.com/browse/LPS-127736).

I will send a server side PR once this is merged and released. Its WIP can be found [here](https://github.com/liferay-frontend/liferay-portal/pull/883)

I've tested it locally and it fixes the LPS and seems not to break anything inside portal.

Note that I have made a refactor for the webpack stuff because it was growing quickly and it was making the `utils` folder unmanageable.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/__fixtures__/webpack/tweakWebpackConfig/npmscripts.config.js
+++ b/projects/npm-tools/packages/npm-scripts/__fixtures__/webpack/tweakWebpackConfig/npmscripts.config.js
@@ -3,10 +3,12 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+const fs = require('fs');
 const os = require('os');
+const path = require('path');
 
 module.exports = {
 	build: {
-		temp: os.tmpdir(),
+		temp: fs.mkdtempSync(path.join(os.tmpdir(), 'npm-scripts-')),
 	},
 };


projects/npm-tools/packages/npm-scripts/__fixtures__/webpack/tweakWebpackConfig/npmscripts.config.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const os = require('os');
+
+module.exports = {
+	build: {
+		temp: os.tmpdir(),
```
izaera, line 10: This is needed so that test don't leave files hanging around. I could as well delete them, but still, if the tests failed or where interrupted the source tree would have been polluted, so better move temp things out.

projects/npm-tools/packages/npm-scripts/src/utils/createTempFile.js
```diff
@@ -9,8 +9,6 @@ const path = require('path');
 const SignalHandler = require('../utils/SignalHandler');
 const getMergedConfig = require('./getMergedConfig');
 
-const BUILD_CONFIG = getMergedConfig('npmscripts', 'build');
```
izaera, line 12: See the commit description. I believe this has become an anti-pattern in this project. It was not the case before, but I had to refactor the JS Toolkit long ago for the very same reason.

The fact is that we cannot rely on who's calling the functions of this module and, by getting the configuration in advance we are effectively caching it and making the assumption that the functions in the module will always be invoked for the very same project. This is -at the very least- false in the unit test, and can be generally false as soon as we use the tools for more than one project in the same process image.

izaera: Of course, it would be safer to pass the config as parameter instead of treating it as a global, but for now, using `chdir` should be enough. In fact, it is so widely assume that process.cwd() is the project directory that it is quite difficult to maintain code that works without that assumption.

projects/npm-tools/packages/npm-scripts/src/utils/writeLibInitEntryPoint.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+module.exports = function (filePath) {
+	const code = `
+const { transformChunkScriptFilename } = 
+	window[Symbol.for('__LIFERAY_WEBPACK_HELPERS__')];
+
+const __saved_webpack_get_script_filename__ = __webpack_get_script_filename__;
+
+__webpack_get_script_filename__ = (chunkId) => {
+	return transformChunkScriptFilename(
```
jbalsas: So, I'm assuming we'll export `transformChunkScriptFilename` from within `DXP` to make it get in here? 

If that's the case, I don't mind the level of indirection, but I'm a bit worried about race conditions since we'll need to make sure wherever those helpers come from are already there before we run any webpack code... is that guaranteed and already in place?

projects/npm-tools/packages/npm-scripts/src/webpack/createFederationConfig.js
```diff
@@ -92,16 +98,12 @@ module.exports = async function () {
...
-		'federation.plugin.config.json',
-		JSON.stringify(federationPluginConfig, null, 2),
-		{autoDelete: false}
-	);
-
 	return {
 		context: process.cwd(),
 		devtool: 'source-map',
-		entry: nullJsFilePath,
+		entry: {
```
izaera, line 104: This is an entry point that, because is called the same as the library, signals webpack to use it as initialization code.

See [this webpack issue](https://github.com/webpack/webpack/issues/12834) for more information.

izaera: Added this :point_up: to the commit message for the future generations.

projects/npm-tools/packages/npm-scripts/src/utils/writeLibInitEntryPoint.js
izaera: Yes it is, because if that hadn't been loaded, you could not be executing webpack code :-). All webpack code is bootstrapped in that same place.

izaera: It will be [here](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-module-launcher/src/main/resources/META-INF/resources/webpack_federation.js), our "webpack loader".

projects/npm-tools/packages/npm-scripts/__fixtures__/webpack/tweakWebpackConfig/npmscripts.config.js
wincent: One thing to bear in mind about `tmpdir()` is that it might not provide a very good location. For example, on Linux, is is going to return `/tmp` which is just a gigantic bucket shared by all users. On macOS it will at least return something that is user-specific (eg. `'/var/folders/xx/huge-long-string-of-random-digits/x`) and which I think is different on every boot.

If you want each test run to have an independent space where it can write temporary files to without collisions (with other users, or with other runs by the same user), you need to combine `tmpdir()` with `fs.mkdtempSync()` like we do in a number of places in this repo; for [example](https://github.com/liferay/liferay-frontend-projects/blob/8954af4210985cc17eb0da7a2bd30026809585cc/projects/npm-tools/packages/npm-scripts/test/utils/minify.js#L19):

```js
const temp = fs.mkdtempSync(path.join(os.tmpdir(), 'blah-'));
```

projects/npm-tools/packages/npm-scripts/src/utils/createTempFile.js
wincent: > See the commit description.

Not sure which one you mean exactly, but ok. 😁

projects/npm-tools/packages/npm-scripts/src/webpack/createFederationConfig.js
```diff
@@ -32,9 +33,14 @@ module.exports = async function () {
...
 	const webContextPath = bnd['Web-ContextPath'] || name;
 
-	const {filePath: nullJsFilePath} = createTempFile('null.js', '');
-	const {filePath: mainFilePath} = createTempFile('index.federation.js', '');
+	const {filePath: libInitJsFilePath} = createTempFile('libInit.js', '', {
+		autoDelete: false,
+	});
+	const {filePath: mainFilePath} = createTempFile('index.federation.js', '', {
+		autoDelete: false,
+	});
```
wincent, line 41: To the point about debuggability and leaving temp files around (until the OS reclaims them), is there anywhere in the build output where the actual location is visibly logged? Asking because it matters on macOS where the `tmpdir()` is going to be random unguessable gibberish; the only way to know where to look is if the console shows where the files went (more so if you take my suggestion up above of combining with `mkdtempSync()`).

projects/npm-tools/packages/npm-scripts/src/webpack/tweakWebpackConfig.js
```diff
@@ -45,6 +46,12 @@ async function tweakWebpackConfig(webpackConfigPath) {
 		mergeBabelLoaderOptions(webpackConfig)
 	);
 
+	createTempFile(
+		'webpack.config.json',
+		JSON.stringify(arrayConfig, null, '\t'),
```
wincent, line 51: attention-to-detail

(for the tabs)

projects/npm-tools/packages/npm-scripts/src/webpack/writeLibInitEntryPoint.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+module.exports = function (filePath) {
+	const code = `
+const { transformChunkScriptFilename } = 
+	window[Symbol.for('__LIFERAY_WEBPACK_HELPERS__')];
+
+const __saved_webpack_get_script_filename__ = __webpack_get_script_filename__;
+
+__webpack_get_script_filename__ = (chunkId) => {
```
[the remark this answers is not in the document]
wincent: Another one for future generations (at least the future generations with access to our Slack channel), [this is the thread](https://liferay.slack.com/archives/C3JBR21HA/p1615282339126800) where the magical way in which webpack wires everything up is explained.

projects/npm-tools/packages/npm-scripts/src/webpack/createFederationConfig.js
izaera: I was thinking in refactoring the temp files stuff a bit, to make it clearer, because since we added the feature to move files apart of source tree (for the `deployFast` command) we changed the paradigm, but the code doesn't reflect it fully (IMO).

So, first, a bit of history...

The `createTempFile` method used to put files in the source tree and delete them on finish to avoid leaving a modified source tree. Because watch tools (specifically `deployFast` + `gradle`) got fooled by that, I created this [config entry](https://github.com/liferay/liferay-frontend-projects/blob/8954af4210985cc17eb0da7a2bd30026809585cc/projects/npm-tools/packages/npm-scripts/src/config/npmscripts.config.js#L176) to move things out of source tree.

That means that:

1. Even though the method is called `createTempFile` it is not really used to create temp files. Or it is, but I like to think of them more as "derived" files (intermediate build artifacts) than the typical random files you would put in `/tmp`. 
2. By default they are written to the build directory (the logical place for intermediate artifacts, probably)
3. There's no need to delete them since they will be removed whenever you `clean` the output.

What can happen now is that someone sees `createTempFile` and thinks that it is used to create true temporary files in the `/tmp` (or equivalent) dir. I don't think we do that in any place in `npm-scripts` now, but if we need it in the future I'm sure we will mess it up :sweat_smile: .

So, my proposal is to rename this method to something else, stop it from autodeleting files and, probably, stop it being configurable and simply put the files in the build directory in a specific place, since they are well known files and, in general, it's very handy to have them available, given the amount of stuff we generate under the hood.

What do you think?

izaera: In the case of `liferay-portal` what you see after running `gradle deploy` is this:

imagen

And specifically the `webpack.config.json` file is full of generated magic options :scream:

wincent (approved):
All looks fine to me, in the sense of "I barely understand this and couldn't have written it myself, but it seems to be in the right language."

see-it-ship-it

projects/npm-tools/packages/npm-scripts/src/webpack/tweakWebpackConfig.js
izaera: :joy:

projects/npm-tools/packages/npm-scripts/src/webpack/createFederationConfig.js
wincent: Yeah, my comment about logging was specifically thinking about debugging _test failures_, where the path may not be so predictable.

> So, my proposal is to rename this method to something else, stop it from autodeleting files and, probably, stop it being configurable and simply put the files in the build directory in a specific place, since they are well known files and, in general, it's very handy to have them available, given the amount of stuff we generate under the hood.
>
> What do you think?

I think _I_ wouldn't change it, but only because I don't remember all the places it is used and wouldn't want to have to analyze them and be sure that it is ok to make such a change. But if you feel confident in doing it and see the value, I wouldn't try to talk you out of it.

izaera: I think it is almost easy. I tried to set autoDelete to false yesterday and had some problems with the tests, but the code seemed to work correctly for liferay-portal. However it's low priority, so I will try to find a more relaxed moment :-D.

projects/npm-tools/packages/npm-scripts/src/webpack/writeLibInitEntryPoint.js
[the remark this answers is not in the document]
izaera: Maybe you could offer a free subscription to our Slack channel when people purchase a `portool` license... :thinking:

projects/npm-tools/packages/npm-scripts/__fixtures__/webpack/tweakWebpackConfig/npmscripts.config.js
izaera: I'll push a commit for this. In a regular desktop it is very rare that parallel tests ever run, but in CI/QA environments it could happen, you never know...

projects/npm-tools/packages/npm-scripts/src/utils/createTempFile.js
izaera: [This one](https://github.com/liferay/liferay-frontend-projects/pull/436/commits/c0aa550c592da8611ddf8b746df115fb3101e1b2). I referred to the commit containing this change, but it seems that  GH offers an easy way to navigate to it from here...

projects/npm-tools/packages/npm-scripts/__fixtures__/webpack/tweakWebpackConfig/npmscripts.config.js
izaera: done

Issue #432, opened 2021-03-09
Turn on no-eval rule by default


wincent:
This came up in [code review the other day](https://github.com/liferay-frontend/liferay-portal/pull/866#pullrequestreview-604879515):

> i'm surprised `eval` is allowed, maybe it's because we're using it in some ancient scripts

We don't explicitly _disallow_ it, we just [use the `eslint:recommended` settings ](https://github.com/liferay/liferay-frontend-projects/blob/3140da045347c4577c7677a664eb5711ba85544b/projects/eslint-config/index.js#L14), and [those](https://eslint.org/docs/rules/) don't include [the `no-eval` rule](https://eslint.org/docs/rules/no-eval).

However, there are hardly any instances of `eval` out there though, so it would be pretty easy for us to turn this one on:

- [modules/apps/commerce/commerce-frontend-js/src/main/resources/META-INF/resources/components/data_renderers/ActionLinkRenderer.js:96](https://github.com/liferay/liferay-portal/blob/dc8652d18acf/modules/apps/commerce/commerce-frontend-js/src/main/resources/META-INF/resources/components/data_renderers/ActionLinkRenderer.js#L96)
- [modules/apps/commerce/commerce-frontend-js/src/main/resources/META-INF/resources/components/data_renderers/ActionsDropdownRenderer.js:121](https://github.com/liferay/liferay-portal/blob/dc8652d18acf/modules/apps/commerce/commerce-frontend-js/src/main/resources/META-INF/resources/components/data_renderers/ActionsDropdownRenderer.js#L121)
- [modules/apps/commerce/commerce-frontend-js/src/main/resources/META-INF/resources/components/dropdown/Dropdown.js:43](https://github.com/liferay/liferay-portal/blob/dc8652d18acf/modules/apps/commerce/commerce-frontend-js/src/main/resources/META-INF/resources/components/dropdown/Dropdown.js#L43)
- [modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/deprecated.js:58](https://github.com/liferay/liferay-portal/blob/dc8652d18acf/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/deprecated.js#L58)
[...]

PR #438, opened 2021-03-10
feat(eslint-config)!: activate no-eval rule

wincent:
As noted in the related issue, we have only 6 instances of `eval()` in liferay-portal right now (and, ironically, 4 in this monorepo — cc @izaera 😉), so this is something we can turn on without too much churn.

See: https://eslint.org/docs/rules/no-eval

Closes: https://github.com/liferay/liferay-frontend-projects/issues/432


wincent:
Doing this now because it is quick and easy, and I am about to cut a release for `eslint-config` anyway.

Issue #417, opened 2021-02-24
Extend sort-class-names rule to handle any attribute ending with "className"


wincent:
Requested here: https://github.com/brianchandotcom/liferay-portal/pull/99097#issuecomment-784377971

We have been sorting class names in JSX [since October 2019](https://github.com/liferay/eslint-config-liferay/pull/114), but note that only for the React `className` attribute. The attribute Brian is talking about here is named `triggerClassName`.

It might be reasonable to do this for any attribute name that matches `(\bc|[a-z]C)lassName` and has a string value. It seems pretty unlikely this would cause many (any) false positives.

wincent:
Just noting here that whatever we do here, we should do the same for the related rules too (`trim-class-names`, `no-duplicate-class-names`).

PR #439, opened 2021-03-10
feat(eslint-config)!: disallow unnecessary string interpolation in JSX

wincent:
By way of: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md

Will fix:

```jsx
<div className={'yo'}>{'hi'}</div>
```

to:

```jsx
<div className="yo">hi</div>
```

Closes: https://github.com/liferay/liferay-frontend-projects/issues/421

PR #440, opened 2021-03-10
feat(eslint-config)!: teach class-names rules about "customClassName" attributes

wincent:
Extends our three existing `class-names` rules:

- `no-duplicate-class-names`
- `sort-class-names`
- `trim-class-names`

to handle `"customClassName"` attributes with string keys as well, like these:

```jsx
<CustomPopover triggerClassName="   foo foo bar  " />
```

Which would be fixed to:

```jsx
<CustomPopover triggerClassName="bar foo" />
```

Closes: https://github.com/liferay/liferay-frontend-projects/issues/417


wincent:
Like https://github.com/liferay/liferay-frontend-projects/pull/438 and https://github.com/liferay/liferay-frontend-projects/pull/439, throwing this one in now before the release because it is low-hanging fruit.

wincent:
ninja-merge

PR #442, opened 2021-03-10
chore: update ESLint version from 7.10.0 to 7.21.0

wincent:
This updates both the version that we use internally in the monorepo and also the dependency that npm-scripts carries with it into liferay-portal.

This combo (latest `@liferay/eslint-config` + latest `eslint`) is working well [over in good old alloy-editor](https://github.com/liferay/alloy-editor/pull/1469), so I think it's safe to roll out everywhere.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
 		"chai-fs": "^2.0.0",
 		"combohandler": "^0.4.0",
 		"copyfiles": "^2.4.0",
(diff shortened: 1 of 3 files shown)


wincent:
This is all obviously awesome, so in it goes.

jbalsas:
> This is all obviously awesome, so in it goes.

iveseenworse

PR #443, opened 2021-03-11
fix(npm-scripts): change Angular adaptation to make it work with Angular 10/11

izaera:
Tested with master and two Angular projects:

- One with Angular 9 to make sure we don't break previous code
- One with Angular 11 to make sure it works with newer versions

It would have been great to have tested this more thoroughly, but there are more hi priority things going on and I can't dedicate as much time as I wish. However, I'm more or less confident that this should fix the issue, so we may merge, release and tell GS to try it and confirm.


maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/loader/normalize-angular-output.ts
```diff
@@ -0,0 +1,71 @@
...
+ * how webpack produced its output.
+ *
+ * This is because Angular version <=10 produce artifacts for ES5 and ES2015
+ * targets, while Angular versions >=10 only produce one type of artifact, not
+ * differentiated by file name.
+ */
+export default function (context: BundlerLoaderContext): BundlerLoaderReturn {
+	const {filePath, log} = context;
+
+	if (filePath.includes('-es') && !filePath.includes('-es5.')) {
```
wincent, line 27: Is it possible that I could have a file with a substring like `products-essential-oils`? Don't know what the naming conventions in Angular are... Maybe we should tighten this up a bit?

maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/loader/normalize-angular-output.ts
```diff
@@ -0,0 +1,71 @@
...
+		context.content = undefined;
+		context.extraArtifacts = {};
+
+		log.info(
+			'normalize-angular-output',
+			`Discarding output because source is not ES5 code`
+		);
+
+		return undefined;
```
wincent, line 36: This line offends me.

offended

maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/loader/normalize-angular-output.ts
```diff
@@ -0,0 +1,71 @@
...
+			delete context.extraArtifacts[filePath];
+
+			log.info(
+				'normalize-angular-output',
+				`Diverted extra artifact ${filePath} to ${newFilePath}`
+			);
+		}
+	});
+
+	return undefined;
```
wincent, line 70: too-much

wincent (approved):
> However, I'm more or less confident that this should fix the issue, so we may merge, release and tell GS to try it and confirm.

Sounds reasonable to me, especially if you are just going to do the toolkit release. (If you are going to do a full release into liferay-portal though with npm-scripts, that may be worth holding off a day or two on — I cut a release last night that I wanted to send to Brian today, and it will need some manual lint fixes to go with it.)

maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/loader/normalize-angular-output.ts
izaera: It is required by `tsc` because the  type is defined as `undefined|something`.

maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/loader/normalize-angular-output.ts
izaera: :thinking: That's true. But I'm not sure if I can tighten it more because that `-es` is in the middle of the name I think. Let me see what I can do, I may put an AI in place just to make sure this doesn't fail. Otherwise it will fail one day after the release :roll_eyes: ...

izaera:
> If you are going to do a full release into liferay-portal though with npm-scripts,

This is only for adaptations, so nothing to do with the portal :+1:

maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/loader/normalize-angular-output.ts
wincent: Really? You don't even [have strict mode on](https://github.com/liferay/liferay-frontend-projects/blob/abe0815c66bc109763902f2ec9b029770ba6c721/maintenance/projects/js-toolkit/tsconfig.json), and AFAIK, not even [strict mode](https://www.typescriptlang.org/tsconfig#strict) will complain about this by default (you have to explicitly turn on `noImplicitReturns`, which is [off by default](https://www.typescriptlang.org/tsconfig#noImplicitReturns)).

ie. paste this into the [TS playground](https://www.typescriptlang.org/play?noImplicitReturns=false&ssl=1&ssc=1&pln=3&pc=2#code/GYVwdgxgLglg9mABADwBQEoBciBGc4A2ApgIZIA+i4AJkcDGEdYgN4BQiniMwiqAsiSgALAHQAnMtTgBbDIgB8iUQCZ0icUSghxSKOJBEA3GwC+QA), ensuring that `noImplicitReturns` is off:

```ts
function x(): boolean | undefined {
    if (Math.random() > .2) return true;
}
```

wincent: I think the reason the default is off is because the "philosophy" of TS (per [their design goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals)) is to allow people to write "natural, idiomatic" JS, and not writing `return undefined;` is an example of that.

maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/loader/normalize-angular-output.ts
wincent: It might be as simple as changing `'-es'` to `'-es.'`.

izaera: OK, it's impossible, because these files have fixed names, they don't depend on the user's project:

```
main-es2015.f7ffa6ef99984376ef2f.js
main-es5.f7ffa6ef99984376ef2f.js
polyfills-es2015.8110071a28231e4fad12.js
polyfills-es5.2ef310e05779620c39a1.js
runtime-es2015.409e6590615fb48d139f.js
runtime-es5.409e6590615fb48d139f.js
styles.3ff695c00d717f2d2a11.css
```

So, they are always `main`, `polyfills` and `runtime`. If they were changed the adaptation wouldn't work because the AMD loader wouldn't find them.

izaera:
So, I'm going to merge this and make a new release of the JS Toolkit :crossed_fingers:

wincent:
Oops, just noticed the scope is wrong (`npm-scripts` instead of `js-toolkit`). Oh well. 🤷

PR #449, opened 2021-03-15
docs(guidelines): add recommendations on state management for React

wincent:
This comes out of [liferay-frontend#874](https://github.com/liferay-frontend/liferay-portal/pull/874).

As discussed there, we want to make sure that people know how/when to use the new global state API, and a huge prerequisite for that is knowing when _not_ to use it, which in turn means knowing how to get the most out of React's built-in state management abstractions, which will continue to cover probably 95% of use cases.

### DIFF
--- a/guidelines/dxp/react_state_management.md
+++ b/guidelines/dxp/react_state_management.md
@@ -1,0 +1,258 @@
+# React state management
+
+State management is a complicated topic which we can't possibly hope to cover fully in a guidelines document like this, but this page provides three things to get you started:
+
+1. General principles for how to manage state in React.
+2. Guidelines on the use of Liferay's global state API, as provided by `@liferay/frontend-js-state-web`.
+3. Links to further reading.
+
+## General principles
+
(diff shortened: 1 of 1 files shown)


wincent:
Just a draft because I haven't finished writing it yet, and I need to switch tasks, so this is a sneak preview, and obviously open to suggestions as well.

jbalsas:
Looks great so far! As early feedback goes... could we identify some good examples in our own codebase to accompany some of the recommendations? 

just-sayin

wincent:
> could we identify some good examples in our own codebase

Now that could take a while... 😂

Just kidding. I am sure there are some good examples in there somewhere.

wincent:
Now that some other pieces are in place or almost merged, "finished" the doc, so taking it out of draft. Haven't linked to examples yet, but it probably makes sense to do that with some real usages of the new state API ported.

wincent:
Chatted with @jbalsas about this, kind of, because I couldn't remember if I'd merged it or not. So that counts as review, right? Going to merge.

bryceosterhaus:
Great docs for this!

PR #451, opened 2021-03-15
feat(npm-scripts): add a shim for npm's `process` to webpack federation build

izaera:
We need this to be able to build the portal without `frontend-js-node-shims` (see [LPS-128668](https://issues.liferay.com/browse/LPS-128668))

Tested in local `ant all`

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/package.json
+++ b/projects/npm-tools/packages/npm-scripts/package.json
@@ -61,6 +61,7 @@
 		"minimist": "^1.2.0",
 		"path-browserify": "^1.0.1",
 		"prettier": "^2.1.2",
+		"process": "^0.11.10",
 		"react": "*",
 		"react-dom": "*",
 		"react-test-renderer": "*",
--- a/projects/npm-tools/packages/npm-scripts/src/webpack/createFederationConfig.js
+++ b/projects/npm-tools/packages/npm-scripts/src/webpack/createFederationConfig.js
@@ -137,6 +137,7 @@ module.exports = async function () {
 		resolve: {
 			fallback: {
 				path: require.resolve('path-browserify'),
+				process: require.resolve('process/browser'),
 			},
 			plugins: [createWebpackSoyResolver()],
 		},
--- a/yarn.lock
+++ b/yarn.lock
@@ -12276,6 +12276,11 @@ process-nextick-args@^2.0.0, process-nextick-args@~2.0.0:
   resolved "https://registry.yarnpkg.com/process-nextick-args/-/process-nextick-args-2.0.1.tgz#7820d9b16120cc55ca9ae7792680ae7dba6d7fe2"
   integrity sha512-3ouUOpQhtgrbOa17J7+uxOTpITYWaGP7/AhoR3+A+/1e9skrzelGi/dXzEYyvbxubEF6Wn2ypscTKiKJFFn1ag==
 
+process@^0.11.10:
+  version "0.11.10"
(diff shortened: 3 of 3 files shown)


wincent (approved):
My feeling about adding another dependency:

sobbing

The main usage we have in our own code is `process.env.NODE_ENV === 'whatever'` checks, which should get rewritten by Babel to `true`/`false` and dead-code-stripped as appropriate (I sure hope that's still working under webpack federation, right — can you confirm?), but I believe some naughty modules we depend on do use `process` for other purposes. 😢

`#SullenApproval`

izaera:
I added that because `process` is still used in the webpack build (at least in development mode). I suppose webpack knows how to prune the `if` but not that it can remove the whole `import`.

In any case, sooner or later, we will need to add this for external users, so.... :shrug: 

Maybe we can revisit it then, to let it be configured from outside of `npm-scripts` but even if we do that, we will still need to add `process` to `liferay-portal`, so there's no way to escape from this unless we stop using `process`.

If it makes you happy think that there are many dependencies out there that already assume that `process`, `util`, even `fs` :scream: will be shimmed by webpack so they don't put too much effort in removing them for browser builds :roll_eyes:

izaera:
> I suppose webpack knows how to prune the if but not that it can remove the whole import.

This is because even if babel loader prunes the if, the `require('process')` would still be present in the header of the pruned file and webpack will fail to find it, though we may need to look at this in more detail because in theory it should be feasible to detect that the require is not needed either.

However, we needed `frontend-js-node-shims` before, so I guess webpack/babel didn't remove the `require` before, either.

izaera:
OK, I was referring to [this](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-react-web/src/main/resources/META-INF/resources/js/index.es.js#L26) and I see know why it is not pruned :-(

The rest of the projects don't complain about it.

izaera:
I have created https://issues.liferay.com/browse/LPS-129187 to delegate the shims to the project (that way we may get rid of this dep in `npm-scripts` and put it just in `frontend-js-react-web` if we want to centralize the hack). However I won't do the LPS nor the configuration yet because I want to continue migrating better than polishing the nuances of the build process.

wincent:
> OK, I was referring to [this](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-react-web/src/main/resources/META-INF/resources/js/index.es.js#L26) and I see know why it is not pruned :-(

Yeah, I know about that one (I added it, after days of fighting with ReactDND... total nightmare getting that to work with v2 of our build/bundling infra.)

izaera:
I'm merging this, as I will need to send another PR to fix the regression with `frontend-js-metal-web`

jbalsas:
Hey @izaera, I'm not convinced about this move. 

On one hand, sure, we'd love to get rid of `frontend-js-node-shims` but on the other I'm afraid of how this will interact with everything else we have and feels like we already have too many moving parts at the moment.

I would expect that instead of adding a whole `process` fallback, we added the proper transforms so that our code would not depend on it, similar to what we do with `process.env`. I'd prefer to have a closer look at what are we using `process` for and avoid shimming it at all if possible.

Also, this close to our first release, maybe we can live with the old module in there to avoid adding unnecessary instability.

jbalsas:
Going to close here in line with other cleanup we're doing in preparation for GA1 and shifting focus towards other tasks for a little bit 😘

PR #454, opened 2021-03-17
fix: don't treat ourselves as a valid import

izaera:
This is needed since we allowed to inherit the base configuration from the liferay-portal project in
npm-scripts.

This is because now we don't have a way to disable the inherited imports other than setting them
to false one by one. But that makes it very difficult to configure projects, because if we add a new
import to the base config, we need to check all other projects one by one.

With this change, we can use the same base config in all projects without making the build fail
because a project excludes itself from the JAR because the bundler thought it was imported
(rather than the project being built).

Example: the [current base npm-scripts config](https://github.com/liferay/liferay-portal/blob/8cf1bfb038cd72870018de6b9f37c082b71e4519/modules/npmscripts.config.js#L32) in portal imports react from frontend-js-react-web and that makes bundler think that it shouldn't include react inside frontend-js-react-web JAR :roll_eyes: .

A way to fix that would be creating this npmscripts.config.js file in `frontend-js-react-web`:

```javascript
module.exports = {
	build: {
		bundler: {
			config: {
				imports: {
				  "@liferay/frontend-js-react-web": false,
...
```

But that's ugly and doesn't scale well because if we add a new import for an existing project we need to remember to modify the existing project. This, in addition to being error prone, may be quite hard to debug so better avoid it with this fix.

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-bundler-plugin-exclude-imports/src/__tests__/index.test.js
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-bundler-plugin-exclude-imports/src/__tests__/index.test.js
@@ -25,6 +25,9 @@ it('empties files when package is listed in imports', () => {
 			},
 			config: {},
 			log: new PluginLogger(),
+			rootPkgJson: {
+				name: 'root',
+			},
 		},
 		{
 			files,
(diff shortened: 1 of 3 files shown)


izaera:
Going to merge this to make a release for portal, given the urgency.

Since the fix makes a lot of sense I don't expect it to break any build out there. In other words, I can't think of any situation where one would want to exclude its own dependencies from the JAR when it is him who's providing them according to the config.

jbalsas:
treatyoself

wincent:
Well, I'm glad you figured out how to do this. (I would probably have needed 3 days to come up with the same change.)

jbalsas:
If I'm reading this correctly... can we now get rid of all those `.npmbundlerrc` files once we get to `liferay-portal`?

We currently have `29` of those and like `11` of `npmscripts.config.js`. Seeing as the latter has a `bundler` key... please tell me we can at least remove one 🙏

izaera:
> can we now get rid of all those .npmbundlerrc files 

Not really, but in any case, I have created https://issues.liferay.com/browse/LPS-129367 and https://github.com/liferay/liferay-frontend-projects/issues/457 to address it :+1:

Issue #386, opened 2021-01-25
Update the no-metal-plugin eslint rule to reflect that 7.4 code should not make use of metal modules


dsanz:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [ ] :gift: Feature request
-   [x] :woman_shrugging: Other

### Description
Some time ago we started to [deprecate metal modules](https://issues.liferay.com/browse/LPS-96715). More recently, we started to [remove metal-dom usages](https://issues.liferay.com/browse/LPS-122383), and now we're [removing all usages of metal](https://issues.liferay.com/browse/LPS-124481) in DXP. This should lead to a deprecation of all metal modules during 7.4 release cycle.

In practise, this means that any developer who uses metal modules in their code should be warned of this deprecation process when building their modules.

**Desired behavior:**
The `no-metal-plugins` preset includes information about all metal modules, in a similar way as it does currently for a few of them. Documentation about the preset is updated.

**Current behavior:**
The `no-metal-plugins` preset includes information about some metal modules as part of the [LPS-96715](https://issues.liferay.com/browse/LPS-96715) deprecation effort

**Repro instructions (if applicable):**

**Other information (environment, versions etc):**
This updated preset shall be used for modules targeting DXP 7.4 as the most recent tasks dealing with usage removals will take effect in that release. It can be used in current master too.
For older releases (7.3 and 7.2), current preset has to be kept.

This task might need to wait till all usage removal tasks are done so that it can provide the proper explanation messages.

wincent:
For the lazy, [here is the list of packages](https://github.com/liferay/liferay-frontend-projects/blob/6252b50e2082f30c6f438cf59a1b2b17d1b1e0b3/projects/eslint-config/plugins/portal/lib/rules/no-metal-plugins.js#L6-L17) we currently lint against in the `no-metal-plugins` rule.

The change should be pretty easy then:

- Add any additional plugins we want to blacklist to the rule.
[...]

PR #456, opened 2021-03-17
feat(eslint-config)!: add several packages to the no-metal-plugin eslint rule

kresimir-coko:
Fixes https://github.com/liferay/liferay-frontend-projects/issues/386

Hey crew, first time doing something like this so please let me know if I missed something or done something wrong, but here's what I've done.

- added `affix`, `position`, and `throttle` and others to the list of deprecated modules. There are leftover modules but they're still used
- fixed the warning text of `keyboard-focus`. I couldn't find our own implementation, other than `useKeyboardNavigation` but that one didn't look the same, so just referenced a default focus behaviour
- added tests for the new deprecated modules

## Notes: 
- It wasn't clear to me if I'm supposed to go to do what @wincent mentioned in https://github.com/liferay/liferay-frontend-projects/issues/386#issuecomment-766996855 - the 7.2 and 7.3 config update
- `metal-isomorphic` is like a testing library it seems, so not sure what to do about that one
- `metal-pagination` is a component with UI, so not sure what to do about that one
- `metal-router` is a Metal variant of the React Router, do we recommend people to use that one, or something else, not sure
- `metal-toggler` is a component with UI, same as pagination
- `metal-web-component` is a specific case too, not sure what to do with it either

### DIFF
--- a/projects/eslint-config/plugins/portal/lib/rules/no-metal-plugins.js
+++ b/projects/eslint-config/plugins/portal/lib/rules/no-metal-plugins.js
@@ -4,14 +4,24 @@
  */
 
 const DEPRECATED_MODULES = [
+	'metal-affix',
 	'metal-ajax',
+	'metal-anim',
 	'metal-aop',
+	'metal-assertions',
 	'metal-clipboard',
 	'metal-debounce',
+	'metal-dom',
+	'metal-key',
 	'metal-keyboard-focus',
+	'metal-multimap',
+	'metal-path-parser',
+	'metal-position',
 	'metal-promise',
+	'metal-scrollspy',
 	'metal-storage',
 	'metal-structs',
+	'metal-throttle',
 	'metal-uri',
 	'metal-useragent',
 ];
@@ -44,22 +54,42 @@ module.exports = {
 		},
 		fixable: null,
 		messages: {
+			'no-metal-affix':
+				'metal-affix is deprecated; use native offset instead',
 			'no-metal-ajax':
 				"metal-ajax is deprecated; use `import {fetch} from 'frontend-js-web';` instead",
+			'no-metal-anim':
+				'metal-anim is deprecated; use native `window.requestAnimationFrame()` instead',
 			'no-metal-aop':
 				"metal-aop is deprecated; use `import {AOP} from 'frontend-js-web';` instead",
+			'no-metal-assertions':
+				'metal-assertions is deprecated; use native `typeof` instead',
 			'no-metal-clipboard':
 				'metal-clipboard is deprecated; use ClipboardJS instead',
 			'no-metal-debounce':
 				"metal-debounce is deprecated; use `import {debounce} from 'frontend-js-web';` instead",
+			'no-metal-dom':
+				'metal-dom is deprecated; see https://issues.liferay.com/browse/LPS-122383 to learn what to use instead',
+			'no-metal-key':
+				'metal-key is deprecated; use native key event handling instead',
 			'no-metal-keyboard-focus':
-				"metal-keyboard-focus is deprecated; use `import {debounce} from 'frontend-js-web';` instead",
+				'metal-keyboard-focus is deprecated; use native focus behaviour instead',
+			'no-metal-multimap':
+				'metal-multimap is deprecated; use native Set instead',
+			'no-metal-path-parser':
+				'metal-path-parser is deprecated; use native string manipulation instead',
+			'no-metal-position':
+				"metal-position is deprecated; use `import {ALIGN_POSITIONS, align} from 'frontend-js-web';` instead",
 			'no-metal-promise':
 				'metal-promise is deprecated; use native Promise instead',
+			'no-metal-scrollspy':
+				'metal-scrollspy is deprecated; use native scroll instead',
 			'no-metal-storage':
 				'metal-storage is deprecated; use a local implementation instead',
 			'no-metal-structs':
 				'metal-structs is deprecated; use plain objects instead',
+			'no-metal-throttle':
+				"metal-throttle is deprecated; `import {throttle} from 'frontend-js-web';` instead",
 			'no-metal-uri': 'metal-uri is deprecated; use native URL instead',
 			'no-metal-useragent':
 				'metal-useragent is deprecated; use `Liferay.Browser` instead',
--- a/projects/eslint-config/plugins/portal/lib/rules/no-metal-plugins.js
+++ b/projects/eslint-config/plugins/portal/lib/rules/no-metal-plugins.js
@@ -12,18 +12,23 @@ const DEPRECATED_MODULES = [
 	'metal-clipboard',
 	'metal-debounce',
 	'metal-dom',
+	'metal-isomorphic',
 	'metal-key',
 	'metal-keyboard-focus',
 	'metal-multimap',
+	'metal-pagination',
 	'metal-path-parser',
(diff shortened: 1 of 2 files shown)


projects/eslint-config/plugins/portal/lib/rules/no-metal-plugins.js
```diff
@@ -44,22 +54,42 @@ module.exports = {
 		},
 		fixable: null,
 		messages: {
+			'no-metal-affix':
+				'metal-affix is deprecated; use native offset instead',
 			'no-metal-ajax':
 				"metal-ajax is deprecated; use `import {fetch} from 'frontend-js-web';` instead",
+			'no-metal-anim':
+				'metal-anim is deprecated; use native `window.requestAnimationFrame()` instead',
```
wincent, line 62: You could drop the "native" and the (implied) `window.` from here. We have some interest in keeping the ESLint error messages relatively narrow because they get spewed out to the console, or inlined in IDEs etc.

I'd say the same for all the ones that say "native" (I know there were already a couple of pre-existing ones but you should feel free to change those too).

projects/eslint-config/plugins/portal/lib/rules/no-metal-plugins.js
```diff
@@ -44,22 +54,42 @@ module.exports = {
...
 			'no-metal-aop':
 				"metal-aop is deprecated; use `import {AOP} from 'frontend-js-web';` instead",
+			'no-metal-assertions':
+				'metal-assertions is deprecated; use native `typeof` instead',
 			'no-metal-clipboard':
 				'metal-clipboard is deprecated; use ClipboardJS instead',
 			'no-metal-debounce':
 				"metal-debounce is deprecated; use `import {debounce} from 'frontend-js-web';` instead",
+			'no-metal-dom':
+				'metal-dom is deprecated; see https://issues.liferay.com/browse/LPS-122383 to learn what to use instead',
```
wincent, line 72: Continuing the theme, delete "to learn what to use instead".

projects/eslint-config/plugins/portal/tests/lib/rules/no-metal-plugins.js
```diff
@@ -17,6 +17,15 @@ const ruleTester = new MultiTester(parserOptions);
 
 ruleTester.run('no-metal-plugins', rule, {
 	invalid: [
+		{
+			code: "import {Affix} from 'metal-affix';",
```
wincent, line 21: While we're at it, I'd make all of these consistent because we currently have some that use the default export and some that use the named export; just go for the default export, avoiding the extra width and punctuation of the curlies.

projects/eslint-config/plugins/portal/tests/lib/rules/no-metal-plugins.js
```diff
@@ -26,6 +35,15 @@ ruleTester.run('no-metal-plugins', rule, {
 				},
 			],
 		},
+		{
+			code: "import {Anim} from 'metal-anim';",
```
wincent, line 39: Case in point: `metal-anim` [doesn't even have a named export](https://github.com/metal/metal-plugins/blob/97ae8683fabd45efadbf12a966fd9d57fcb9073f/packages/metal-anim/src/Anim.js#L97).

projects/eslint-config/plugins/portal/tests/lib/rules/no-metal-plugins.js
```diff
@@ -62,6 +98,24 @@ ruleTester.run('no-metal-plugins', rule, {
 				},
 			],
 		},
+		{
+			code: "import {MultiMap} from 'metal-multimap';",
```
wincent, line 102: Another package that [doesn't even have this named export](https://github.com/metal/metal-multimap/blob/f1149882dbfda3c1d082a61a2cc756ac1c85a2af/src/MultiMap.js#L141).

projects/eslint-config/plugins/portal/lib/rules/no-metal-plugins.js
```diff
@@ -44,22 +54,42 @@ module.exports = {
...
+			'no-metal-key':
+				'metal-key is deprecated; use native key event handling instead',
 			'no-metal-keyboard-focus':
-				"metal-keyboard-focus is deprecated; use `import {debounce} from 'frontend-js-web';` instead",
+				'metal-keyboard-focus is deprecated; use native focus behaviour instead',
+			'no-metal-multimap':
+				'metal-multimap is deprecated; use native Set instead',
```
wincent, line 78: This seems wrong. A `Set` isn't an appropriate replacement (it only holds values); I think you mean `Map` (which holds key/value pairs).

wincent (commented):
> * added `affix`, `position`, and `throttle` and others to the list of deprecated modules. Those are leftover modules but they're still used

I don't think we can add them to the list if they are still used. We don't want to have to add lint suppressions. The idea is to add them to the list _after_ the usages are gone, to prevent new usages from creeping back in.

> * It wasn't clear to me if I'm supposed to go to do what @wincent mentioned in [#386 (comment)](https://github.com/liferay/liferay-frontend-projects/issues/386#issuecomment-766996855) - the 7.2 and 7.3 config update

Yes, you should do it.

> * `metal-isomorphic` is like a testing library it seems, so not sure what to do about that one
> * `metal-pagination` is a component with UI, so not sure what to do about that one
> * `metal-router` is a Metal variant of the React Router, do we recommend people to use that one, or something else, not sure
> * `metal-toggler` is a component with UI, same as pagination
> * `metal-web-component` is a specific case too, not sure what to do with it either

For all these ones, if they aren't being used and a replacement isn't clear, I think it's fine to just say it's deprecated and include a link to the LPS or some other suitable place (even this PR).

I only have nits inline which should be quick to change. I haven't checked that all the packages you added here actually _are_ unused in `master`; we're relying on you to do that.

PS. I am kind of horrified to learn from this PR that we have [a `metal-multimap` package](https://npmjs.com/package/metal-multimap) that lives in [its own repo](https://github.com/metal/metal-multimap), in addition to [the `Multimap` implementation already provided](https://github.com/metal/metal-plugins/blob/master/packages/metal-structs/src/MultiMap.js) by [the `metal-structs` package](https://www.npmjs.com/package/metal-structs), that lives in [a different repo](https://github.com/metal/metal-plugins) (not to be confused with [yet another repo](https://github.com/metal/metal.js) that holds still more packages).

wincent:
Ah, forgot to mention, before merging this, let's pick a PR title that describes _what_ was done (seeing as PR titles are used to generate the changelog, and a title like "feat: Update the no-metal-plugin eslint rule" doesn't really tell anybody what actually happened).

It should probably be something like `feat(eslint-config)!: add several packages to no-metal-plugin rule` (the `eslint-config` scope just to be a good monorepo citizen, and the `!` to indicate that this is actually a breaking change, because updating the config in a repo may cause new lint errors to appear).

kresimir-coko:
Thanks @wincent, great comments. 

I think I have a typo in the sentence you mentioned - `Those are leftover modules but they're still used` should be `There are leftover modules but they're still used`, completely changes the meaning of the sentence, huh 🤦 . I'll edit it now.

I'll tackle this today.

kresimir-coko:
@wincent I've pushed a new commit, addressing all of your comments and double-checking that all of the listed packages are **not used on DXP master**

wincent (approved):
Cool.

PR #460, opened 2021-03-18
chore(monorepo): temporarily fix lockfile failure in CI

wincent:
[CI is failing](https://github.com/liferay/liferay-frontend-projects/runs/2140506487) on master (cfb2c217c449f717e473) right now because we just prepared the v23.0.0 release of `@liferay/eslint-config`, but `@liferay/npm-scripts` still depends on v22.0.0.

Now, @kresimir-coko will fix that tomorrow when he completes the release and updates the dependency, so for now I am just freshening the lockfile to keep CI green. (I could also fix this by updating the dependency myself, but that would require me to finish the release, and I don't want to do that.)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1402,6 +1402,21 @@
     "@types/yargs" "^15.0.0"
     chalk "^4.0.0"
 
+"@liferay/eslint-config@22.0.0":
+  version "22.0.0"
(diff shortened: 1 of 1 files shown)


wincent:
Trivial and we want `master` green, so merging this.

PR #462, opened 2021-03-18
feat(changelog-generator): suggest when version bump should be major

wincent:
Motivated by [today's Slack thread](https://liferay.slack.com/archives/C3JBR21HA/p1616068068087200), this should help us follow SemVer a little more consistently, even if you don't know that we follow SemVer in our projects.

This doesn't handle every single case in which a version bump should be major, but it does handle the common case of running in `--interactive` mode and then letting the user choose from the list. Basically, if any of the changes are breaking, it will print a warning that says:

    Breaking changes detected; "major" or "premajor" version recommended.

We don't force it, but we strongly hint that the user should do it.

What cases doesn't this handle? Well, if you're not running in `--interactive` mode, or you explicitly supply a spelled-out version number, we don't check that number.

In an ideal world, we'd be 100% robust and catch every single little edge case, but I am looking for a quick win here and don't want to do a huge refactor. As it is, the design is starting to look a bit thin and stretched at this point; note how in order to make this work I had to move things around a fair bit anyway:

The `formatChanges()` function is the only place we actually look at the text of the changes, and therefore the only place where we are in a position to know that changes are actually breaking. `formatChanges()` returns a string to its caller (`generate()`), and `generate()` in turn returns a string to its own callers (of which there are two).

So, the minor refactoring here is to split `formatChanges()` into two functions:

- `parseChanges()`, which now returns the parsed sections of the changelog as a Map instead of a formatted string. At this point, we have enough information to detect the presence of breaking changes.
- `formatSections()`, which turns the section Map into a string.

Given that `generate()` calls both of those, `generate()` can now return a tuple (a chunk of text plus a `breaking` flag) instead of just a string, and we can use that to show the appropriate warning.

So, I think this is going to work, but as I said, the design is starting to show signs of strain. This originally started off as a cute little script written in a functional style, and we have layered on more and more features. It is now approaching the point where it could use a moderate (but not excessive) dose of OO-ification; most of the logic would still be pure functions, but we'd have an object that we could accumulate some working state in and avoid having to pass these odd data types (tuples and such) through multiple layers. We could also pass a mutable state object around, but at that point you're losing most of the benefit of the functional style. Maybe the next big change we make, if we ever do, will be the one that pushes us over the edge to do this.

Test plan: Merge some fake merge commits and run generator.

In the absence of breaking changes we see:

    [--interactive] Preview of changes 👀
    ## [npm-scripts/v0.0.0-placeholder.0](https://github.com/liferay/liferay-frontend-projects/tree/npm-scripts/v0.0.0-placeholder.0) (2021-03-22)

    [Full changelog](https://github.com/liferay/liferay-frontend-projects/compare/npm-scripts/v39.1.0...npm-scripts/v0.0.0-placeholder.0)

    ### :new: Features

    -   feat: some stuff ([\#4614](https://github.com/liferay/liferay-frontend-projects/pull/4614))

    [--interactive] Would write ./CHANGELOG.md 86704 bytes ✨

    1. major
    2. minor
    3. patch
    4. premajor
    5. preminor
    6. prepatch
    7. prerelease

    Please choose a version from the list, or provide one in full (or enter to abort):

With breaking changes, we see:

    [--interactive] Preview of changes 👀
    ## [npm-scripts/v0.0.0-placeholder.0](https://github.com/liferay/liferay-frontend-projects/tree/npm-scripts/v0.0.0-placeholder.0) (2021-03-22)

    [Full changelog](https://github.com/liferay/liferay-frontend-projects/compare/npm-scripts/v39.1.0...npm-scripts/v0.0.0-placeholder.0)

    ### :boom: Breaking changes

    -   feat!: broke thing ([\#4615](https://github.com/liferay/liferay-frontend-projects/pull/4615))

    ### :new: Features

    -   feat!: broke thing ([\#4615](https://github.com/liferay/liferay-frontend-projects/pull/4615))
    -   feat: some stuff ([\#4614](https://github.com/liferay/liferay-frontend-projects/pull/4614))

    [--interactive] Would write ./CHANGELOG.md 86930 bytes ✨
    warning: Breaking changes detected; "major" or "premajor" version recommended.

    1. major
    2. minor
    3. patch
    4. premajor
    5. preminor
    6. prepatch
    7. prerelease

    Please choose a version from the list, or provide one in full (or enter to abort):

(Note: in the terminal this is colorized, so hopefully the warning stands out enough.)

### DIFF
--- a/projects/npm-tools/packages/changelog-generator/src/index.js
+++ b/projects/npm-tools/packages/changelog-generator/src/index.js
@@ -33,6 +33,14 @@ function escape(string) {
 		.replace(/>/g, '&gt;');
 }
 
+function formatDate(date) {
+	const year = date.getFullYear();
+	const month = (date.getMonth() + 1).toString().padStart(2, '0');
+	const day = date.getDate().toString().padStart(2, '0');
+
+	return `${year}-${month}-${day}`;
+}
+
 /**
  * Conventional commit types, and equivalent human-readable labels, in the order
  * that we wish them to appear in the changelog.
@@ -64,45 +72,7 @@ const TYPES = {
 	/* eslint-enable sort-keys */
 };
 
-/**
- * Aliases mapping common mistakes to legit Conventional Commits types.
- */
-const ALIASES = {
-	bug: 'fix',
-	doc: 'docs',
-	feature: 'feat',
-};
-
-const TYPE_REGEXP = /^\s*(\w+)(\([^)]+\))?(!)?:\s+.+/;
-
-const BREAKING_TRAILER_REGEXP = /^BREAKING[ -]CHANGE:/m;
-
-async function formatChanges(changes, remote) {
-	const sections = new Map(Object.keys(TYPES).map((type) => [type, []]));
-
-	changes.forEach(({description, number}) => {
-		const match = description.match(TYPE_REGEXP);
-
-		const type = ALIASES[match && match[1]] || (match && match[1]);
-
-		const section = sections.get(type) || sections.get('misc');
-
-		const link = linkToPullRequest(number, remote);
-
-		const entry = link
-			? `-   ${escape(description)} (${link})`
-			: `-   ${escape(description)}`;
-
-		section.push(entry);
-
-		const breaking =
-			(match && match[3]) || BREAKING_TRAILER_REGEXP.test(description);
-
-		if (breaking) {
-			sections.get('breaking').push(entry);
-		}
-	});
-
+function formatSections(sections) {
 	return Array.from(sections.entries())
 		.map(([type, entries]) => {
 			if (entries.length) {
@@ -113,14 +83,6 @@ async function formatChanges(changes, remote) {
 		.join('\n\n');
 }
 
-function formatDate(date) {
-	const year = date.getFullYear();
-	const month = (date.getMonth() + 1).toString().padStart(2, '0');
-	const day = date.getDate().toString().padStart(2, '0');
-
-	return `${year}-${month}-${day}`;
-}
-
 async function generate({date, from, remote, to, version}) {
 	const changes = await getChanges(from, to);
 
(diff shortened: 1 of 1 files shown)


wincent:
cc @kresimir-coko

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -519,6 +487,12 @@ async function go(options) {
 	if (options.interactive && options.version.endsWith(PLACEHOLDER_VERSION)) {
 		info(`[--interactive] Would write ${outfile} ${written} bytes ✨`);
 
+		if (breaking) {
+			warning(
+				'Breaking changes detected; "major" or "premajor" version recommended.'
+			);
+		}
```
wincent, line 494: Lots of things moving around in the PR, but these lines are the motivation.

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -437,7 +405,9 @@ async function go(options) {
 	const remote = await getRemote(options);
 	let from = await getPreviousReleasedVersion(to, tagPattern);
 	const date = await getDate(to);
-	let contents = await generate({
+
+	/* eslint-disable-next-line prefer-const */
```
wincent, line 409: Sigh... one reason why I don't like "abusing" tuples. We want one of these (`contents`) to be reassignable but not the other (`breaking`) so ESLint will complain unless we suppress it.

wincent:
Taking this out of draft now because I've actually tested it now (see updated "Test plan:" above). Also rebased so we can get a green CI run (via [471edcc9e9](https://github.com/liferay/liferay-frontend-projects/commit/471edcc9e935a79a9d18f77ab38246ab6847c832)).

wincent:
I figure this has been here long enough that if anybody had anything to say about it they probably would have by now so I am just going to merge it.

PR #465, opened 2021-03-22
fix(js-toolkit): don't fail in css-loader for scoped dependencies

izaera:
This is needed for [LPS-129367](https://issues.liferay.com/browse/LPS-129367), otherwise the portal does not build.

I will need to release a new version of the toolkit and npm-scripts after this is merged.


izaera:
If nobody has anything to say I'll merge this as it is a fix for a bug (not a feature)...

wincent (approved):
> If nobody has anything to say I'll merge this as it is a fix for a bug (not a feature)...

Seems to be in the right language.

Issue #450, opened 2021-03-15
Update to Babel 7 for JS Generator


jbalsas:
### Issue type

-   [x] :gift: Feature request

### Description

**Desired behavior:**
The boilerplate generated by `yo liferay-js` when generating a React portlet by default allows usage of newer React syntax to be used and transpiled.

Based on https://github.com/ethib137/liferay-react-example which is a description of what needs to be done to update the necessary dependencies on a newly generated project, **steps 5,6 & 7 should be unnecessary**.

**Current behavior:**
The default set of dependencies is outdated and needs of manual steps to bring it up to date.

jonmak08:
Adding https://issues.liferay.com/browse/IFI-2390 for reference.

jonmak08:
Resolved and published with release https://github.com/liferay/liferay-frontend-projects/tree/liferay-js-toolkit/v2.24.3

jbalsas:
Installation notes have been updated by @ethib137 at https://github.com/ethib137/liferay-react-example/commit/6f255b35051c902880f18473ab30bb52a7e55e1c. I assume they work... 😅 

Thanks @ethib137 for taking care of that! 🙇

ethib137:
It's all working as expected! :) Thanks for the update everyone!

PR #466, opened 2021-03-23
feat(js-toolkit): Update to Babel 7 for JS Generator (#450)

jonmak08:
Hey @liferay,

Attached is an update for https://github.com/liferay/liferay-frontend-projects/issues/450.

Please let me know if you have any questions. Thanks!

### DIFF
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-vanilla-portlet/vanilla-portlet.js
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-vanilla-portlet/vanilla-portlet.js
@@ -55,8 +55,9 @@ export default class extends Generator {
 		// Configure build
 
 		if (useBabel) {
-			pkgJson.addDevDependency('babel-cli', '6.26.0');
-			pkgJson.addDevDependency('babel-preset-env', '1.7.0');
+			pkgJson.addDevDependency('@babel/cli', '^7.7.5');
+			pkgJson.addDevDependency('@babel/core', '^7.7.5');
+			pkgJson.addDevDependency('@babel/preset-env', '^7.7.6');
 			pkgJson.addBuildStep('babel --source-maps -d build src');
 			cp.copyFile('.babelrc');
 		}
@@ -68,7 +69,6 @@ export default class extends Generator {
 		// Configure webpack
 
 		if (useBabel) {
-			pkgJson.addDevDependency('babel-loader', '7.1.5');
 			npmbuildrc.addWebpackRule(/src\/.*\.js$/, 'babel-loader');
 		}
 
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-react-portlet/dependencies.json
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-react-portlet/dependencies.json
@@ -1,8 +1,9 @@
 {
 	"devDependencies": {
-		"babel-cli": "6.26.0",
-		"babel-preset-env": "1.7.0",
-		"babel-preset-react": "6.24.1"
+		"@babel/cli": "^7.7.5",
+		"@babel/core": "^7.7.5",
+		"@babel/preset-env": "^7.7.6",
+		"@babel/preset-react": "^7.7.4"
 	},
 	"dependencies": {
 		"react": "16.8.6",
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-react-portlet/react-portlet.js
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-react-portlet/react-portlet.js
@@ -51,7 +51,6 @@ export default class extends Generator {
 
 		// Configure webpack
 
-		pkgJson.addDevDependency('babel-loader', '7.1.5');
 		npmbuildrc.addWebpackRule(/src\/.*\.js$/, 'babel-loader');
 
 		// Prepare text labels
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-react-portlet/templates/.babelrc.ejs
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-react-portlet/templates/.babelrc.ejs
@@ -1,3 +1,3 @@
 {
-	"presets": ["env", "react"]
+	"presets": ["@babel/preset-env", "@babel/preset-react"]
 }
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-shared-bundle/shared-bundle.js
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-shared-bundle/shared-bundle.js
@@ -55,8 +55,9 @@ export default class extends Generator {
 
 		// Configure build
 
-		pkgJson.addDevDependency('babel-cli', '6.26.0');
-		pkgJson.addDevDependency('babel-preset-env', '1.7.0');
+		pkgJson.addDevDependency('@babel/cli', '^7.7.5');
+		pkgJson.addDevDependency('@babel/core', '^7.7.5');
+		pkgJson.addDevDependency('@babel/preset-env', '^7.7.6');
 		pkgJson.addBuildStep('babel --source-maps -d build src');
 		cp.copyFile('.babelrc');
 
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-shared-bundle/templates/.babelrc.ejs
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-shared-bundle/templates/.babelrc.ejs
@@ -1,3 +1,3 @@
 {
-	"presets": ["env"]
+	"presets": ["@babel/preset-env"]
 }
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-vanilla-portlet/templates/.babelrc.ejs
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-vanilla-portlet/templates/.babelrc.ejs
@@ -1,3 +1,3 @@
 {
-	"presets": ["env"]
+	"presets": ["@babel/preset-env"]
 }


izaera:
I'm not sure if this is going to work, because updating Babel from v6 to v7 is a big change. I think it may break some plugins...

So, if we really need to upgrade to Babel 7 we may need to check everything thoroughly :thinking:

izaera:
The problem is that we use babel under the hood in the liferay-npm-bundler, so I'm not sure what would happen if we have Babel 7 and Babel 6 in the same project (in the generated projects I mean).

izaera:
Oh, one more thing, the PR is failing the Semantic Pull Request check. This is because of the commit's title, that has an space between `fix` and `(` (I think).

In any case, I recommend visiting the [guidelines](https://github.com/liferay/liferay-frontend-projects/tree/master/guidelines) repo, and specifically the [commit_messages.md](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/commit_messages.md) and [pull_requests.md](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/pull_requests.md) documents to learn what is expected.

wincent:
I am a little confused by the title...

> update React JS Generator to latest React version to be able to use hooks

...because I don't actually see any change to the React version in here?

confused-scratching-head

jonmak08:
Hey @izaera,

For testing the change, are there specific tests that you run or steps that you follow to test everything?  I generated a project with the Babel 6 then I updated my package locally to Babel 7 and redeployed the package and everything seemed to work fine.  I am not sure if there is a better way to test the changes or what is expected of the plugins, etc.?

jonmak08:
Hey @wincent,

Yes the changes are just upgrading to Babel 7.  I just took the title from the issue that was filed (https://github.com/liferay/liferay-frontend-projects/issues/450).

wincent:
> Yes the changes are just upgrading to Babel 7. I just took the title from the issue that was filed (#450).

I still don't understand, because it's not just the title; it's the entire description of that issue. What does Babel (a JS transpiler that runs at build-time) have to do with "being able to use hooks"? [React 16.8.0](https://reactjs.org/blog/2019/02/06/react-v16.8.0.html) is "the one with hooks", and we were already using that. Maybe the title and the description are just wrong. @jbalsas: can you clarify what exactly was meant?

jbalsas:
Sorry, probably my fault... I quickly drafted the issue based on a direct request by Brian that the steps in https://github.com/ethib137/liferay-react-example/blob/f7d032ff251d858b251abb0586ab20459d321c3c/README.md could be simplified. In there it stated that this was necessary and I didn't think it through :(

I see @ethib137 updated the instructions yesterday, so the goal here is to figure out why/if [Updating Dependencies](https://github.com/ethib137/liferay-react-example/blob/master/README.md#updating-dependencies) is in fact needed and make that step unnecessary if that's the case.

@ethib137, could you please add some context here for us to better understand the issue? 🙏

ethib137:
Hey @jbalsas, It looks like this is a mistake on my part. Sorry :(. I had thrown together these instructions very quickly and in the process misremembered an issue I'd had in the past. You are correct, the update of the babel dependencies is not required in order to use React hooks. I had needed it as a result of using "spread syntax" and had conflated the issues. Sorry for the confusion.

I still would love to have Babel updated in order to be able to use some of these features, but this is obviously a much lower priority for this reason.

jonmak08:
If the goal of this is to allow the spread operator to be transpiled, why not just add the plugin `babel-plugin-transform-object-rest-spread`?  I only say this because updating to Babel 7 might have other side effects as stated by @izaera.

I have updated this pull request to add that plugin to review if that would be a better solution for the time being and possibly if we release a new version of the js-toolkit, we can update to Babel 7 at that point.

Thoughts @wincent, @jbalsas, @izaera?

ethib137:
Hey @jonmak08, my reason for upgrading to Babel 7 instead of using `babel-plugin-transform-object-rest-spread` is because it was easier to upgrade to Babel 7. As far as I can tell, Babel's documentation has removed everything related to version 6 except for upgrade instructions. As a result if I search for `babel-plugin-transform-object-rest-spread` I don't find it in Babel's docs. For me the only obvious thing to do was to follow the docs which required upgrading to the latest version. 

Staying on Babel 6 creates more friction for developers. From my experience, when you search for how to do something in Babel, the primary answers you will get are related to version 7. If we upgrade to version 7 we will be able to improve the developer experience. Are we aware of any issues with upgrading to 7? Personally, I haven't found any yet.

wincent:
I get both viewpoints: the conservative update (with just the plugin) seems like a reasonable default position that minimizes risk. The desire to move onto the latest because the Babel docs are incomplete is also totally understandable (the v7 docs themselves are pretty bad, but finding info on v6 is harder).

> The problem is that we use babel under the hood in the liferay-npm-bundler, so I'm not sure what would happen if we have Babel 7 and Babel 6 in the same project (in the generated projects I mean).

FWIW, it's probably going to be fine.

In `liferay-portal` right now, for example, we have `babel-core` 6.2.3 because `liferay-npm-bundler` depends on it. But we also have `@babel/core` (7.8.4) because `@liferay/npm-scripts` depends on it. Seems they can co-exist just fine because the operations involved are different (ie. it's not two versions of Babel being loading in the same process; it's a different version being loaded by two different processes).

So, I personally would just go with the big update, but I'm also fine with doing the minimal one. Either one is an improvement over the current state.

izaera:
I don't have any problem with updating to Babel 7. In fact I would love to see that change because people has been asking for it since more than one year ago.

The only caveat is that we need to do a full regression of all the features to make sure we don't break anything (if we want to do it The Right Way :tm:) and there's no QA for this, it's all manual. Said that it's not a huge task (I would expect from 1 day to no more than 1 week in case anything fails), but it needs to be done.

Then, we have two places where we use babel: in the project itself (that's the user's side) and inside the bundler, to transform the code to AMD. 

I would say updating the user side (in the generator) shouldn't matter because we are just generating a template but, once it's generated, it's up to the user to do whatever he wants with it. So we simply begin to generate projects for babel 7 at some point in the version 2 series (of the bundler) and all past projects remain in babel 6 until the users update them.

On our side (bundler) we use 1 or 2 plugins, so it should be easy too. Probably just updating package.jsons...

What worries me most is the two scenarios where you have:

1. A new project (babel 7) which you need to build with and "old" bundler (babel 6). This is very strange, but it could happen for a short time if we need to downgrade someone because of a support ticket, or something like that.
2. An old project (babel 6) which upgrades to a "new" bundler (babel 7).

If we can make sure both work (especially 2, which is the most probable and will happen for a long time until everybody migrates; if they do...) then it's OK.

Then we would also need to test all adaptations to make sure there are no strange interactions. I don't expect this to happen but who knows...

I think it's only a matter or prioritization. If we have the resources to do it (say 1-2 man-week; very rough estimate, just the order of magnitude) then let's do it.

ethib137:
Hey guys, is there any update on this? 

Are there currently the resources available / is this worth prioritizing?

jbalsas:
Hey @ethib137, we asked internally for help QA'ing these changes since we're afraid of possible side-effects on existing projects/customers. 

Since we're in the middle of the first 7.4 GA release, it might take our QA team a bit longer to get to this.

/cc @manoelcyreno, @john-co

maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-vanilla-portlet/vanilla-portlet.js
```diff
@@ -55,8 +55,9 @@ export default class extends Generator {
 		// Configure build
 
 		if (useBabel) {
-			pkgJson.addDevDependency('babel-cli', '6.26.0');
-			pkgJson.addDevDependency('babel-preset-env', '1.7.0');
+			pkgJson.addDevDependency('@babel/cli', '^7.7.5');
```
izaera, line 58: @jonmak08 Why do you need to change the vanilla JS portlet too? Isn't it enough with changing the react one? :thinking:

izaera: And the same question for the shared bundle target...

jonmak08: I thought if we were going to upgrade babel, we wanted to upgrade it everywhere in the toolkit, but if not, we can just upgrade it in only the react portlet templates.

jbalsas: I think the main concern @izaera had was with the bundler that is invoked as part of the build script resulting from the generator:

```json
"build": "npm run copy-assets && liferay-npm-bundler"
```

In its latest released version (I think) it [still uses babel 6](https://github.com/liferay/liferay-js-toolkit/blob/master/packages/liferay-npm-bundler/package.json#L18) and it's this combination the one that both, has us worried that it might not work as expected and that changing it there might cause issues elsewhere in the toolchain.

izaera: Well, we may change it here too, test it, and see what happens. As long as we pass it through QA and it works...

@wincent [said](https://github.com/liferay/liferay-frontend-projects/pull/466#issuecomment-806450947) we already mix babel 7 and 6 in liferay-portal and nothing happens, so I guess it is safer than it seems.

My previous comments may be a bit misleading because I was assuming (without being too much aware of it) that we needed to change babel inside the bundler too but I realized yesterday (due to a comment of @jbalsas in slack) that this change only affects user's projects, so it is very possible that some people out there may be already using babel 7 to transpile and the bundler to deploy. We don't know, but I wouldn't find it strange given that nobody is complaining about babel 6 since long ago (note that another reason could be that nobody is using the toolkit too :sweat_smile: ).

So I would say, let's change it, test it to make sure nothing breaks and release.

@john-co is having problems setting the old QA code because it broke when we migrated to liferay-frontend-projects, so I will have a look at it today and see what can be done to fix/change it.

jbalsas: > [...] so it is very possible that some people out there may be already using babel 7 to transpile and the bundler to deploy [...]

We know @ethib137 was, that's how we got here 😉

ethib137: I use it all the time and tell everyone to use it. 😉

john-co:
Update: I was able to successfully deploy javascript and react widgets using the js-toolkit. However, I had issues with shared-bundle. I got the same issue without the babel 7 update so likely there's something I'm missing. 

Would like to confirm with @izaera if there's anything else needed to properly build/test the shared bundle portlet as mentioned in slack channel: https://liferay.slack.com/archives/C3JBR21HA/p1617840728305500?thread_ts=1617806445.294000&cid=C3JBR21HA

izaera:
That error message is no longer printed in master (don't know for 7.3). It used to appear when you deployed any toolkit-made OSGi bundle that had no JS files inside. However this is clearly not an error (it could be a warning), because you may be deploying other things. You can make it disappear by adding a simple .js file to your shared bundle (even an empty one would do, I think), but I would say it is harmless. And in any case, if it was working wrong it would be an error in the server, not the bundler, so I would say we are safe...

john-co:
Thanks @izaera ! Then this looks good on QA side. :heavy_check_mark:

jonmak08:
Hey @izaera,

I am trying to make the updates to the `CHANGELOG.md` based on what is stated in https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/CONTRIBUTING.md#releasing-new-versions, but it does not seem to pick up my changes for this pull request even if I merge the changes in.  Am I missing something?

Please let me know if you have any questions. Thanks!

izaera:
@jonmak08 :thinking: But it is not merged yet... Could it be that the changelog generator needs the PR to be merged here, in GitHub?

Feel free to merge this, BTW. The usual review process we follow is that when someone gives the OK, things can be merged. This has been approved by QA and everybody else in the thread so it's OK and you can go ahead :muscle:

jonmak08:
Closing pull request as changes have been merged at https://github.com/liferay/liferay-frontend-projects/commit/17add93153a9a7bb9492d7bb8b44ec5b08929796.

wincent:
> Could it be that the changelog generator needs the PR to be merged here, in GitHub?

That's right. The generator looks at the merges on the disk in the repo; it doesn't connect to the GitHub API at all. So, merge on GitHub, `pull` in your local repo, and then changelog generator will be able to "see" the merge(s).

wincent:
Having said that, it won't "see" 17add93153a9a7bb9492d7bb8b44ec5b08929796 because it looks like it was pushed directly to master without a merge:

Screenshot 2021-04-14 -103002-hlmprv5B@2x

How this works is explained in [the changelog generator docs](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/packages/changelog-generator/README.md):

> The generated changelog is based on the merge commits produced by GitHub, which means that the quality of the changelog depends on the quality of the pull request titles (please see [the recommendations](https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/commit_messages.md) for producing high quality commit messages and PR titles in the [liferay-frontend-guidelines](https://github.com/liferay/liferay-frontend-guidelines) repo). Additionally, if you create merge commits by hand rather than using GitHub, or if you push changes without creating merges, the results may be suboptimal.

So it is a "dumb" tool that will do a good job only if we give it well-formed input to consume.

PR #472, opened 2021-03-26
feat(npm-scripts): add @types/react-dom dependency

wincent:
This allows us to use TS in React components, as seen in PRs like:

Note that `npm-scripts` may not be the definitive home for this; it's just one of the two places that we can put it:

1.  Inside `frontend-js-react-web`, because that's where our `react` dependency comes from, and it makes sense to version the React package and its type definitions together (better still would be if the React team shipped TypeScript types with the package).
2.  Inside `npm-scripts` because that is "The Blessed" way of getting devDependencies into liferay-portal; in fact, the only reason we get away with it in `frontend-js-react-web` is due to [a "temporary suppression"](https://github.com/liferay/liferay-portal/blob/a9386e72048dbab268d97ca5fd295449579ad0a6/source-formatter-suppressions.xml#L157-L159) that has been in the repo [since August 2019](https://github.com/liferay/liferay-portal/commit/0f14d0806adae5f425ed38e2da8913bd7c93d9b0).

So, this commit goes with "2". Once [the LPS](https://issues.liferay.com/browse/LPS-99901) cited in the above commit is resolved... whenever that is... we can reconsider things.

**Test plan:**

1. In the above PR, `yarn remove @types/react-dom` inside `frontend-js-react-web`.
2. In `modules/`, do `yarn add path/to/npm-scripts`.
3. Verify that `yarn run build` still works without errors inside `frontend-js-react-web`.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/package.json
+++ b/projects/npm-tools/packages/npm-scripts/package.json
@@ -23,6 +23,7 @@
 		"@testing-library/react-hooks": "^3.4.2",
 		"@testing-library/user-event": "^4.2.4",
 		"@types/jest": "^26.0.14",
+		"@types/react-dom": "17.0.3",
 		"@types/vfile-message": "2.0.0",
 		"babel-eslint": "^10.0.3",
 		"babel-jest": "^25.1.0",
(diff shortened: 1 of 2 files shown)


wincent:

ninja-merge-3

Am I doing it right?

PR #473, opened 2021-03-26
feat(npm-scripts): automatically run TS tests

wincent:
In PRs like:

We have had to explicitly adjust the `testMatch` directive to catch TS files. We may as well do this automatically throughout liferay-portal.

Seeing as I was changing this, note that I made the pattern a little stricter (adding `<rootDir>`) because this brings us into more exact alignment with our conventions (which clearly state that source should go in `src/` and tests should go in `test/`).

**Note:** When we integrate this with liferay-portal, we not only have to clean out the `jest.config.js` that exists in those two PRs, but also the `jest` field in the `package.json` of `remote-app-client-js`, which is mentioned below.

**Test plan:**

1. Run `yarn run test` in modules using TypeScript as a first spot check (`remote-app-client-js`, plus the two modules in the PRs above). Note that tests run correctly even when you delete the local `jest.config.js` overrides.

2. Run `portool test` from `modules/` (this is just a script that enables me to run `yarn run test` in every project that has tests) and see them all pass.


wincent:

ninja-merge-4

Am I doing it right?

PR #478, opened 2021-03-30
feat(npm-scripts): manage TS type artifacts

wincent:
~This is still a WIP so parking it here as a draft.~

- [x] Teach `build` subcommand to emit TS type artifacts.
- [x] Add `types` subcommand to do a global build of all TS type artifacts across `liferay-portal`.
- [x] Teach `check` to confirm that TS type artifacts are up-to-date when running `ci:test:sf`.
- [x] Auto-generate `tsconfig.json` files.
- [x] Plug auto-generation into `build` subcommand.
- [x] Plug auto-generation into `check` subcommand.
- [x] Write human-facing docs of how this works.

~So tomorrow I'm going to look at that last one. The pieces I've already added to this PR should make it fairly straightforward: we have a function that returns us info about the dependency graph between TypeScript projects, including their name, their location on disk, what their `main` field points at etc, so we can use that to fill out the required `paths` and `references` section in the `tsconfig.json`.~

~My testing indicates that we can add extra fields to that file without `tsc` complaining, so we'll leverage that to put in some `__generated__` property and an optional `__overrides__` one which can be used in those expected rare cases (like `remote-app-client-js`) where the vanilla config won't work. So far, I think that all the other use cases are going to be just fine with the vanilla set-up. 🤞~

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/README.md
+++ b/projects/npm-tools/packages/npm-scripts/README.md
@@ -165,6 +165,14 @@ liferay-npm-scripts theme build
 
 Runs the "build" task, providing it with the configuration it needs to find core dependencies such as the [`liferay-frontend-theme-styled` base theme files](https://github.com/liferay/liferay-portal/tree/master/modules/apps/frontend-theme/frontend-theme-styled/src/main/resources/META-INF/resources/_styled).
 
+### types
+
+```sh
+liferay-npm-scripts types
+```
+
+Freshens all TypeScript type definition files in a `liferay-portal` checkout. Normally, these artifacts (`.d.ts` and `tsconfig.tsbuildinfo` files) would be committed along with changes to the corresponding module, but this command exists as a convenience for doing a global refresh across the entire repo (which would otherwise be tedious to do by hand because the projects must be built in dependency order).
+
 ## Config
 
 > Note: as of v2.x the config file was renamed from `.liferaynpmscriptsrc` to `npmscripts.config.js`
--- a/projects/npm-tools/packages/npm-scripts/src/index.js
+++ b/projects/npm-tools/packages/npm-scripts/src/index.js
@@ -57,6 +57,10 @@ module.exports = async function () {
 			require('./scripts/theme').run(...ARGS_ARRAY.slice(1));
 		},
 
+		types() {
+			require('./scripts/types')();
+		},
+
 		webpack() {
 			require('./scripts/webpack')(...ARGS_ARRAY.slice(1));
 		},
--- a/projects/npm-tools/packages/npm-scripts/src/index.js
+++ b/projects/npm-tools/packages/npm-scripts/src/index.js
@@ -7,6 +7,7 @@ const minimist = require('minimist');
 
 const ProcessExitError = require('./utils/ProcessExitError');
 const instrument = require('./utils/instrument');
+const log = require('./utils/log');
 
 module.exports = async function () {
 	const ARGS_ARRAY = process.argv.slice(2);
@@ -37,17 +38,15 @@ module.exports = async function () {
 			// Storybook is temporarily disabled until it supports webpack 5
 			// require('./scripts/storybook')();
 
-			/* eslint-disable-next-line no-console */
-			console.log(`
-			
-	WARNING:
-
-	Storybook has been temporarily disabled because it does not support 
-	webpack 5.
-
-	See https://bit.ly/35zFX4E for more information.
-
-`);
+			log(
+				'',
+				'WARNING:',
+				'',
+				'Storybook has been temporarily disabled because it does not',
+				'support webpack 5.',
+				'',
+				'See https://bit.ly/35zFX4E for more information.'
+			);
 		},
 
 		test() {
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
@@ -7,6 +7,7 @@ const fs = require('fs');
 const path = require('path');
 
 let buildSass = require('../sass/build');
+let runTSC = require('../typescript/runTSC');
 const createTempFile = require('../utils/createTempFile');
 const getMergedConfig = require('../utils/getMergedConfig');
 const instrument = require('../utils/instrument');
@@ -16,7 +17,6 @@ let runBridge = require('../utils/runBridge');
 let runBundler = require('../utils/runBundler');
 const setEnv = require('../utils/setEnv');
 let {buildSoy, cleanSoy, soyExists, translateSoy} = require('../utils/soy');
-const spawnSync = require('../utils/spawnSync');
 const validateConfig = require('../utils/validateConfig');
 let createBridges = require('../webpack/createBridges');
 let webpack = require('./webpack');
@@ -32,6 +32,7 @@ const CWD = process.cwd();
 	runBabel,
 	runBridge,
 	runBundler,
+	runTSC,
 	soyExists,
 	translateSoy,
 	webpack,
@@ -44,6 +45,7 @@ const CWD = process.cwd();
 	runBabel,
 	runBridge,
 	runBundler,
+	runTSC,
 	soyExists,
 	translateSoy,
 	webpack,
@@ -90,8 +92,10 @@ module.exports = async function (...args) {
 	const runLegacyBuild = !federation || federation.mode !== 'default';
 
 	if (inputPathExists && runLegacyBuild) {
-		if (fs.existsSync('tsconfig.json')) {
-			spawnSync('tsc', ['--noEmit']);
+		const isTypeScript = fs.existsSync('tsconfig.json');
+
+		if (isTypeScript) {
+			runTSC();
 		}
 
 		runBabel(
--- a/projects/npm-tools/packages/npm-scripts/src/typescript/README.md
+++ b/projects/npm-tools/packages/npm-scripts/src/typescript/README.md
@@ -1,0 +1,62 @@
+# TypeScript build support in Liferay DXP
+
+If you're reading this page you may have arrived from a link inside a `tsconfig.json` file in [`liferay-portal`](https://github.com/liferay/liferay-portal). The purpose of this document is to describe how those configuration files are managed and how the TypeScript build process works in general.
+
+## High-level overview
+
+We use [TypeScript's](https://www.typescriptlang.org/) [project references feature](https://www.typescriptlang.org/docs/handbook/project-references.html) because it allows us to share type information between projects (corresponding to OSGi modules) without having to perform a global build of all TypeScript in DXP for every change.
+
+Each project writes out a set of `.d.ts` files and a `tsconfig.tsbuildinfo` file in a `types` directory that sits alongside `src` and should be committed to the repo. Any project which wishes to depend on another can use the previously committed type information, thus relying on the other project without needing to build it first. This will enable us to preserve fast "local" development inside each project even as the overall amount of TypeScript across the repo grows.
+
+Note that we use `tsc` (the TypeScript compiler) _only_ for type-checking. The actual work of transforming code from `src/` for deployment is done via Babel (which just strips out the TypeScript type annotations without performing any type-checking of its own) or webpack.
+
+## How configuration works
+
+Each project must have a `tsconfig.json` file at its root, alongside the `package.json`. To start using TypeScript in a project, a `touch tsconfig.json` is sufficient. When running `yarn build` (either directly, or indirectly, via `gradle deploy` or similar), `npm-scripts` will update the configuration file based on three inputs:
+
+1. **[The base config](../config/tsconfig-base.json) that is bundled inside `npm-scripts` and is common across all projects.**
+2. **An optional `@overrides` field that the project can use to overwrite the default settings.** For the most part, this should not be used; please [open an issue](https://github.com/liferay/liferay-frontend-projects/issues/new/choose) if you discover a use case that is not met by the default configuration.
+3. **Information about the TypeScript dependency graph that is derived from the `dependencies` declared in the `package.json` files in `liferay-portal`.** That is, for a project like `@liferay/frontend-js-react-web` to depend on `@liferay/frontend-js-state-web`, it must declare that dependency in the `package.json` file. Note that adding something to `dependencies` makes the dependency visible to our TypeScript build process _but it has nothing to do with how modules are loaded at runtime_; be aware that you may need a `liferay-npm-bundler` `imports` configuration as well ([defaults can be seen here](https://github.com/liferay/liferay-portal/blob/5523f3a3b89cd25deb367a6fdea3d8bcbe420b04/modules/npmscripts.config.js#L31-L194)) in order for one project to load JS from another project at runtime.
+
+Any time the contents of this file are out of date, the `build` subcommand will print a large warning reminding you to include the changes in a commit.
+
+**Note:** The `main` field in the `package.json` must be correctly configured for any module that uses TypeScript. It should indicate the path of the main entry point for the module, relative to the `src/main/resources/META-INF/resources/` directory, and it should end with a `.js` extension, even if the source file is a `.ts` file. This is because the `main` field is used at _runtime_ by our module loading infrastructure, so it should reflect the name that the file will have _after_ it has been transformed from TS into JS.
+
+## How building works
+
+As stated above, our build process uses project references to allow each sub-project to be built in relative isolation. But note that if `A` depends on `B`, and `B` depends on `C`, then a build of `A` can only succeed if `B` has been built first (and the resulting artifacts committed to the repository), and in turn `B` won't succeed unless `C` has been built first.
+
+As such, our tooling will print a reminder to commit updated artifacts any time they change, and our `check` script (described below) will report stale artifacts when run in the context of CI (`ci:test:sf`) if there are any changes that affect a TypeScript module. (Note that we don't have much TypeScript yet, so this check is very fast; as adoption grows, we will monitor performance of both the build and the check.)
+
+If you try to build a project when its dependencies have not previously been built, you will see an error like:
+
+```
+error TS6305: Output file '../some/path/to/something/like/index.d.ts'
+has not been built from source file '../some/path/to/something/like/index.ts'.
+```
+
+If such a situation arises (although it shouldn't, due to the `ci:test:sf` measures already mentioned), you can either build the dependent project by hand (eg. `cd ../path/to/other/project && yarn build`) or you can run the provided `types` subcommand to refresh _all_ of the types in the repo at once, in dependency order (eg. `yarn run liferay-npm-scripts types`).
+
+## How checking works
+
+In order to keep `ci:test:sf` fast, we only check for stale TypeScript artifacts if a PR includes changes that touch a directory containing a `tsconfig.json` file. Under the covers, we detect this situation by observing the `LIFERAY_NPM_SCRIPTS_WORKING_BRANCH_NAME` environment variable, which is only set in CI, and running a `git diff` against the merge base (ie. the point where the PR's branch diverged from the main line, usually `master`).
+
+Note that type-checking is a fundamentally global operation, because a change in any part of the system can in theory affect any other part. Due to the small number of projects we are starting with, we are not making any attempts to optimize this away, but we do have options available to us for parallelizing the checks, as permitted by the dependency graph. So, for now, if any changes are made to a TypeScript project, we check the types for all TypeScript projects and report a failure if any of them are stale.
+
+## Porting existing projects to TypeScript
+
+[This example PR](https://github.com/liferay-frontend/liferay-portal/pull/942) shows how a module (in this case `@liferay/frontend-js-react-web`) can be ported in steps. For example, a single file can be ported to TypeScript without interfering with the build, and then successively other files can be migrated until coverage is complete (although, see "Limitations" below). To port a file, simply change the extension from `.js` to `.ts` and run the build (`yarn run build`) and fix any reported errors.
+
+Note that `yarn run tsc --noEmit --watch` will put the compiler into "watch" mode for a shorter feedback loop. And for an even tighter loop, configure your editor to use tooling based on the [Language Server Protocol](https://microsoft.github.io/language-server-protocol/); this will vary by environment, but can offer many useful features such as inline documentation (tooltips), autocomplete, "jump-to-definition", and so on.
+
+## Limitations
+
+At the time of writing, we still rely quite heavily on the `Liferay` global object that is assembled (defined and later augmented) in several dispersed locations within the repo. We plan to provide some centralized type information for this critical object soon, which should help catch usage mistakes from inside TypeScript files.
+
+## See also
+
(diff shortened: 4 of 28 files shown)


projects/npm-tools/packages/npm-scripts/README.md
```diff
@@ -165,6 +165,14 @@ liferay-npm-scripts theme build
 
 Runs the "build" task, providing it with the configuration it needs to find core dependencies such as the [`liferay-frontend-theme-styled` base theme files](https://github.com/liferay/liferay-portal/tree/master/modules/apps/frontend-theme/frontend-theme-styled/src/main/resources/META-INF/resources/_styled).
 
+### types
+
+```sh
+liferay-npm-scripts types
```
jbalsas, line 171: Are you thinking we should run this before we start compiling modules as part of `ant all`, like we do `yarn install`?

projects/npm-tools/packages/npm-scripts/src/index.js
```diff
@@ -37,17 +38,15 @@ module.exports = async function () {
...
-			
-	WARNING:
-
-	Storybook has been temporarily disabled because it does not support 
-	webpack 5.
-
-	See https://bit.ly/35zFX4E for more information.
-
-`);
+			log(
```
jbalsas, line 41: huh... is this still true? we're still using wp5? :D

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
```diff
@@ -0,0 +1,234 @@
...
+		const z = {name: 'z', dependencies: {}};
+
+		expect(
+			getTypeScriptBuildOrder({a, b, c, d, j, k, l, m, n, x, y, z})
+		).toEqual([z, j, a, k, b, y, x, l, c, m, n, d]);
+
+		/* eslint-enable sort-keys */
+	});
+
+	it('complains if it detects a cycle', () => {
```
jbalsas, line 203: Is this a hard limitation? I've briefly googled for TS cyclic refs and there seems to be some literature but didn't fully get into it. I know deps loading order is solvable to a certain extent in the JS context, so wondering how's that different here (I assume partial type loading/compilation is not a thing)

What should we tell a developer that faces this issue? How can we provide some guidance on how to refactor their code to avoid cycles?

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
```diff
@@ -0,0 +1,234 @@
...
+				'@liferay/remote-app-client-js': {
+					dependencies: {},
+					name: '@liferay/remote-app-client-js',
+				},
+				'@liferay/some-other-module': {
+					dependencies: {},
+					name: '@liferay/some-other-module',
+				},
+			})
+		).toEqual([
```
jbalsas, line 41: Is this stable in that it always returns in the same order or depends on the input? Is there a strong reason for one thing or the other?

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
```diff
@@ -0,0 +1,234 @@
...
+		const l = {name: 'l', dependencies: {z: '*', x: '*'}};
+		const m = {name: 'm', dependencies: {x: '*'}};
+		const n = {name: 'n', dependencies: {}};
+		const x = {name: 'x', dependencies: {y: '*'}};
+		const y = {name: 'y', dependencies: {}};
+		const z = {name: 'z', dependencies: {}};
+
+		expect(
+			getTypeScriptBuildOrder({a, b, c, d, j, k, l, m, n, x, y, z})
+		).toEqual([z, j, a, k, b, y, x, l, c, m, n, d]);
```
jbalsas, line 198: Out of curiosity... where would an unrelated dependency `U` (or a tree) that has no shared `dependencies` with the rest of the tree (A,B,C...X,Y,Z) sit in the end result? Before, after, in between? Is it reproducible?

In this case, I'm wondering if loading unrelated deps first or sorting by smaller graphs first could result in freeing up work faster. Of course, this hardly applies here so I'm thinking more about a JS loading context, so feel free to ignore this and everything else I said 😂

projects/npm-tools/packages/npm-scripts/src/utils/getJestModuleNameMapper.js
```diff
@@ -33,10 +29,9 @@ const IGNORE_GLOBS = ['node_modules/**'];
  *
  * We create such mappings by:
  *
- *    1. Reading the Yarn workspace globs defined in the top-level
- *       "modules/package.json".
- *    2. Expanding those globs to identify possible projects.
- *    3. Selecting only projects which have a "package.json" with a "main"
+ *    1. Iterating over projects identified by the Yarn workspace globs
+ *       defined in the top-level "modules/package.json".
+ *    2. Selecting only projects which have a "package.json" with a "main"
```
jbalsas, line 34: What's the reason for this? Not that I'm against, just wondering...

projects/npm-tools/packages/npm-scripts/src/typescript/getTypeScriptDependencyGraph.js
```diff
@@ -0,0 +1,62 @@
...
+
+const getWorkspaces = require('../utils/getWorkspaces');
+
+/**
+ * Across the set of Yarn workspaces, determine the dependency graph of
+ * TypeScript projects.
+ *
+ * Note that in the future, `tsc` itself may be able to do this for us.
+ *
+ * See: https://github.com/microsoft/TypeScript/issues/25376
```
jbalsas, line 17: Loving the energy in the last comment:

Screen_Shot_2021-03-31_at_12_49_06

shutupandtakemymoney

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -90,8 +92,10 @@ module.exports = async function (...args) {
 	const runLegacyBuild = !federation || federation.mode !== 'default';
 
 	if (inputPathExists && runLegacyBuild) {
-		if (fs.existsSync('tsconfig.json')) {
-			spawnSync('tsc', ['--noEmit']);
+		const isTypeScript = fs.existsSync('tsconfig.json');
```
jbalsas, line 95: Last question... I think we mentioned `tsconfig.json` might be auto-generated from within the tool but I'm not seeing that in here... is that still part of the plan for the initial roll-out or are we planning to do manual `tsconfig.json` for now and re-evaluate?

jbalsas: Screen Shot 2021-03-31 at 12 54 00

facepalm

projects/npm-tools/packages/npm-scripts/src/index.js
wincent: You know [I asked about removing federation related code here](https://liferay.slack.com/archives/C3JBR21HA/p1617026400057500), but was told not to. 😢

projects/npm-tools/packages/npm-scripts/README.md
wincent: We could, but that's not what I was thinking. The idea is that the artifacts get updated as part of individual `build` in each module, and they produce changes that should be committed along with that module.

So we accompany that with an `ci:test:sf` check that ensures they really are up-do-date (ie. were committed), and this subcommand is dual purpose:

- A convenience that enables us to do a "global refresh" if somebody ever merges something that hasn't passed SF.
- A thing we can use as an implementation detail of the `ci:test:sf` check to report when things are stale.

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
wincent: AFAIK, it is a hard limitation. You cannot do a topological sort of a dependency graph that has cycles because, by definition, there _is_ no order in which you can guarantee that every item's dependencies have been built before the item itself. It is possible that `tsc` might have some magic in there to support cycles between projects (now or in the future); I didn't even search, beyond noting that if you try building `frontend-js-react-web` before `frontend-js-state-web` has produced its artifacts, you get an error:

```
error TS6305: Output file
'$PORTAL/modules/apps/frontend-js/frontend-js-state-web/build/node/packageRunBuild/dist/src/main/resources/META-INF/resources/index.d.ts'
has not been built from source file
'$PORTAL/modules/apps/frontend-js/frontend-js-state-web/src/main/resources/META-INF/resources/index.ts'.
```

There are some forms of cyclic refs that are fine in TS, but those won't be project-level refs. For example, given two files in the same project: a file `a.ts` can import a _type_ from `b.ts`, and `b.ts` can import stuff (types or values) from `a.ts`; in fact, circular type imports aren't even uncommon in TS in my experience. But circular _value_ imports should be avoided as a matter of hygiene, IMO.

Our bundler/loader didn't support circular imports for a while, IIRC, and for good reason. I think we ended up hacking support in there only because there were misbehaved third-party deps that had cycles in them, and we couldn't control those.

As far as advising devs on how to avoid cycles, the answer is always the same: hoist — if `foo` and `bar` depend mutually on one another, extract cyclic part up into `common` so that the mutually dependent bits live in the same file. In some cases, this will actually eliminate `foo` and/or `bar`, but in others there may be some remnant; either way though, the dependency graph becomes acyclic. If the cycle is longer (eg. `foo` depends on `bar` depends on `baz` depends on `foo`), the answer is still to hoist.

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
wincent: Some examples are so trivial as to be stable no matter what. This one, concretely, depends on input order. As you move to more complicated graphs (even this little one), you get stable output unless you change the order the dependencies are declared, or the order in which you list the nodes in graph, _but_ the output is guaranteed to indicate a valid topological ordering.

This all stems from the fact that there can be more than one valid topological orderings of a graph. The simplest example that is actually interesting that illustrates this is the "diamond-shaped" graph seen elsewhere in this test file:

```
         C
        / \        ^
       A   D       | dependencies point upwards
        \ /        |
         B
```

`C` _must_ come first, then it doesn't matter what order you build `A` + `D` in, and `B` can only happen after all the others are done. So the two valid orderings are `[C, A, D, B]` and `[C, D, A, B]`. Which one you produce depends on implementation details + the ordering you pass in, as I mentioned above.

> Is there a strong reason for one thing or the other?

If we really cared about stability, I think we could guarantee it by enforcing input order (ie. that the nodes in the graph were passed in in lexicographical order, and that they declared their dependencies in lexicographical order too).

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
wincent: > Before, after, in between? It depends?

Without having tested it, I would say it depends on where you list it in the graph. JS will iterate over objects in insertion order, so the possible orderings are:

- If `U` is inserted first, it will appear first in the output.
- If `U` is inserted last, it will appear last in the output.
- If `U` is inserted somewhere in the middle, it will appear after all the nodes reachable via DFS from the preceding nodes have been visited.

> In this case, I'm wondering if loading unrelated deps first or sorting by smaller graphs first could result in freeing up work faster. 

An excellent thought to bear in mind when we start scaling this. Imagine we have 200 TS projects in the future. We can break the graph up into connected components and run the builds for those in parallel. Of course, one would expect a very large part of the graph to be one huge connected component (many things, for example, will depend on `frontend-js-web` and/or `frontend-js-react-web` and/or `frontend-js-state-web`), but we can look at parallelizing the build within connected components as well. But obviously, no sense in doing that until we actually need to.

projects/npm-tools/packages/npm-scripts/src/utils/getJestModuleNameMapper.js
wincent: The need for a valid `main` value is explained a bit higher up in the comment. It's because we need to generate a pair of mappings, and one of them references the `main` file:

```json
{
    "^frontend-js-web$": "<rootDir>../../frontend-js/frontend-js-web/src/main/resources/META-INF/resources/index.es.js",
    "^frontend-js-web/(.*)": "<rootDir>../../frontend-js/frontend-js-web/src/main/resources/META-INF/resources/$1"
}
```

That's what makes `import {x} from 'frontend-js-web'` work in the context of Jest, because Jest can figure out that it actually means `<rootDir>../../frontend-js/frontend-js-web/src/main/resources/META-INF/resources/index.es.js`.

projects/npm-tools/packages/npm-scripts/src/typescript/getTypeScriptDependencyGraph.js
wincent: Yeah, just thinking how wonderful it would be if we could wrap `ant`/`gradle` in Bazel...

projects/npm-tools/packages/npm-scripts/src/utils/getJestModuleNameMapper.js
jbalsas: What would happen if you had a module that didn't export anything and wasn't imported by anyone? Would it need to create a bogus `main` value in order to be able to use TS internally?

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
jbalsas: Sounds like an interesting problem space to explore in the future!!

yum

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
jbalsas: Sounds good! 👍

projects/npm-tools/packages/npm-scripts/src/index.js
jbalsas: For what it's worth, I said I'm not super into webpack right now, so I'd say we should do some cleanup and get rid of everything that we don't use. Tooling is complex enough without having to reason with things that we may or may not use.

Maybe @izaera can pick this up once you're done with TS and he's done with DatasetDisplay... 🤔

projects/npm-tools/packages/npm-scripts/README.md
jbalsas: Ok, that's the part I was missing. That'd be good and won't impact build time, which is already off the charts 👍

jbalsas (commented):
When in doubt, this message should serve as approval to move this forward, now and in the future, with or without (but preferably with) ninja gifs.

projects/npm-tools/packages/npm-scripts/src/utils/getJestModuleNameMapper.js
wincent: I think all modules should have a `main`, even if it doesn't export anything (rare, but modules that exist solely for their side-effects: all the parts of the system still need to now _what_ to run, and that's what `main` is for).

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
wincent: Do you have any concrete examples in mind? I'm aware of `frontend-js-web` + `frontend-js-aui-web` having a mysterious spooky-at-a-distance mutual dependency because they cohabit the same space via the `FragmentHost` mechanism, and parts of `Liferay` are actually set up by JSP etc. But I am not too worried about `aui-web` as we shouldn't really be touching it (other than to delete it). Any other instances to have on our radar?

projects/npm-tools/packages/npm-scripts/README.md
wincent: It's written down somewhere (Jira ticket? commit message? other PR?) but the whole idea of using project references is to enable fast builds, because each project only has to worry about its own little bit of the build and you avoid expensive global operations.

projects/npm-tools/packages/npm-scripts/test/typescript/getTypeScriptBuildOrder.js
jbalsas: > Do you have any concrete examples in mind? 

I think `frontend-taglib-clay` <-> `frontend-js-react-web` (with or without `frontend-js-web` in between). We know the react one uses `ClayIconSpriteContext` when rendering and that in the clay one we likely use hooks offered by the react module.

wincent: Taglibs, you say...

taglibs

jbalsas: smellofnewmodules

wincent:
Today's update. The logic for auto-generating the config is done and just needs to be wired up (presuming I don't find any horrible bugs). All the details are in the last commit (currently [518b7a5ad822c](https://github.com/liferay/liferay-frontend-projects/pull/478/commits/518b7a5ad822c76e01beda5d668c9315350124cc)). I see CI is red, but tests pass locally for me, _I swear!_ — will look into that.

---

**feat(npm-scripts): add `configureTypeScript()` to manage tsconfig.json**

**High-level overview:**

- This function takes an existing `tsconfig.json` file and brings it up-to-date based on three inputs: (1) A base config that is bundled inside `npm-scripts` and is common across all projects; (2) An `@overrides` field that contains any, er, overrides, that a project might need to apply; and (3) Information about the TS dependency graph that is needed to set up project references between projects.

- The function returns `true` if everything is up-to-date and `false` otherwise, which means that we will be able to plug it in to the rest of the system to perform two operations: (1) Freshen the config (can also be used to generated a brand new one if you just do `touch tsconfig.json` first to identify the project as a TS one); or (2) Check that the config is up to date. So the plan is to plug (1) into the "build" subcommand and (2) into the "check" subcommand (in CI).

- In order to detect changes and prevent unintended manual edits, every config file has a `@generated` hash and clear `@readonly` warning to keep edits inside the `@overrides` section alone. There is also a `@see` file with a short-link to [this URL](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/packages/npm-scripts/src/typescript/README.md) (which I haven't written yet, but will, and which will explain all of this).

- Note that the hash only cares about the values in the file, not the whitespace or formatting. This means that if you modify the file with Java SourceFormatter or Prettier the hash will remain the same. We already have a mess where those two tools fight with one another of what JSON should look like, so we do our best here to at least not make thing worse there. When we write JSON, we do so in a format which (I hope) matches what SourceFormatter wants, and we only write it when absolutely necessary (ie. because the values have changed).

There is one TODO noted inline, as well as the missing docs I mentioned above. Finally, I need to test this in liferay-portal and actually integrate it with the "build" and "check" subcommands, but I'll do that in another commit.

**Details:**

- `configureTypeScript()` is a simple function that takes the dependency graph information (from `getTypeScriptDependencyGraph()`) and no other parameters; it configures the "tsconfig.json" file in current working directory. I did it this way in order to make it simple for callers, even though it makes writing the tests harder. Whatevs. The tests were going to be hard no matter what because you need a bunch of fixtures.

- Given those fixtures, we can copy them on every test (slow), or do some naughty writing in-place (ie. to the "tsconfig.json" files in the "__fixtures__" directories). I did the naughty thing because I don't want headaches with temporary files on Windows in CI. (Speaking of which, there are a few `path.join` type things that may require a good review/test on Windows.)

- Doing the naughty thing requires us to modify the `.gitignore`, because we don't want changes to these `tsconfig.json` files appearing in the `git diff` output. Note that while I was in there I stripped out the useless "node_modules" entry (already covered by the `.gitignore` in the repo root).

- We tell Jest to ignore the `__fixtures__` directory because otherwise a `yarn test --watch` run will infinite loop because every time you run the tests, you touch the file-system, which causes Jest to run again, which causes you to touch the file-system...

- Our base `tsconfig.json` that we use as a template is actually called `tsconfig-base.json` because I don't want `tsc` or any editor LSP picking it up and thinking that the containing `config/` folder is the root of a TS project.

- The tests have a few `sort-keys` lint suppressions in them for readability reasons (to make the precendence of the spread/merge operations clear).

**Test plan:** For now, just unit tests.

projects/npm-tools/packages/npm-scripts/src/index.js
izaera: Just ping me when your are ready @georgel-pop-lr  :point_right: https://github.com/liferay/liferay-frontend-projects/issues/480 

> You know I asked about removing federation related code here, but was told not to. :cry: 

there there

wincent: this-is-not-over

projects/npm-tools/packages/npm-scripts/src/typescript/README.md
```diff
@@ -0,0 +1,62 @@
+# TypeScript build support in Liferay DXP
+
+If you're reading this page you may have arrived from a link inside a `tsconfig.json` file in [`liferay-portal`](https://github.com/liferay/liferay-portal). The purpose of this document is to describe how those configuration files are managed and how the TypeScript build process works in general.
+
+## High-level overview
+
+We use [TypeScript's](https://www.typescriptlang.org/) [project references feature](https://www.typescriptlang.org/docs/handbook/project-references.html) because it allows us to share type information between projects (corresponding to OSGi modules) without having to perform a global build of all TypeScript in DXP for every change.
+
+Each project writes out a set of `.d.ts` files and a `tsconfig.tsbuildinfo` file in a `types` directory that sits alongside `src` and should be committed to the repo. Any project which wishes to depend on another can use the previously committed type information, thus relying on the other project without needing to build it first. This will enable us to preserve fast "local" development inside each project even as the overall amount of TypeScript across the repo grows.
```
wincent, line 9: About these docs: once gets merged I'll have some concrete config to link to and this can become a bit less abstract.

Reminder to self: clarify what happens with `tsconfig.tsbuildinfo` files (they can't be committed, so they have to go in `tmp/`, as explained in "refactor(npm-scripts): store tsbuildinfo in tmp/").

wincent:
Ok, given discussion on Slack, going to merge this. Based on testing it in liferay-portal in [this PR](https://github.com/liferay-frontend/liferay-portal/pull/942) with the [v42.0.0-pre.3 release](https://github.com/liferay/liferay-frontend-projects/releases/tag/npm-scripts%2Fv42.0.0-pre.3), it seems to work. I rebased it and squashed a couple of the commits. Will merge once I see CI green. Can always follow up with fixes if feedback comes in or we discover problems with it.

wincent:
```
FAIL projects/amd-loader/test/loader/loader.js
  ● Loader › require can be called with different signatures › string, function, null

    : Timeout - Async callback was not invoked within the 200 ms timeout specified by jest.setTimeout.Timeout
```

Pretty obviously not a real failure, so running again. Only happens on Windows.

bryceosterhaus:
So exciting to comeback from leave and see TS support! Great job

PR #482, opened 2021-04-06
docs(npm-scripts): update explanation of `tsconfig.tsbuildinfo`

wincent:
As seen in 458e4908bbe1ab58d780616f6d68, we had to change our build tooling to _not_ commit these files, principally because they contain machine-specific information. So, remove the old reference to `tsconfig.tsbuildinfo` and replace it with a paragraph further down explaining its purpose and usage.


wincent:
Doc updates that wouldn't hurt a fly, so going to _moige_ this.

innocent

Issue #483, opened 2021-04-06
Teach `liferay-npm-scripts {check,fix}` about `$INIT_CWD`


wincent:
At the moment if you run `yarn run checkFormat` (or `yarn run format`) in an OSGi module that doesn't have a `package.json` file, `yarn` will walk up to the top-level `modules/` folder and run from there, which means _everything_ in the repo gets checked (slow).

Turns out `npm`/`yarn` both set `$INIT_CWD` so you can figure out where you originally started, which means we can scope our check/fix to that directory. This seems to be an undocumented feature, as far as I can tell (discovered it by logging `process.env` and inspecting), but it seems that we can leverage it if it's there, and degrade gracefully if it isn't.

bryceosterhaus:
> Turns out npm/yarn both set $INIT_CWD

This is a great addition, had no idea that existed

PR #484, opened 2021-04-06
feat(npm-scripts): allow format/lint to run without package.json

wincent:
As noted in the doc comment, Yarn (when we do `yarn run liferay-npm-scripts`) will `cd` to the closest project root (ie. directory that has a `package.json` file) before invoking our tool. By that time, it is "too late", to know where we were invoked from. Fortunately, it sets an `INIT_CWD` environment variable with this information.

Whenever we run the `lint()` or `format()` functions (whether in `check` or `fix` modes), we check to see if `INIT_CWD` is defined and different from our apparent "current" working directory. If there is a discrepancy, that means that the user is probably working down inside a `src/` directory (or similar), or they are running in a project that doesn't have a `package.json` (eg. `click-to-chat-web`). Nevertheless, projects like `click-to-chat-web` do have JSP in them even though they don't have a `package.json`, so it is reasonable to want to format them with our frontend formatting tools.

Now, prior to this change, running `yarn run format` in a place like `click-to-chat-web` would result in `yarn` starting from the "top" level (`modules/`) directory and running formatting for the entire repo, which would work but is slow. Running from a subdirectory inside a project with a `package.json` (like `frontend-js-react-web`) wasn't quite as bad, because `yarn` would `cd` up to the project root where the `package.json` lives and run from there; our default globs would take effect (which have the form `/src/**/*.jsp` etc).

After this change, if we detect that `INIT_CWD` is set and different, that means we are "down" somewhere where there is no `package.json`. We walk up the filesystem looking for a `build.gradle` or a `package.json`, whichever we see first, and we run from there. This means that one of three scenarios may apply:

- Inside `click-to-chat-web` you get exactly what you would want and the JSP files get checked/formatted.
- Inside a subdirectory of `frontend-js-react-web` we behave as though we'd run from the project root.
- In a place like `modules/apps/frontend-js`, which does have a `build.gradle`, we stop there and run with the default globs (eg. `/src/**` etc) which effectively means we format nothing (previously we would format _everything_, which is slow).

So what we have is strictly an improvement over what we had, but we should note these "gotchas":

- `gradlew formatSource` will not invoke `liferay-npm-scripts` in modules without `package.json` files. This is something that we could ask Hugo to change if we cared. I suspect we don't want to do that though because it could have unpleasant side-effects (eg. on an older branch not using this version of `@liferay/npm-scripts`, it would fall back to the old behavior of running frontend formatting across the entire repo just because somebody ran `gradlew formatSource`; not good because it is too slow).
- I guess you could run this from "crazy" places which have `build.gradle` files but aren't really what we'd consider to be "projects" (I'm thinking of places under "third-party/" or "util/" or "sdk/" for example). But if you do that with an explicit `yarn run format` you probably are asking for it, and unintended formatting won't happen anyway unless your directory has `src/**` stuff in it. On the off chance that you do mess something up, our Java Source Formatter has a reasonable chance of catching anything egregious during `ci:test:sf`, and our formatting changes are supposed to be relatively benign in any case. Overall, I think it's worth it.

**Test plan:**

1. Copy over these changes to a liferay-portal checkout and add some `console.log()` to see the current working directory, globs, and paths after glob expansion. Run under all the scenarios listed above, plus from the `modules/` root and see that it continues working as before.
2. Run unit tests to make sure we didn't break anything. Note the hacky 😱 setting of `INIT_CWD` necessary to stop our `format()` tests from `cd`-ing out of where we want them to actually run.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/483

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/format.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/format.js
@@ -10,6 +10,7 @@ const isJSP = require('../jsp/isJSP');
 const prettier = require('../prettier');
 const getMergedConfig = require('../utils/getMergedConfig');
 const getPaths = require('../utils/getPaths');
+const inCurrentWorkingDirectory = require('../utils/inCurrentWorkingDirectory');
 const log = require('../utils/log');
 const {SpawnError} = require('../utils/spawnSync');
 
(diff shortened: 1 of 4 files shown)


izaera:
It looks OK.

My only doubt is what would happen outside of liferay-portal scenarios (for example in blade workspaces). It sounds like it would work too, but we should have it in mind when the devtools team upgrade the npm-scripts version.

wincent:
> My only doubt is what would happen outside of liferay-portal scenarios (for example in blade workspaces). It sounds like it would work too, but we should have it in mind when the devtools team upgrade the npm-scripts version.

Yep. Good point. I understand that most Blade workspaces probably aren't existing in a Yarn workspaces context, and many customers are probably using `npm`. I think both `npm` and Yarn-not-in-a-workspace have a similar behavior, in that they will `cd` up looking for `package.json` in order to run.

So, with this change, the biggest difference would be that `liferay-npm-scripts format` would run at the first found `build.gradle`, if there was one, before getting to that `package.json`. That _sounds_ pretty innocuous to me.

hand-waving

PR #487, opened 2021-04-07
feat(npm-scripts): format generated `.d.ts` files

wincent:
As [requested here](https://github.com/brianchandotcom/liferay-portal/pull/100574#issuecomment-814446826).

Note that this still isn't going to be the most beautiful code ever, but here is the change:

- [Before gist](https://gist.github.com/wincent/9a55b26389b9e1e15ec802f7feef0b13)
- [After gist](https://gist.github.com/wincent/3ff31e80550cc00d3df39a03659ed0b0)

Note that `tsc` tends to run things together, avoiding totally empty lines, so the style is quite cramped. We don't have any lints that force stuff to be spaced out (ie. even if I were to run ESLint over the files in addition to Prettier), so this is about as good as it gets. Note that at least the docblock comments have an empty line above them, so that's something. Also note that we hackily insert an extra line after the license header to stop it from looking horrible. But more than that, I don't think is worth doing.

Also note that I had to make a few functions `async` so that I could `await` on the result of `configureTypeScript()`, because I actually need to know the `outDir` in order where to figure out where to look for `.d.ts` files to format. And obviously, I changed the return type of that function from boolean to instead return the config itself.

One final thing to note (it's all about the notes): a `.tsx` file will produce a `.d.ts` file, not a `.d.tsx` file, so that's why we only need the former in our glob patterns.

**Test plan:** Copy this over to a liferay-portal checkout. Run `yarn run liferay-npm-scripts types` and see it apply changes like those I already showed in the Gist links above. See a warning printed about the artifacts not being up-to-date. Commit them. Run it again. See no warning. Rejoice.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/typescript/runTSC.js
+++ b/projects/npm-tools/packages/npm-scripts/src/typescript/runTSC.js
@@ -3,10 +3,14 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+const fs = require('fs');
 const path = require('path');
 
+const prettier = require('../prettier');
 const color = require('../utils/color');
+const expandGlobs = require('../utils/expandGlobs');
 const findRoot = require('../utils/findRoot');
+const getMergedConfig = require('../utils/getMergedConfig');
 const git = require('../utils/git');
 const log = require('../utils/log');
 const spawnSync = require('../utils/spawnSync');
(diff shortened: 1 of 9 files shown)


projects/npm-tools/packages/npm-scripts/src/typescript/runTSC.js
```diff
@@ -62,4 +70,43 @@ function runTSC() {
...
+				...config,
+				filepath,
+			};
+
+			const formatted = await prettier.format(source, prettierOptions);
+
+			// Fix missing trailing space after license header.
+
+			const fixed = formatted.replace(
+				/^\/\*\*\n(?: \*[^\n]*\n)* \*\/\n[^\n]/,
```
wincent, line 96: Sorry about the readability here, but I don't think we want to go as far as doing an AST-based transform for this little detail.

Here's a translation to make it more obvious what's going on here:

Screenshot 2021-04-07 -182032-X4lSuSLZ@2x

jbalsas: You should've started with this, now it makes total sense

jbalsas:
> [...] Run it again. See no warning. Rejoice

Good luck with that!!

wincent:
Going to interpret that as resounding endorsement and merge.

wincent:
In `liferay-portal` now via: https://github.com/brianchandotcom/liferay-portal/pull/100603

Issue #481, opened 2021-04-06
Compile list of starter resources and tips for working with TypeScript


wincent:
As [requested here](https://github.com/liferay-frontend/liferay-portal/pull/942#discussion_r607687846).

> There are zillions of resources out there but we should compile a list. The [official docs](https://www.typescriptlang.org/docs/) are great and there are good third-party resources like the [TypeScript Deep Dive](https://basarat.gitbook.io/typescript/).
>
> In this specific instance though, my process is:
>
> 1. Want to figure out type for event listener.
> 2. Go to [`lib.dom.d.ts`](https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts) and find [something relevant](https://github.com/microsoft/TypeScript/blob/3b06ef1a733904013d67cb1e1fdb366082f0f9dd/lib/lib.dom.d.ts#L4941-L4942).
>
> Note that I didn't literally go to the `lib.dom.d.ts` file on GitHub. I actually just used my "go to definition" keybinding inside Vim on `document.addEventListener` instance, and it jumped straight there (the `lib` files are all inside `node_modules`; they come with TypeScript). So this is one of those things that you don't know about until you know about it, but once you do, it is a treasure chest of useful info.

PR #488, opened 2021-04-08
docs(guidelines): document some additional TS resources

wincent:
With recent changes in our TS build and adoption in DXP, it's time to update this document. Closing the related issue because even an initial list like this is not a bad starting point. We can always invest more in building up the list over time, but I don't really feel like scouring the internet for learning materials right now; I think we should just get to work and observe what kind of questions/needs get surfaced.

Note I'm also including a back-link to the updated page from our "build" docs.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/481

### DIFF
--- a/guidelines/general/typescript.md
+++ b/guidelines/general/typescript.md
@@ -1,11 +1,31 @@
 # TypeScript
 
-## Can we use TypeScript in Liferay projects?
+## Status of TypeScript in Liferay projects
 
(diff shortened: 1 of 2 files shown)


jbalsas (approved):
slowclap

wincent:
Greg Hurrell, Technical Writer.

technical-writer

Issue #480, opened 2021-04-05
Remove webpack federation support


izaera:
### Issue type

-   :woman_shrugging: Other

### Description
Remove webpack federation support.

**Desired behavior:**
Because we won't be using it, we should remove it in a single PR so that it can be tracked in the future if necessary but, at the same time, remove complexity from npm-scripts.

See https://github.com/liferay/liferay-frontend-projects/pull/478#discussion_r604751255

PR #490, opened 2021-04-08
feat(npm-scripts)!: remove webpack federation support

izaera:
This is the implementation of issue #480

I have tested it locally with `ant all` in `liferay-portal` and it built correctly and had no apparent failures at runtime.


projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -108,25 +103,13 @@ module.exports = async function (...args) {
...
 
-	const runFederationBuild = federation && federation.mode !== 'disabled';
-
-	if (fs.existsSync('webpack.config.js') || runFederationBuild) {
+	if (fs.existsSync('webpack.config.js')) {
 		webpack(...args);
 	}
 
-	if (runLegacyBuild && bundling !== false) {
+	if (BUILD_CONFIG.bundler) {
```
wincent, line 110: Back before federation, [we used to call `runBundler()` unconditionally](https://github.com/liferay/liferay-frontend-projects/blob/8ff8cb02894f0b6dbcdc9e082e5baca61987ef15/projects/npm-tools/packages/npm-scripts/src/scripts/build.js#L102). Maybe we still should?

wincent (approved):
I believe you can delete [`getProjectMainModuleFilePath.js`](https://github.com/liferay/liferay-frontend-projects/blob/543c1c08e2fb24f0ed77ae22076ea419b34a3fd2/projects/npm-tools/packages/npm-scripts/src/utils/getProjectMainModuleFilePath.js) too, right?

Otherwise, lg2m based on my half-brained review.

wincent:
The other question that occurs to me, @izaera, is why do [we still have/need `tweakWebpackConfig()`](https://github.com/izaera/liferay-frontend-projects/blob/remove-webpack/projects/npm-tools/packages/npm-scripts/src/webpack/tweakWebpackConfig.js)? (It was added in f0cec1c5bc3979c60f62916510c63301a266030f, in name of federation.)

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: This is "new", in the sense that we didn't have that before, but since it proved itself useful when I introduced it for webpack federation, I'm leaving it.

izaera:
> I believe you can delete getProjectMainModuleFilePath.js too, right?

So it seems. Thx.

izaera:
> The other question that occurs to me, @izaera, is why do we still have/need tweakWebpackConfig()? (It was added in f0cec1c, in name of federation.)

Well, it was added, but we already had `mergeBabelLoaderOptions` which is still used. Before `tweakWebpackConfig` the webpack config manipulation was somewhat spread in the code, so I extracted it to a module (tweakWebpackConfig) where everything modifying the webpack configuration was held.

The "problem" with tweaking webpack configuration is that we can either do it before invoking webpack (modifying its input file) or inside the webpack process (when it loads `webpack.config.js`). I moved all the logic to `tweakWebpackConfig` so that everything took place when webpack is executed since there are things that cannot be done before that. I thought that, better than spreading things in different places of the code base and different moments in time, it made sense to have everything in one single place...

wincent:
> Well, it was added, but we already had `mergeBabelLoaderOptions` which is still used.

Got it. GitHub pretty much hid that from the diff view when I was looking at that commit. Whoops. 😬

PR #493, opened 2021-04-12
docs(guidelines): move repo sync documentation into Confluence

wincent:
Because the new documentation contains sensitive information (hostnames, IPs, usernames inside the VPN etc), we've [moved it into Confluence](https://liferay.atlassian.net/wiki/spaces/ENGFRONTENDINFRA/pages/1409615973/liferay-frontend+repo+sync) where it is somewhat safer from prying eyes. In Atlassian we trust.

PR #495, opened 2021-04-12
test(monorepo): make GitHub actions more granular

wincent:
It occurred to me while looking at PRs that only touch files under "guidelines/" ([recent example](https://github.com/liferay/liferay-frontend-projects/pull/493)) that we're running some pretty useless tests (for example, the full Jest test suite on both Linux and Windows).

So, this commit splits up our "workflows" files more granularly and makes use of the `paths-ignore` feature to make sure that we don't run tests needlessly.

-   "global" gets exploded out into multiple files; instead of running all the time unconditionally (and always doing "check-lockfile", "format-check", "lint", "test", and "test-windows"), we now have:
    -   "format:check" (always runs).
    -   "lint:check" (runs whenever there are code-ish changes; ie. any PR that isn't totally under "guidelines").
    -   "lockfile:check" (always runs).
    -   "projects" (runs tests on Linux and Windows for anything that isn't a doc-only change, or a change in one of the workspaces that is managed separately elsewhere, like "maintenance/projects/js-toolkit" and friends).

The only tricky bit here is noting that "lint:check" excludes PRs that only touch this explicitly listed set of directories:

-   "guidelines/css/"
-   "guidelines/dxp/"
-   "guidelines/general/"

The reason I did this instead of just listing "guidelines/" is that there _is_ code in that repo (eg. see the "guidelines/support/" directory), so we _do_ want that linted. Maybe one day in the future we will hoist the link checker up into "support/packages/", and at that point we can go for a simpler configuration here. Note that in the meantime I think this is better than the alternative: we could have just listed "guidelines/" but that would risk failing to catch issues in the code; but with the explicit listing, the worst that can happen is that we add some other subdirectory in the future (eg. "guidelines/foo/") and end up accidentally but harmlessly running the lints in CI. We can adjust the list if somebody ever notices...

Only other thing to note is that I synchronized all the cache breakers (to "3" instead of a mix of "3" and "2") along the way.


wincent:
As for how to test this, I guess I'd just go with ~"push and pray"~ "merge and monitor".

wincent:
What could possibly go wrong?

ninja-merge-10

PR #502, opened 2021-04-20
docs(senna): update deprecation notice to reflect new location

wincent:
The text came from the old repo, which is about to be archived. So let's update the text to make sense in its new home.


wincent:
Going to merge this because it is a trivial doc update.

wincent:
Obligatory `#NinjaMerge` gif that I forgot to include in previous comment:

ninja-merge-2

bryceosterhaus:
Not to be too picky here, but I've been noticing a lot of these ninja merges using the same gifs. Per our Ninja Merge Guidelines ™️,  think they require unique gifs for every merge.

wincent:
> Not to be too picky here, but I've been noticing a lot of these ninja merges using the same gifs.

You got me there. I've definitely been slipping. My only concern is that I'm starting to scrape the bottom of the barrel and will soon have to start using American Ninja Warrior gifs... 😟 

ninja-merge-11

PR #507, opened 2021-04-21
test(monorepo): update GitHub actions config to account for Senna

wincent:
We're not sure if we are going to be able to run `gulp test:saucelabs` like [we used to under Travis](https://travis-ci.org/github/liferay/senna.js/builds/589639702).

The `.travis.yml` config file has an encrypted key in it from which I obviously can't retrieve the plain-text, although Chema did share a possible `SAUCE_ACCESS_KEY` with me over chat that _might_ work. I'm sticking that in the team's 1Password vault for safe-keeping.

So, given that Senna is on life support (and that a local `yarn run gulp test` still works), in this commit we are removing the old Travis config and setting up GitHub actions instead. All they do is verify that the lockfile is non-stale and the project actually builds.

If we ever want to resurrect the Sauce tests, or implement a replacement, the pieces are in place to do so: we just need to add another entry under `steps:` in our GitHub action config. If we do that, [the "sauce-connect-action"](https://github.com/saucelabs/sauce-connect-action) may come in handy.


wincent:
Want to cut a Senna release so merging some stuff.

PR #509, opened 2021-04-21
chore(senna): adjust release process

wincent:
- Remove "build" directory. This just clutters up the repo and makes for another step during the release. The released package has zero references to anything under "build/", so it is useless.
- Teach Prettier to ignore "lib" directory.
- Add a ".yarnrc" file so that we can generate changelog updates and use `yarn version` to publish.
- Add "preversion" and "postversion" scripts so that we can publish using `@liferay/js-publish`.
- Update CONTRIBUTING.md: The release process is only described briefly here because I need to actually test it out to make sure it works. After the v2.7.10 release I will loop back and add any necessary clarification.


wincent:
I don't think anything can go wrong with this. It only removes 33k lines.

I checked 7.3 to make sure that `frontend-js-spa-web` isn't doing anything funky and reaching into `build/`, and it isn't...

ninja-merge-12

Issue #515, opened 2021-04-28
group-imports rule doesn't cope well with consecutive comments between imports


wincent:
### Issue type

-   :bug: Bug report

### Description

**Desired behavior:**

Doesn't get confused by consecutive comments between imports.

**Current behavior:**

Gets confused by:

```ts
import ClayTooltip from '@clayui/tooltip';
import {render, useTimeout} from '@liferay/frontend-js-react-web';

// Hack to get around frontend-js-web not being in TS
// @ts-ignore

import {ALIGN_POSITIONS as POSITIONS, align, delegate} from 'frontend-js-web';
import React, {
	useEffect,
	useLayoutEffect,
	useReducer,
	useRef,
	useState,
} from 'react';
```

As seen [here](https://github.com/liferay-frontend/liferay-portal/pull/992#issuecomment-827803097).

PR #517, opened 2021-04-28
fix(eslint-config): fix group-imports edge case

wincent:
Stops spurious complaints about `imports must be grouped (unexpected blank line before: "frontend-js-web")` given code as seen in the included regression test.

Closes: https://github.com/liferay/liferay-frontend-projects/issues/515

### DIFF
--- a/projects/eslint-config/plugins/liferay/lib/common/imports.js
+++ b/projects/eslint-config/plugins/liferay/lib/common/imports.js
@@ -6,30 +6,44 @@
 function getLeadingComments(node, context) {
 	const code = context.getSourceCode();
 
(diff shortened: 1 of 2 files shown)


wincent:
Ok, pushed another commit with the actual fix, but I still haven't tested it in liferay-portal yet, so keeping it as a draft. Will follow-up when I get back from picking up the kids from school.

wincent:
Taking this out of draft because I've had the chance to actually test it in `liferay-portal` now with Bryce's PR. Needed one small tweak, but I think it's good enough at this point. Will see what CI says, with a view to...

ninja-merge-14

Issue #235, opened 2020-11-13
Allow customizing webpack plugins in the package liferay-npm-build-support for lnbs-start


bozzbot:
Hi Liferay Team,
we are starting a new project and want to use the liferay-npm-build to create JS only Portlets. What we found is missing in the `liferay-npm-build-support` / `lnbs-start` was to allow the configuration of additional webpack plugins - e.g. the `vue-loader` which is needed for Vue.js 3.

As a first idea there could be a `plugins` section in the `.npmbuildrc`. E.g.
```json
{
	"translatorTextKey": "",
	"supportedLocales": [],
	"webpack": {
		"rules": [
		],
		"plugins": [
		     "new require('copy-webpack-plugin')(['../assets'])",
		     "new (require('vue-loader/dist/plugin'))"
		]
	}
}
```
Even if it is kind of strange to have the plugins written like this...

I assume there are good reasons against this, at it is not currently in the `liferay-npm-build-support`. I can provide a PR with such a configuration if you think that makes sense.

wincent:
cc @izaera who wrote `liferay-npm-build-support` and is additionally doing a bunch of webpack related stuff now; he'll probably be able to make a more informed comment on this than I can.

izaera:
I was checking this with product management because I'm unsure of the plans regarding `liferay-npm-build-support` as it was released as an experimental feature and hasn't evolved much since then. I'll report as soon as I get an answer.

bozzbot:
@wincent @izaera  Thank you for your fast feedback! Are there other "Liferay compliant ways" to run a Portlet in local development then?
For V3 there will be a different approach I assume?

izaera:
> Are there other "Liferay compliant ways" to run a Portlet [...]

I guess `deploy` target. I know it's not live development, but a lot of people use it and it shouldn't take too long, as the build of the `node_modules` directory (which is usually the one taking most of the build time) is cached.

For v3 we hope to be able to use webpack's watch feature.

izaera:
I have changed the title of the issue to a more generic one so that we can give a solution to all webpack configuration extension issues in the `npm run start` context.

PR #523, opened 2021-04-30
fix(js-toolkit): declare Liferay mock before requiring main module when running 'npm run start'

rozamirandar:
Recently created an Angular Widget using `yo liferay-js` and tried to run `npm run start` and got the following error:
image

I moved the declaration of the mocked Liferay object before requiring the main module. I also added the mock for `Liferay.ThemeDisplay.getPathContext` which is used by the Angular components to declare the `templateUrl` property:

```
@Component({
	templateUrl: 
		Liferay.ThemeDisplay.getPathContext() + 
		'/o/angular-sample/app/app.component.html'
})
export class AppComponent {
...
```

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/resources/start/index.js.ejs
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/resources/start/index.js.ejs
@@ -1,3 +1,16 @@
+window.Liferay = {
+	Language: {
+		get: function(key) {
+			return key;
+		}
+	},
+	ThemeDisplay: {
+		getPathContext: function() {
+			return '';
+		}
+	}
+};
+
 var main = require('<%= mainModule %>');
 
 var params = {
@@ -7,14 +20,6 @@ var params = {
 	configuration: {
 		portletPreferences: {},
 		system: {}
-	}
-};
-
-window.Liferay = {
-	Language: {
-		get: function(key) {
-			return key;
-		}
 	}
 };
 


izaera:
LGTM. If no one has anything against it, I will merge this.

This strategy of patching the `npm run start` feature as we go is a bit inefficient, but until we do #235 and move most of the feature to the project side, I guess we can live with it.

maintenance/projects/js-toolkit/packages/liferay-npm-build-support/src/resources/start/index.js.ejs
```diff
@@ -1,3 +1,16 @@
+window.Liferay = {
+	Language: {
+		get: function(key) {
+			return key;
+		}
+	},
+	ThemeDisplay: {
+		getPathContext: function() {
+			return '';
```
jbalsas, line 9: Feels like maybe at some point we could do better here... knowing which server we're running, maybe getting a path wouldn't be so hard and would make this work a bit better. Just thinking out loud here.

izaera: :thinking: We are planning to let the `npm run start` functionality be managed by users, so there's no point in making it smarter, I think... 

Since this type of configurations will be done by hand, by a user knowing the exact nuances of his setup, it will be easier to mock `Liferay` as needed without having to contribute a new PR and wait for a new release of the JS Toolkit.

jbalsas: Sounds good to me.. just pinging @dsanz here since he might be bored and wanting to add more stuff onto his thinking list (just kidding) 😂

izaera: In fact, I'm preparing a PR to externalize the `npm run start` thing. I'll send it in minutes, and we can discuss there...

izaera:
This is the PR I talked about in [this comment](https://github.com/liferay/liferay-frontend-projects/pull/523#discussion_r625685549) :point_right: https://github.com/liferay/liferay-frontend-projects/pull/526

In theory, if we merge that, we don't need this. Though we can still merge it, if we want, as it will still be used as a starting point.

wincent:
> LGTM. If no one has anything against it, I will merge this.

No objections from me.

izaera:
Let's merge this, since we are going to merge https://github.com/liferay/liferay-frontend-projects/pull/526 soon

PR #533, opened 2021-05-11
docs(js-themes-toolkit): add a new step to local themes toolkit test procedure

izaera:
We have found that the process may fail without this step.


izaera:
Merging this as it is only a minor change in documentation

PR #537, opened 2021-05-13
fix(npm-scripts): support importing css files under any node_modules directory

izaera:
I have tested this feature with the [PR linked in the issue](https://github.com/liferay-lima/liferay-portal/pull/1802/files).

It fixes the CSS issue as well as the PropsValues one.

Please read the commits because they explain the reasons behind the changes to the code.


bryceosterhaus (approved):
And I am fine with either using a resolver or the moduleNameMapper, was just curious why we would use one vs the other.

izaera:
> Is there a benefit to using a resolver rather than adding the moduleNameMapper to our config?

Not a benefit per se, but it makes things easier (and BTW probably faster :sweat_smile:, but that was not the goal). The reason is -more or less- explained in the issue, but I'll summarize it here because it's all spread in the conversation:

We already had a transformer for CSS files in-place before this file so I started investigating why it didn't work in this case.

The problem with the CSS transform is [this line](https://github.com/liferay/liferay-frontend-projects/blob/f97ae3e0a79b7f9e583bf48294a37539d6e8cdfc/projects/npm-tools/packages/npm-scripts/src/config/jest.config.js#L27) that is disabling transforms for any file coming from `node_modules`, which is this case.

[Here](https://github.com/liferay/liferay-npm-tools/commit/12d9bc435bb522e19755ecb45c2047c04eaf30ff#diff-c99ffb3236953a6159ea099c2345c9ffdde98884b5e1488bc35e930145c1eebc)'s the reason why we added that transformIgnorePatterns:

>    Note also that I had to set up transformIgnorePatterns because once we
>    have rewritten own-project references, we want to make sure that we
>    require the files that we actually built (with buildSoy) rather than
>    falling through to our transformSoy stub).

So it's apparently a restriction we imposed to ourselves to make sure we are not using invalid files. 

Then I thought I would try to move that restriction to the transformer so that we can remove the `transformIgnorePatterns`, but realized that the `/node_modules/` transform ignore pattern is part of [jest's default](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring) so I preferred to be conservative and not remove it because who knows what kind of horrible things may happen in the future if we deviate from the norm. 

And that's why I fixed it with a resolver instead of a mapper.

In fact, as I said above it may even make more sense in cases like this where you want to map everything to the same content (n to 1 mapping) because if you do it at the resolver level you skip the resolution + the mapping, which is costly.

Hope that makes sense...

izaera:
I'm merging this since the portal teams need it. Will release a new version.

izaera:
I've released https://www.npmjs.com/package/@liferay/npm-scripts/v/44.0.2

@carloslancha will integrate it in his PR

carloslancha:
https://github.com/liferay-lima/liferay-portal/pull/1824

bryceosterhaus:
@izaera, yep that makes sense about the resolver vs mapper, thanks for the explanation 🎉

PR #540, opened 2021-05-17
feat(npm-scripts): update liferay-theme-tasks version

izaera:
Update to latest liferay-theme-tasks versions to avoid yarn duplication in the repo


izaera:
I'm merging this as it is only a version number increase.

Issue #544, opened 2021-05-17
Create the Liferay CE Portal 7.4.0 GA1 target platform


izaera:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [X] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [X] :two: v2.x
-   [ ] :three: v3.x

### Description
Based on:

- https://issues.liferay.com/browse/LPS-132149
- https://github.com/izaera/fat-jar-sample/issues/1

Create a new target platform for the latest released version, to show how the functionality would work. 

Given that we will have multiple platforms and each one will be a different npm package I think we should create a new directory at the same level as [js-themes-toolkit](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/js-themes-toolkit), or [npm-tools](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/npm-tools) with a `packages` subfolder and one project per target platform.

The following could be used as a naming schema:

- `@liferay/ce-portal-7.4.0-ga1`
- `@liferay/dxp-7.3`

Or anything like that...

nhpatt:
Ideally we should use the same naming scheme as the target plaform in java. I think it's like:
```
dxp-7.3-sp1
dxp-7.2-sp4
dxp-7.1-sp6
dxp-7.0-sp16
portal-7.4-ga1
portal-7.3-ga7
portal-7.2-ga2
portal-7.1-ga4
portal-7.0-ga7
```
With our @liferay/ prefix.

izaera:
:+1:

izaera:
I'm seeing now something we won't be covering... :thinking: 

We have third party dependencies (like `react`, `lodash`, etc.) which we can get from the portal at runtime and from npm at build time. However, we don't publish anything in npm for our internal OSGi bundles (like `frontend-js-web`, `frontend-js-state-web`, etc.) so those cannot be available at build time, only at runtime.

This doesn't have to be a problem per se because they are not needed for the artifacts to be built and deployed, but if someone tries to consume our internal modules for tests or TypeScript typings, for example, this may have consequences in the future so I'm just noting it here so that we all know about it.

nhpatt:
> However, we don't publish anything in npm for our internal OSGi bundles (like frontend-js-web, frontend-js-state-web, etc.) so those cannot be available at build time, only at runtime.

[...]

Issue #543, opened 2021-05-17
Use Babel 7 in target platforms


izaera:
> 👀 Please check the [troubleshooting guide](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/How-to-troubleshoot-your-setups.md) before reporting anything. It contains important information on how to fix or diagnose errors.

### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [X] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [X] :two: v2.x
-   [ ] :three: v3.x

### Description
Based on:

- https://issues.liferay.com/browse/LPS-132149
- https://github.com/izaera/fat-jar-sample/issues/1

Create a new loader to invoke Babel 7 similar to [the current one for babel 6](https://github.com/liferay/liferay-frontend-projects/tree/109e64d1a0394b3262faa61adb991ad224bc8373/maintenance/projects/js-toolkit/packages/liferay-npm-bundler-loader-babel-loader).

But this time create it in `liferay-npm-build-support` as we do for adaptations, to stop extending the number of packages in the JS Toolkit.

izaera:
We may as well invoke babel from the target platform script [just before the bundler](https://github.com/liferay/liferay-frontend-projects/pull/546/files#diff-8cbe038698a422a3fbb09f05c95e2c11d84cf9b9cdf7c1903d66b1f876475d3dR31) if we don't want to add another loader.

In that case, we would change the target script and [remove this rule](https://github.com/liferay/liferay-frontend-projects/pull/546/files#diff-88aa9e0cf5efff761ae7035c7f3a73af7844e64cbd2b3b3dadad63d0061b8ec6R182-R194)

bryceosterhaus:
I agree, I think we should invoke it within the platform script, that way we align closer to npm-scripts for a potential future merge.

izaera:
I have changed the description of the task based on the previous comment.

nhpatt:
+1 to @bryceosterhaus comment, I would call it from the platform script.

PR #546, opened 2021-05-18
feat: create initial target-plaforms projects

izaera:
This is the first PR for the target-platforms feature. It implements #544 partially (we may need to add more features).

I've tested this against https://github.com/izaera/fat-jar-sample/tree/cbe4a1a5166c0f23d71a36a0ae4e118cbda683c2 with a local npm registry and it works but, as I said, we will need more things on top of this, so we can merge it, even if it is unfinished, if we all agree it's the way to go.


izaera:
I've added all people that may be interested as reviewers, but feel free to decline my invitation 
feel free

jbalsas:
> I've added all people that may be interested as reviewers, but feel free to decline my invitation

Didn't invite me

sad

izaera:
Ohohoho! This is wonderful! Because we've added the whole portal as a dependency, `yarn` wants to have all those deps available in my local `liferay-frontend-projects`

ovation

izaera:
> Didn't invite me

Who are you? Are you from Liferay?  :monocle_face:

izaera:
> Because we've added the whole portal as a dependency, `yarn` wants to have all those deps available in `liferay-frontend-projects`

This is clearly not desirable, because as the number of target platforms grows, we will have more and more versions of the same package multiple times. For example: we will have one version for clay as it was in 7.2, another as in 7.3, another for 7.4, and again several for each of those branches, making the repo unmanageable.

The only solution that comes to my mind is to exclude the target platform projects from the yarn workspace. I'm gonna check the implications of that locally, but I believe it's our only possibility :thinking: 

If anyone can think of any other solution, feel free to propose it :+1:

izaera:
I've added some commits that should fix the dependencies issue :crossed_fingers:

projects/target-platforms/packages/create-plaform/bin/create-platform.js
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env node
```
bryceosterhaus: So this file essentially generates a `package.json` and `config.json` to a new directory? And this would be the package we would publish to npm?

btw I am reading this through commit by commit, so things might change in future commits I'm not aware of yet.

projects/target-platforms/packages/create-plaform/bin/create-platform.js
```diff
@@ -82,6 +82,9 @@ const output = {
 		sources: ['assets', 'src'],
 	},
 	pkgJson: {
+		bin: {
+			'liferay-platform-scripts': './liferay-platform-scripts.js',
```
bryceosterhaus: I'm a little hesitant with the naming since we already have `liferay-workspace-scripts` and `liferay-npm-scripts`. Adding another `liferay-*-scripts` might just add some confusion later down the line.

Would doing something like `liferay-platform` make more sense and reduce the chance of confusion? I can't think of a great name at the moment, but I will think about this more... Ideally customers would only have to set this once and whenever they go from 7.4-GA4 to 7.5-GA1, they would just update their dependency.

bryceosterhaus (commented):
> The only solution that comes to my mind is to exclude the target platform projects from the yarn workspace

This makes sense to me, we are really only building artifacts to be published and not consumed anywhere else in the repo.

-----

This looks good so far and I like having to just run a command to generate the platform based on your dxp repo, it makes it much easier to maintain. Because we are dependent on a dxp repo, do we have any safe-guards to make sure it isn't referencing a dirty repo? Also, this makes me think, when are we expecting the target platform to be generated? Do we have to wait for the official GA release before we can run this script? Or will this script be run by bchan somehow whenever we release a GA?


Overall though, this is starting to look good. Great work!

projects/target-platforms/packages/create-plaform/bin/create-platform.js
izaera: I had the same problem and thought about `liferay-platform` too :sweat_smile:, but seemed a bit awkward: 

`liferay-platform build`

Looks like you are building the platform, not *for* the platform...

izaera: In any case, we can merge it, and change it to a better name in the future.

projects/target-platforms/packages/create-plaform/bin/create-platform.js
izaera: Yes, it generates a new platform project next to the place where the script lives. So, the idea is that we run `yarn link` inside the `create-platform` project and then invoke it like:

`create-platform /path/to/your/liferay-portal portal-N.N-gaN`

I didn't put too much safeguards into the script because we will be the only ones running it so I think explaining how it works in the README or CONTRIBUTING files should be enough. Also, it is very probable that we may need to tweak it every time it is run because of little changes to the `liferay-portal` source tree, so I expect it to evolve every time we run it so I consider it more a helping tool than a real stable tool that anyone could use (at least for now).

I was not going to create a script at first, but then I tried to define the platform by hand and it was so complex and error prone that I decided to automate it (especially to avoid human errors).

Additionally, the `create-platform` script contains [assets](https://github.com/liferay/liferay-frontend-projects/pull/546/files#diff-8cbe038698a422a3fbb09f05c95e2c11d84cf9b9cdf7c1903d66b1f876475d3d) that are copied to every plaform it generates. I don't know if this structure will last long, but it suffices for now.

The main goal is to have a centralized point where we can manage these things, lint them, etc, but without needing to maintain another npm package to be shared by all platforms. I think this structure is better because having a shared package for all platforms looks like something quite difficult (given how much we change the build process from version to version) but if we stabilize portal's build enough in the future we can switch to that model.

In any case, the good thing is that everything is kept under the "platform" cover, so we may change anything we like in the future without impacting users.

Right now, the only APIs we are publishing are:

- The name of the platform dependency
- The name of the build script and its API (`build` command for now)

I think those are the only two things we won't be able to change in the future without incurring in a breaking change.

izaera: I've added [this commit](https://github.com/liferay/liferay-frontend-projects/pull/546/commits/25b88ebcdb9f209787da9f49d5a40862af1ddaca) to explain it all.

projects/target-platforms/packages/create-plaform/bin/create-platform.js
izaera: Maybe `liferay` alone? :thinking: 
So that command would be `liferay build`. 

That way you would even be able to do things like: `yarn liferay lint` or `yarn liferay deploy` without having to create anything in the `scripts` section of the `package.json`...

jbalsas: Did a quick industry-wide poll:

| Product | npm cli package | cli binary |
| --- | --- | --- |
| [Netlify](https://docs.netlify.com/cli/get-started/) | `netlify-cli` | `netlify` |
| [Salsesforce](https://developer.salesforce.com/docs/atlas.en-us.232.0.sfdx_setup.meta/sfdx_setup/sfdx_setup_install_cli.htm) | `sfdx-cli` | `sfdx` |
| [Next](https://nextjs.org/docs/api-reference/cli) | `next` | `next` |
| [Vercel](https://vercel.com/cli) | `vercel` | `vercel` |
| [Shopify](https://shopify.dev/tools/cli/installation) | `shopify-cli` | `shopify` |

Based on this, I think ideally we might want to have some workflow like:

```bash
> npm i -g liferay-cli
> liferay build
```

bryceosterhaus: Yeah we aren't really a `*-cli` package since this is used primarily via an npm script. I could imagine a cli package in the future for generating themes or portlets like @izaera mentioned. So maybe we also want to think about that in relation to this. I do like the idea of saving the `liferay` bin for something you globally install. What about `liferay-target`?

| npm cli package | cli binary |
| --- | --- |
| `@liferay/cli` (package TBD) | `liferay new theme` |
| `@liferay/target-7.4-GA1` | `liferay-target build` |
| `@liferay/target-7.3-GA4` | `liferay-target build` |

izaera: We have a naming scheme for the platform packages already, @nhpatt gave us the [pattern to follow](https://github.com/liferay/liferay-frontend-projects/issues/544#issuecomment-842991531), based on how the Java platforms are named.

Regarding the `liferay` binary, my proposal was to:

1. Embed the `liferay` script in the platforms for now. That would allow us to put `"build": "liferay build"` in the `package.json`.
2. When we create the tool, give it a package name (eg: `@liferay/cli`) and make it install a binary called `liferay` too. This way uses can do `liferay new theme` for example.
3. Make `@liferay/cli`'s `liferay` use the local `liferay` in projects that have it, as other standard tools (eg: `lerna`, and I believe `ng`) do, which let you install them globally and locally and the global one delegates to the local one as needed.

This would let us maintain backwards compatibility without the need to release `@liferay/cli` now or make breaking changes in the platforms in the future. I think... :thinking:

target-platforms/packages/portal-7.4-ga1/config.json
```diff
@@ -0,0 +1,197 @@
...
+		{
+			"description": "Copy static assets",
+			"test": "^assets/.*",
+			"use": [
+				{
+					"loader": "copy-loader"
+				}
+			]
+		},
+		{
```
izaera, line 182: Please have a look at this [relevant comment](https://github.com/liferay/liferay-frontend-projects/issues/543#issuecomment-844984684) when reviewing this.

projects/target-platforms/packages/create-plaform/bin/create-platform.js
bryceosterhaus: I like that idea and I think you are right that by creating that binary now we don't actually have to worry about when we make the new package. So I am all for using just `liferay`

izaera: I've pushed more commits to account for the renaming.

izaera:
We still need to decide something about [where to invoke babel from](https://github.com/liferay/liferay-frontend-projects/issues/543#issuecomment-844984684) to polish this feature.

Depending on what we decide we may discard the #543 issue adding a modification to the `liferay` script or leave it as it is now.

izaera:
I have added commits to invoke babel 7 from the target platform and updated the [PoC accordingly](https://github.com/izaera/fat-jar-sample/blob/master/packages/fat-test/package.json#L9)

IINM this should be all we need to run a target platform (for now).

We still would need to add support for extras like `deploy`, `scss`?, etc, but I don't know what's the initial scope of the target platform feature so I considered this PR finished unless some other requests are made.

izaera:
Once the target platform is settled, we need to add support for it in the JS Toolkit generator or create the `@liferay/cli`, maybe.

izaera:
I've created https://github.com/liferay/liferay-frontend-projects/issues/556 to discuss about the project generator.

target-platforms/CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+$ cd ~/liferay-portal ↩
+
+# Fetch all remote tags from the `upstream` remote
+$ git fetch upstream --tags ↩
+
+# Checkout the target platform tag
+$ git checkout 7.4.0-ga1 ↩
+
+# Run yarn in modules folder to update packages at node_modules
+$ cd modules && yarn ↩
```
bryceosterhaus, line 40: We might want to make sure we run `yarn` with dxp's versions just in case there is a mismatch.

We mention something like this nom our npm-tools controbuting doc
```
env PATH=../build/node/bin:$PATH node ../build/node/lib/node_modules/yarn/yarn-1.13.0.js
```

bryceosterhaus (commented):
This looks great! We can focus this PR on `build` for now and add deploy and sass later.

target-platforms/CONTRIBUTING.md
izaera: Don't worry too much about this: the process is so manual that it's almost impossible to do something wrong because, after running the script you can't blindly commit the resulting modifications without reviewing them in Git. In fact, you need to pass SF or lint will complain because of how JSON is formatted.

Think of the `create-platform` as a helper to do 90% of the job you would have to do manually. But, after using it 3-4 times I think it will be very improbable that we can release newer versions without having to do small manual tweaks before.

In any case, I'll add a prompt to show the portal revision and confirm, in the script. Just in case :+1:

izaera: Sorry, I misunderstood the message. You mean using portal's downloaded `yarn` :+1:

bryceosterhaus (approved):
LGTM! I'll let you merge or wait in case you want feedback from anyone else

Issue #542, opened 2021-05-17
Provide some way for the `bundler` to infer a preset based on the configured target plaform


izaera:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [X] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [X] :two: v2.x
-   [ ] :three: v3.x

### Description
Based on:

- https://issues.liferay.com/browse/LPS-132149
- https://github.com/izaera/fat-jar-sample/issues/1

We need to provide some way for the `bundler` to infer a preset based on the configured target plaform.

A possible way would be to make the bundler scan all direct dependencies of the project and, if one of them is a preset, use it if nothing else is configured in `.npmbundlerrc`. We should, of course, fail the build if more than one target platform is found.

That would allow us to implement points 1 and 4 of https://github.com/izaera/fat-jar-sample/issues/1

izaera:
At first glance I though that scanning direct project dependencies for presets would be enough, but that could be a breaking change because now we are applying the default preset when nothing is specified even if the project has preset dependencies.

I know it's very strange that a project has preset dependencies that it doesn't use, but I prefer to be correct than break things outside. We never know what users are doing with the tools, so it's not fair to break things.

For that reason, we will need something to tag autopreset projects so that `liferay-npm-bundler` knows when it is safe to auto-apply them on absence of any preset configuration. Following the same pattern we use in themes, I'll use the `keywords` in `package.json` for that purpose so that only presets tagged as `liferay-npm-bundler` + `autopreset` are taken into account.

I've run a [search for autopreset in npmjs.com](https://www.npmjs.com/search?q=autopreset) and it yields nothing, so it looks like a good tag for this.

izaera:
:thinking: There may be a better way to detect if a preset is an autopreset: if a project has a direct dependency that is itself a preset and that dependency has a direct dependency to `liferay-npm-bundler` (i.e: it brings `liferay-npm-bundler` to the project's build as our target platforms do) then it's an autopreset :tada: .

That doesn't break anything of what we've already released (because our current presets don't have dependencies to `liferay-npm-bundler`) and it's also (almost) impossible that anyone is doing such thing outside, because it would lead to cyclic dependencies. Also, in case they are doing it, chances are that they do it for the same purpose.

[...]

PR #547, opened 2021-05-19
feat(js-toolkit): support for autopresets

izaera:
This implements https://github.com/liferay/liferay-frontend-projects/issues/542

I've tested it locally using a local npm registry for the preset and yarn-linking the bundler in a local copy of [the PoC project](https://github.com/izaera/fat-jar-sample/tree/886597222590a965ab1b641d91cc1bd41d8f2f44) and it works as expected.

In addition I've provided some unit tests to avoid breaking it in the future.

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
@@ -363,11 +363,11 @@ export class Project {
 	}
 
 	_findAutopresets(): string[] {
-		const {dependencies, devDependencies} = this._pkgJson;
+		const {dependencies = {}, devDependencies = {}} = this._pkgJson;
 
 		return Object.keys({
-			...(dependencies || {}),
-			...(devDependencies || {}),
+			...dependencies,
+			...devDependencies,
 		}).reduce((autoPresets, pkgName) => {
 			try {
 				const {dependencies} = this.require(pkgName + '/package.json');
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
@@ -362,6 +362,71 @@ export class Project {
 		}
 	}
 
+	_findAutopresets(): string[] {
+		const {dependencies, devDependencies} = this._pkgJson;
+
+		return Object.keys({
+			...(dependencies || {}),
+			...(devDependencies || {}),
+		}).reduce((autoPresets, pkgName) => {
+			try {
+				const {dependencies} = this.require(pkgName + '/package.json');
+
+				if (dependencies && dependencies['liferay-npm-bundler']) {
+					autoPresets.push(pkgName);
+				}
+			}
+			catch (err) {
+
+				// ignore
+
+			}
+
+			return autoPresets;
+		}, []);
+	}
+
+	_loadDefaultPreset(): string {
+		let presetFilePath;
+
+		const autoPresets = this._findAutopresets();
+
+		if (autoPresets.length > 1) {
+			throw new Error(
+				'Multiple autopreset dependencies found in project (' +
+					autoPresets +
+					'): please remove the invalid ones or ' +
+					'explicitly define the preset to be used in the ' +
+					'.npmbundlerrc file'
+			);
+		}
+		else if (autoPresets.length) {
+			presetFilePath = this.resolve(autoPresets[0]);
+
+			this._toolsDir = new FilePath(
+				path.dirname(this.resolve(autoPresets[0] + '/package.json'))
+			);
+		}
+		else {
+
+			// If no preset was found, use the default one
+
+			presetFilePath = require.resolve(
+				'liferay-npm-bundler-preset-standard'
+			);
+
+			this._toolsDir = new FilePath(
+				path.dirname(
+					require.resolve(
+						'liferay-npm-bundler-preset-standard/package.json'
+					)
+				)
+			);
+		}
+
+		return presetFilePath;
+	}
+
 	_loadNpmbundlerrc(): void {
 		const npmbundlerrcPath = this._configFile.asNative;
 
@@ -374,17 +439,7 @@ export class Project {
 		let presetFilePath;
 
 		if (config.preset === undefined) {
-			presetFilePath = require.resolve(
-				'liferay-npm-bundler-preset-standard'
-			);
-
-			this._toolsDir = new FilePath(
-				path.dirname(
-					require.resolve(
-						'liferay-npm-bundler-preset-standard/package.json'
-					)
-				)
-			);
+			presetFilePath = this._loadDefaultPreset();
 		}
 		else if (config.preset === '' || config.preset === false) {
 
@@ -437,7 +492,7 @@ export class Project {
 	private _configFile: FilePath;
 
 	private _npmbundlerrc: object;
-	private _pkgJson: object;
+	private _pkgJson: {dependencies: object; devDependencies: object};
 	private _pkgManager: PkgManager;
 
 	/** Absolute path to project directory */
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-autopreset/node_modules/autopreset/config.json
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-autopreset/node_modules/autopreset/config.json
@@ -1,0 +1,3 @@
+{
+	"output": "autopreset-build"
+}
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-autopreset/node_modules/autopreset/package.json
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-autopreset/node_modules/autopreset/package.json
@@ -1,0 +1,8 @@
+{
+	"name": "autopreset",
+	"version": "1.0.0",
+	"main": "config.json",
+	"dependencies": {
+		"liferay-npm-bundler": "*"
+	}
+}
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-autopreset/package.json
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-autopreset/package.json
@@ -1,0 +1,7 @@
+{
+	"name": "with-autopreset",
+	"version": "1.0.0",
+	"dependencies": {
+		"autopreset": "1.0.0"
+	}
+}
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/.npmbundlerrc
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/.npmbundlerrc
@@ -1,0 +1,3 @@
+{
+	"preset": "preset"
+}
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/node_modules/autopreset/config.json
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/node_modules/autopreset/config.json
@@ -1,0 +1,3 @@
+{
+	"output": "autopreset-build"
+}
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/node_modules/autopreset/package.json
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/node_modules/autopreset/package.json
@@ -1,0 +1,8 @@
+{
+	"name": "autopreset",
+	"version": "1.0.0",
+	"main": "config.json",
+	"dependencies": {
+		"liferay-npm-bundler": "*"
+	}
+}
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/node_modules/preset/config.json
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/node_modules/preset/config.json
@@ -1,0 +1,3 @@
+{
+	"output": "preset-build"
+}
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/node_modules/preset/package.json
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/__fixtures__/project/with-ignored-autopreset/node_modules/preset/package.json
@@ -1,0 +1,5 @@
+{
+	"name": "preset",
+	"version": "1.0.0",
+	"main": "config.json"
+}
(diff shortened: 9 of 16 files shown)


maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -362,18 +362,55 @@ export class Project {
...
+					autoPresets +
+					'): please remove the invalid ones or ' +
+					'explicitly define the preset to be used in the ' +
+					'.npmbundlerrc file'
+			);
+		}
+		else if (autoPresets.length) {
+			presetFilePath = this.resolve(autoPresets[0]);
+
+			this._toolsDir = new FilePath(
```
jbalsas, line 406: Not sure how much logging we're doing, but it might not be a bad idea to somehow log (here or elsewhere) which preset is being used and if relevant, why. Might save us and customers from some head scratching in the future.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -362,18 +362,55 @@ export class Project {
...
-		if (config.preset === undefined) {
+		const autoPresets = this._findAutopresets();
+
+		if (autoPresets.length > 1) {
+			throw new Error(
+				'Multiple autopreset dependencies found in project (' +
+					autoPresets +
+					'): please remove the invalid ones or ' +
+					'explicitly define the preset to be used in the ' +
+					'.npmbundlerrc file'
```
jbalsas, line 400: I would soooo love if we could get rid of `.npmbundlerrc`... 🙏

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: All those details are shown in the report file (we decided to dump all relevant build information there long ago, since the logging was unmanageable), but it's true that we don't write the name of the preset being used, just the resultant configuration, so I'll add it too :+1:

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: What do you mean? We can build projects without `.npmbundlerrc` already, as long as the user is satisfied with the default values. In fact this PR makes it even easier to have projects without `.npmbundlerrc` because you can create an autopreset for that :tada:

izaera: Before the autopreset only our default preset was "magic", but that was not enough for most of the projects.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
jbalsas: > [...] we decided to dump all relevant build information there long ago [...]

I dispute that. Especially the "**all**" portion of it. I think we should revisit that decision. It's fine to drop a massive amount information in a report that customers know to activate and navigate or not. I don't think that's contrary to making our output friendlier.

If we're going to make relevant decisions for the users, it wouldn't hurt to let them know in advance. Compare the following workflows after an error:

// Workflow 1
- Check log, see "Compiling with preset liferay-ce-ga1-7.2"
- Oops... I'm building for 7.3
- Update it

// Workflow 2
- Check log, see nothing
- Edit `.npmbundlerrc` to add `report: true`
- Build again to generate a report
- Locate the report.html file and open it
- Scan it to find "Compiling with preset liferay-ce-ga1-7.2"
- Oops... I'm building for 7.3
- Update it

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
jbalsas: I wish we didn't have a thing that's `.npmbundlerrc` specially if we build more stuff and then it starts overlapping with `npmscripts.config.js` and eventually others. I just get the feeling there are way too many config marker files all over the place and their names aren't really descriptive/unique to understand what should each of them do.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: I like the idea of having output. I always thought that not logging anything wasn't a good idea, but we decided to remove everything in the past because we didn't want any message to show up in the console.

We also have a [verbose](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/.npmbundlerrc-file-reference.md#verbose) flag but it can only be passed in the `.npmbundlerrc` (and not as an argument, like we can do for the report and the JAR file creation) so it's a bit ~complex to use~ useless.

We may:

1) Dump some output by default setting the `verbose` flag's default value to (now it is `false` by default). This is what webpack does, IINM.
2) Don't show anything and provide a CLI switch to activate verbosity.
3) Other thing...

I would go with 1 since it's the most usual way of working for tools like this...

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: And where would we put that information? I find having a tool like this one (a bundler) without any configuration almost impossible :thinking: 

We also need files like `.babelrc`, `.eslint`, `webpack.config.js`, etc. to build projects. Unless we make something like `npm-scripts` that abstracts you from all that, I can't see the solution (even if we get rid of `.npmbundlerrc`, because we still have the other files).

And, if we make an umbrella tool like `npm-scripts` then, what's the problem with having a `.npmbundlerrc` file created under the hood as we do in portal? It's transparent to the user...

One idea I had long ago was to make a programmatic API for the bundler so that a higher level tool could configure and call it without the need for any file (like you can do with Babel, for example). It shouldn't be that difficult because, since we introduced the `Project` class for abstracting the configuration, the bundler's core is agnostic of the `.npmbundlerrc` (it can come from any other place we like).

izaera: But even in that case, as I said, we need some other files for eslint, webpack, etc... :-(

Maybe there's some way to internalize them inside the bundler like we can do for babel (you can put babelrc inside npmbundlerrc) but I don't think it's that useful. And it is probably an abuse of the bundler if we want to specialize it in just bundling.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
bryceosterhaus: > but we decided to remove everything in the past because we didn't want any message to show up in the console.

Maybe we could slowly move towards logging simple messages? There is definitely a fine line between helpfulness and noise, but I do enjoy small one-liners that let me know something happened.

For example, here is a Next.js server.
Screen Shot 2021-05-19 at 11 18 12 AM

I really like how they have statuses for messages like "Event" and "Wait". Now this is different because its a live server, but I gain confidence as a developer when I see information like that.

In our case here, I like the idea of something as simple as `JS Bundler: Compiling with preset liferay-ce-ga1-7.2` and then we can use the verbose flag to dump more granular details of the build. We could just add a line somewhere in this output 
Screen Shot 2021-05-19 at 11 23 02 AM

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
bryceosterhaus: > there are way too many config marker files all over the place and their names aren't really descriptive/unique

Do we have a list of what all these would be? Our own homebrewed ones are `.npmbridgerc`,  `.npmbundlerrc` and `npmscripts.config.js` right? Third party ones being eslint, prettier, webpack, babel, 

@jbalsas is your concern due to having the extra file, the name of it, or just the config in general? Or all the above? 

- We could move config into the package.json to avoid an extra file. 
- Naming is hard and especially in a project as large as DXP.
- I don't think it'd be possible to throw out the config altogether, we can provide really good defaults, but there will always be cases that we need to account for that defaults won't satisfy.

Future thought here, would it be possible to converge our configs to be more like `npmscripts.config.js`? If we aim for a singular config files, we might be able to reduce the amount of confusion between each of them. Or at least we could possible aim for a singular config for liferay js related things.

nhpatt: > we need some other files for eslint, webpack...

But those are standard, so JS developers or even Java developers expect them. The issue is ours, those are the ones that we need to avoid or make them as descriptive and simple as possible.

bryceosterhaus: Yep! Totally agree, @nhpatt.

izaera: > We could move config into the package.json to avoid an extra file

I don't like this one too much because `package.json` files are published to npmjs.com and there's always the risk of leaking something in the config. I understand that some of the config may be useful for the consumers of the packages, but some other is better left unpublished, I think...

izaera: Regarding the other files, we have:

- .npmbridgerc
- .npmbundlerrc 
- npmscripts.config.js

Additionally, there's the possibility to embed some other standard files inside .npmbundlerrc. For example, it is possible to embed a .babelrc file inside the .npmbundlerrc (this behaviour is copied from webpack). However, we can also allow to have a standalone .babelrc file next to the .npmbundelerrc if we want. All we need to decide is what we want.

In general, as I said, the bundler uses a .npmbundlerrc file (like any other standard tool, like webpack). That file is fully documented in [our docs](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/.npmbundlerrc-file-reference.md) and it is totally optional because:

1. Every setting has a default
2. If the file is missing, we apply the standard preset automatically (and now, if we merge this, any autopreset that the tools finds configured in the project).

Said that, it's impossible for some projects to have no .npmbundlerrc file because there are some things that are simply impossible to infer (e.g: bundler imports). For that cases, you could create an autopreset (like our target platforms) that configure the bundler without any need to have a .npmbundlerrc.

We also need to consider the **divergence between portal's build and customer's builds**, which are quite different, but the bundler needs to account for both. For example:

- Portal: the JAR is created by gradle, uses src/main/... folders, has npm-scripts on top
- Customers: the JAR is created by blade or the bundler itself, doesn't use src/main/... folders

Also, from the portal's POV, the bundler is something we want to hide, but from the customers' POV the bundler is really a tool. Even if we don't want to expose them to it, there are several customers using it for their own workflows, built on top of it. So, in that regards, it is similar to webpack.

I understand that target platforms are intended to work like npm-scripts (hiding the bundler as much as possible), but we still have the other use case, where advanced users need the bare tool.

----

In any case, ~can we create an issue to track this and decide there exactly what would be ideal for us?~ I've created [an issue](https://github.com/liferay/liferay-frontend-projects/issues/551) to track this. We can add the `liferay-cli` proposal there, and the final layout the different develper roles (portal, customer, advanced customer) would expect.

Regarding this PR, it doesn't change the model of configuration we have now. If any, it enhances it because it goes in the direction of getting rid of the npmbundlerrc.

izaera: The `.npmbridgerc` file is for a totally different tool that simply creates JS modules that re-export other JS modules. This was created to maintain backwards compatiblity of very old JS modules we had in portal (pre-bundler stuff that used the AMD loader without any versioning of npm packages present in their URLs).

Maybe it is better if we remove all that (I think it could be possible by now) instead of finding a way to hide the `npmbridgerc`. We haven't added more bridges (because they are not really needed) and we wouldn't want to add them :sweat_smile: . 

In the worst case scenario we could always generate the current bridges we have in portal, check them into git and delete the bridge creator tool forever.

This is similar to what we did in the `frontend-js-node-shims` project (checking in hand made custom packages as if they had been built by the bundler, instead of complicating the bundler to build them) and it has gone well: it hasn't given any significant problems, and we are now in a position where we can remove `frontend-js-node-shims` in the future if we want. And all that without having impacted the JS Toolkit.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: > Maybe we could slowly move towards logging simple messages? There is definitely a fine line between helpfulness and noise, but I do enjoy small one-liners that let me know something happened.

That's [already done](https://github.com/izaera/liferay-js-toolkit/commit/02e8806465f90cab886f61f6e0ebab9ae10c2811). The only caveat is that, by default [we set `verbose` to `false`](https://github.com/izaera/liferay-js-toolkit/commit/02e8806465f90cab886f61f6e0ebab9ae10c2811#diff-9235b7a19f38965392ca7ba4ffb6347143a9ebda3b395fc93879f003d788f1baR177) and, to change it, you need to create a .npmbundlerrc file and set it to `true`.

So, my proposal was:

1. Create a `verbose` command line switch as we have for [--dump-report](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/.npmbundlerrc-file-reference.md#--dump-report) (that could be specified both in the command line and in the npmbundlerrc file)
2. Change `verbose`'s default to `true` so that we log some information about what's happening (as we did in the past, before `verbose` was introduced)
3. Add a log message for the preset in use (if any)
4. Refine any other messages we are logging already

I will create a [new issue](https://github.com/liferay/liferay-frontend-projects/issues/552) for this, since it can be done detached from this PR.

izaera: > use the verbose flag to dump more granular details of the build

That's what the report file does. Due to the complexity of the bundler builds, where many things can go wrong deep inside third party packages when we find weird code, we create a HTML file on user's request to explain what is happening under the hood and what decisions are taken when building the project.

I think it is better to simply:

1. Log essential messages by default as proposed in the issue I've created
2. Let users disable verbosity so that nothing is printed in the console: this is useful for cases like liferay-portal where we don't want to print anything (the reason why the [commit linked above](https://github.com/izaera/liferay-js-toolkit/commit/02e8806465f90cab886f61f6e0ebab9ae10c2811) was merged), or similar projects where the bundler is called multiple times and you don't really care about the output in the full log, only on a per-project basis, when anything goes wrong.
3. Dump any other info to the report file. 
 
The report file is one of the first things [we ask users for](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/How-to-troubleshoot-your-setups.md) when it comes to support them, and it has helped several people to understand a bit better what the bundler does under the hood, so I think it's worth having it.

We could even increase its functionality with things like [this table of detailed explanations about bundler messages](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/Report-messages.md) which the bundler report files link directly.

izaera: Of course, as @jbalsas says in his workflow 2, essential things should go in the log and report should only be used for things that can break.

However, it is sometimes unclear what essential is, because in that example, we consider the preset essential because it leads to user confusion. But that's only because of the way the case is explained that it's centering the cause of the error in the target platform (something that the user will probably not change so often and not very error prone IMO): 

I mean, what are the chances that you need to change the target platform? And in that case, what are the chances of configuring a bad platform? We are assuming that it will happen often and thus we are showing a message log in the console. But we can apply the same reasoning to anything that could go wrong.

I mean, whenever you encounter an error you wish the tool/framework you are using would notify you clearly about that if it didn't, but most of the times that's not the case :shrug: 

On the other hand, if you clog the console with "useful" log messages that hint you about possible errors (aka: warnings) there comes a time when it stops being useful and it's like having nothing.

So, I lean more toward to what @bryceosterhaus says: 

> I gain confidence as a developer when I see information like that.

So, I propose not to focus on what is needed in the log to avoid problems, but more on what we would like to see as developers to get a feeling of what's going on.

To finish: in this case I would like to see the autopreset in the console because it's my target platform, and I would also like to see the typical finalization message as well as some intermediate messages to know what the tool is doing (for example: one per bundled package or something like that) and know if it stuck, for example.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -362,18 +362,55 @@ export class Project {
...
 
-	_loadNpmbundlerrc(): void {
-		const npmbundlerrcPath = this._configFile.asNative;
+	_findAutopresets(): string[] {
+		const {dependencies, devDependencies} = this._pkgJson;
+
+		return Object.keys({
+			...(dependencies || {}),
+			...(devDependencies || {}),
+		}).reduce((autoPresets, pkgName) => {
```
bryceosterhaus, line 371: You could use default values when destructuring.
```suggestion
const { dependencies = {}, devDependencies = {} } = this._pkgJson;

return Object.keys({
  ...dependencies,
  ...devDependencies,
}).reduce((autoPresets, pkgName) => {

```

bryceosterhaus (approved):
Sorry, I don't think I actually reviewed the code before.

It looks good, I just had one small nitpick on readability. Feel free to merge

PR #548, opened 2021-05-19
docs(gh-jira-sync): update exalate user mapping adding headless squad

brunofernandezg:
Just updated the user mappings (GH-Jira) in frontend-projects and Clay Jira-Exalate sync scripts.

### DIFF
--- a/support/github-to-jira-synchronization/clay.md
+++ b/support/github-to-jira-synchronization/clay.md
@@ -156,10 +156,10 @@ def findUser(replicaUser) {
         ,"pat270": "patrick.yeo"
         ,"kresimir-coko": "kresimir.coko"
         ,"markocikos": "marko.cikos"
-		,"nhpatt": "javier.gamarra"
-		,"javierdearcos": "javier.dearcos"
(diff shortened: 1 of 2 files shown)


brunofernandezg:
At first I was going to wait to fix the problem with issues sync to do those changes, but I think this is an easy change that would be perfect to test the update process for the first time. As soon as is merged I'll open an IS ticket to update the Scripts in Jira.

Although probably headless squad members won't work on Clay repo for now, I think is better to have the same user map for all scripts so it's easier to maintain.

bryceosterhaus (changes requested):
I think there might be some whitespace off here. Take a look at the rendered output
Screen Shot 2021-05-19 at 11 02 58 AM

brunofernandezg:
Thanks @bryceosterhaus, you are right. I didn't realize tabs were coded as four white spaces 🤦
Fixed

bryceosterhaus:
@brunofernandezg github's whitespacing can be atrocious.

Greg better fix this

PR #558, opened 2021-05-25
docs(gh-jira-sync): update exalate user mapping sort order (alphabeti…

brunofernandezg:
…cally)

Exalate scripts usermappings are now alphabetically sorted


bryceosterhaus:
🥇

Issue #520, opened 2021-04-30
Autorun Prettier over generated files


wincent:
### Issue type

-   :gift: Feature request

### Description

I am pretty sure we've talked about doing this one in the past but hadn't taken the trouble to actually implement it because the publish workflow tends to work well enough in most cases _and_ we've already taken some care to make our output match Prettier's (eg. see [\#230](https://github.com/liferay/liferay-npm-tools/pull/230)).

Yesterday @jonmak08 found another example that produces formatting that Prettier considers to be broken ([Slack thread](https://liferay.slack.com/archives/C3JBR21HA/p1619738723493600)); specifically, the changelog had:

```markdown
-   fix(js-themes-toolkit): npm package for liferay-frontend-theme-* needs...
```

instead of:

```markdown
-   fix(js-themes-toolkit): npm package for liferay-frontend-theme-\* needs...
```

Now, I don't think we want to get too deep into the business of formatting the bullet points in these changelog entries, because it can get complicated real fast. For example, if somebody ever had `*blah*` in a pull request title, intending to produce emphasis, we could easily mistakenly escape that to `\*blah\*`.

So, I think the best approach is just to run Prettier on the output as a post-step. We could make it so that if ever run on a project where Prettier is not already present, we just skip the step. But if it is there, we can use the Prettier API to format just the new entries and avoid touching the rest of the file, which could conceivably produce a bit of churn.

We're already using [a similar trick over in the `js-toolkit`](https://github.com/liferay/liferay-frontend-projects/blob/bd6092ecb6b0398cdf717205d16b8225fec07d03/maintenance/projects/js-toolkit/scripts/publish.sh#L32-L34) (the implementation is very different there, because it's a Bash script, and an internal-only one, but the core idea of using Prettier in a post-processing step is the same).

wincent:
> I am pretty sure we've talked about doing this one in the past

Haha, found this ancient item in my personal TODO app:

<img width="387" alt="Screenshot 2021-04-30 -153046-uZ9tyFGC@2x" src="https://user-images.githubusercontent.com/7074/116702248-26d5d280-a9c9-11eb-9897-87b40d833bb2.png">

PR #559, opened 2021-05-25
feat(npm-tools): Autorun Prettier over generated CHANGELOG.md file

ambrinchaudhary:
Fixes https://github.com/liferay/liferay-frontend-projects/issues/520

Opening as Draft to receive feedback. What I understood from the issue is that we should run prettier once the `CHANGELOG.md` file is generated. 

I know there are some things missing like: 

> if ever run on a project where Prettier is not already present, we just skip the step.

or that maybe I could have reused the `run` function defined in `git.js`, but first I wanted to know if the approach is the correct. I may be totally wrong....so thanks in advance for all the feedback you could provide me. 

**Steps to reproduce:** 

1. Go to `liferay-frontend-projects/projects/npm-tools/packages/changelog-generator` directory
2. Run `node bin/liferay-changelog-generator.js --version=x.y.z --force --no-update-tags --regenerate`


projects/npm-tools/packages/changelog-generator/src/command.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const {exec} = require('child_process');
+const {error} = require('./console');
+
+function runCommand(command, ...args) {
+	exec(command, (err, stdout, stderr) => {
```
izaera, line 10: you will need to use `spawnSync` here because otherwise the script won't wait for the `prettier` process to finish

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -602,6 +603,8 @@ async function main(_node, _script, ...args) {
 
 	try {
 		await go(options);
+
```
izaera, line 606: I would add this step to the `go` function, probably, since it seems to contain all the logic of the changelog generator.

However, I see now that the `updateTags` logic has been extracted from `go`, so... :shrug:

bryceosterhaus (commented):
Hey @ambrinchaudhary, I think this is a good start so far. One thing we need to think about is whether we want to run prettier via cli command or importing the dependency and using its API directly. [Here is an example of us using it directly in our npm-scripts package.](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/packages/npm-scripts/src/prettier/index.js#L119)

Now one thing we need to make sure is that the formatting will pick up the proper prettier config wherever the changelog-generator is used. So it may be advantageous to spawn the CLI command directly since I think it will automatically pick up the config. Maybe you could test both ways of using prettier and see if there is an easy way to pick up the config options for the node API? The benefit of using the node API directly is that we won't have to write the file multiple times, we can just run prettier over the content before writing the file.

As for where this logic goes, it might make sense to be in the `write()` function, or possibly right before we use that function

bryceosterhaus:
Also, whats with these other files?

Screen Shot 2021-05-25 at 12 49 55 PM

izaera:
>  run prettier via cli command or importing the dependency and using its API directly

Mmm, in fact we have the [workspace-scripts](https://github.com/liferay/liferay-frontend-projects/tree/master/support/packages/workspace-scripts) that are supposed to be called to setup the configuration and invoke the tools uniformly across all projects in this repo.

Also, they are invoked by the CI to run everything in every project.

We should give this a thought, maybe, because I've seen instances where you may want to run `format` or `test` (for speed) for a single project/file and it's not possible or doesn't work quite well at all...

The key thing here is that we may want to use the same scripts/configurations for all projects in the monorepo, and not let every project do whatever it wants. But at the same time we don't want to run targets for the whole repo all the time. So, a balance...

This :point_up: probably makes it better to invoke prettier by executing it rather than directly invoking it (unless we provide a programmatic API for `workspace-scripts`, I guess...)

izaera:
So, instead of running `yarn run prettier` directly [in this line](https://github.com/liferay/liferay-frontend-projects/pull/559/files#diff-591e31b6de1c00894f912b639e3f3ff1c0dc4742a505fc1e0690e707184bc2f5R607) we should maybe run `liferay-npm-scripts prettier` as [in here](https://github.com/liferay/liferay-frontend-projects/blob/master/support/packages/workspace-scripts/src/index.js#L46).

TBH, I've never liked this scheme too much. I like the idea of having an uniform configuration and processes, but I would also like to make integration with IDEs or different developer workflows easier... :thinking:

bryceosterhaus:
@izaera, I don't think the changelog-generator is used in the same context as npm-scripts. This generator is primarily used for our frontend projects and so I think we would want prettier to be run from the generator itself. And including npm-scripts in the generator seems like overkill. The README states it is used in a few places, both in this repo and in other repos, so I think we would want prettier to be included with the generator.

Used in...
-   [Alloy Editor](https://github.com/liferay/alloy-editor) ([CHANGELOG](https://github.com/liferay/alloy-editor/blob/master/CHANGELOG.md)).
-   [@liferay/eslint-config](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/eslint-config) ([CHANGELOG](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/eslint-config/CHANGELOG.md)).
-   [liferay-ckeditor](https://github.com/liferay/liferay-ckeditor) ([CHANGELOG](https://github.com/liferay/liferay-ckeditor/blob/master/CHANGELOG.md)).
-   [liferay-js-themes-toolkit](https://github.com/liferay/liferay-js-themes-toolkit) (see per-package changelogs in [`packages/*` subdirectories](https://github.com/liferay/liferay-js-themes-toolkit/tree/master/packages)).
-   [liferay-npm-tools](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/npm-tools) (see per-package changelogs in [`packages/*` subdirectories](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/npm-tools/packages)).

From [the issue itself](https://github.com/liferay/liferay-frontend-projects/issues/520), I think this makes sense to approach it this way.
> So, I think the best approach is just to run Prettier on the output as a post-step. We could make it so that if ever run on a project where Prettier is not already present, we just skip the step. But if it is there, we can use the Prettier API to format just the new entries and avoid touching the rest of the file, which could conceivably produce a bit of churn.

izaera:
> I don't think the changelog-generator is used in the same context as npm-scripts. This generator is primarily used for our frontend projects

I see. So it's used in more projects that are outside of this monorepo...

In that case, I'm not sure if running `prettier` will be enough because we've already seen some scenarios where `prettier` alone "fights" against `npm-scripts` when [it comes to linting](https://liferay.slack.com/archives/CNBG06JS3/p1622034766037700). So I'm afraid we can have the same issue with `prettier` now, or in the future.

The reason behind this is the following: this monorepo uses `npm-scripts` to [format things](https://github.com/izaera/liferay-frontend-projects/blob/cbb539ab0a45d19f9679eb94149516f3da70f2c8/support/packages/workspace-scripts/src/index.js#L46-L54), while others (the one you linked) use `prettier` for that.

Because our `npm-scripts` carry ["hidden" base configurations](https://github.com/izaera/liferay-frontend-projects/blob/master/projects/npm-tools/packages/npm-scripts/src/config/prettier.json) which `prettier` won't see if run directly (as in `yarn run prettier` or `npx prettier` from a project folder of this monorepo), the two tools may format things differently.

Thus, if the changelog runs `prettier` it may or may not fix formatting problems in this repo. It all depends on what's to be fixed and what the "hidden" rules are in a specific moment of time.

So, what can we do about this? If you think about it, the root cause of the problem is that `npm-scripts` can be considered as if it was a tool totally different from `prettier`. That applies to this scenario, IDE integrations, ... Based on that assumption, the possible solutions would be:

1. Stop using `npm-scripts` to format this repo
2. Make the changelog invoke some abstract task like `yarn format` and delegate the formatting to the project (whatever tool it uses)
3. Use `npm-scripts` to format every repo out there
4. ...

I would love to see 1 happen, because that would allow to integrate this monorepo with IDEs more easily and would remove a dependency on an artifact that is part of the monorepo itself. It would also decouple us from the portal.

Additionally, I would fix the bug by implementing 2 so that each project decides how to format things and we don't introduce any assumption about that in the changelog generator. Though I may understand we want such assumptions to be in-place to enforce using them, in which case, we wouldn't implement 2.

izaera:
One more thing: maybe for the case of this issue, it's enough to run `prettier` even if we have the "hidden" configuration in `npm-scripts` because the resultant configurations are compatible enough (in the end, `npm-scripts` is using `prettier` so it's not a totally different tool, as I said, but a semi-compatible tool) to format the changelog files without causing CI issues in this monorepo.

However, I would love to see these issues fixed some day so that we can integrate these projects with different developer workflows/IDEs without too much hassle.

izaera:
> I'm not sure if it is better to run yarn format as suggested by @izaera and delegate the formatting to the project...

I would say it's formally better, but being practical, if invoking `prettier` does it, we may go with that, for me.

bryceosterhaus:
> this monorepo uses npm-scripts to format things



Whoa, I did not realize this. From a macro level, that seems really odd because the aim of npm-scripts is to be used in liferay-portal.

> 1. Stop using npm-scripts to format this repo
> 2. Make the changelog invoke some abstract task like yarn format and delegate the formatting to the project (whatever tool it uses)

I like this direction and for the sake of this issue specifically, we can focus on number 2. And since 1 isn't causing us direct issues, other than being sort of inconsistent, I think we can just go with it for now until we have issues.

So, @ambrinchaudhary, maybe you can investigate what we would need to add for that? My guess is that we would need to add logic to check if there is `format` script in the package.json of the package you are running this in and then run that script after writing the CHANGELOG.

ambrinchaudhary:
Ok, then, things I have changed in the pull: 

1. Using `spawnSync` instead of `exec`. 
2. Running `yarn format`, so we delegate the formatting to the project which runs the generator, and only if there is a `format` script in the _package.json_ . The only bad(?) thing here is that with yarn format you cannot choose a particular file to format (only the CHANGELOG.md) file, so it may format other files I guess?.... or am I wrong?
3. Format it in the `write()` function. 

Any further discussion is most than welcome ;-)

izaera:
> The only bad(?) thing here is that with yarn format you cannot choose a particular file to format (only the CHANGELOG.md) file, so it may format other files I guess?.... or am I wrong?

Yes but:

1. That'll be good, because if some other file is not formatted, then you are fixing it
2. It should be impossible, because you are running the changelog generator inside the release process, which cannot be performed unless the CI is green (and the CI won't be green unless everything is properly formatted).

If this were a tool for public use things could change, because we cannot make assumption 2, but since we are using it only internally, I would say we're safe.

projects/npm-tools/packages/changelog-generator/src/yarn.js
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const child_process = require('child_process');
+const fs = require('fs');
+
+const {error, info} = require('./console');
+
+const pjson = JSON.parse(fs.readFileSync('package.json'));
```
izaera, line 11: We don't use abbreviatures, so `packageJson` (or sometimes, I use `pkgJson`). But in general we don't like abbreviatures (I think there's something on that on the guidelines...).

izaera:
I left [this feedback](https://github.com/liferay/liferay-frontend-projects/pull/559#pullrequestreview-672021688). Other than that, LGTM :+1:

projects/npm-tools/packages/changelog-generator/src/yarn.js
ambrinchaudhary: Changed and pushed, thanks for your feedback @izaera!

wincent:
> Whoa, I did not realize this. From a macro level, that seems really odd because the aim of npm-scripts is to be used in liferay-portal.

FWIW, it's kind of explained in [the original PR](https://github.com/liferay/liferay-frontend-projects/pull/146):

> This also means that we'll use the exact same formatting style in this repo as we use in liferay-portal.

ie. consistency... it was super annoying to be writing

```
}
else {
```

in Liferay's most important repo, but not in our other projects.

One reason I _didn't_ write down at the time, but was thinking, was that by "dog fooding" the formatting in the monorepo, we would catch and notice issues before they got to `liferay-portal`, and that's great because updating `@liferay/npm-scripts` in `liferay-portal` requires a release, which is annoying.

> One thing we need to think about is whether we want to run prettier via cli command or importing the dependency and using its API directly.

When I wrote the initial issue, I was thinking about using the API directly. Kind of like this (pseudo-code):

```js
try {
   const prettier = require('prettier');

   // call API here...
}
catch {
   // Skip formatting; somebody is using the changelog generator
   // in a project that doesn't have Prettier.
}
```

Note that if Prettier isn't available, it's no big deal. Degrade gracefully. The worst that can happen is that the changelog looks just like it did without this PR. Throwing an error isn't very friendly, and is technically a breaking change.

Either way, whether forking out to `yarn` or `prettier`, or using the API directly, I _wasn't_ thinking of formatting every file in the repo like this PR does; rather, just running it over the changelog itself (obviating @izaera's concern about how the format task runs over the whole repo).

Anyway, I'm going to go back to my other life now. 👋

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -902,6 +903,8 @@ async function write(options, preview, contents) {
 	else {
 		await writeFileAsync(options.outfile, contents);
 	}
+
+	yarn('format');
```
wincent, line 907: Note that this line runs unconditionally, bypassing the `--dry-run` flag. Not a huge deal, but if a tool mutates that filesystem when running with `--dry-run`, user will probably interpret it as a bug.

izaera: Good catch :+1:  (who's this guy, BTW? :thinking: )

ambrinchaudhary: Fixed!! My fault, I didn't realise that when `--dry-run` option is executed we do not write the file, but just show a preview. 
Thanks @wincent and @izaera!

bryceosterhaus:
@ambrinchaudhary looks good! Only thing I noticed is that we probably need to wrap the format call in a try/catch. If the project doesn't have a format script, we want it to fail gracefully and not complain that a format script doesnt exist.

ambrinchaudhary:
> @ambrinchaudhary looks good! Only thing I noticed is that we probably need to wrap the format call in a try/catch. If the project doesn't have a format script, we want it to fail gracefully and not complain that a format script doesnt exist.

Hi @bryceosterhaus, I think that is already covered [here](https://github.com/liferay/liferay-frontend-projects/pull/559/files#diff-c5cda8d4aad811628a53eedb10f525f7cc0c744e1d8dc7a6b7d97763fcb1904aR14). If the command does not exist, it won't run (and better than adding try/catchs for each yarn command we may want to run). And I think this fits with what is detailed in the original issue: 

> We could make it so that if ever run on a project where Prettier is not already present, we just skip the step.

So there won't be any crash if the format does not exist ;-).

bryceosterhaus:
@ambrinchaudhary ah yes, I guess I need to go get my eyes checked 😆 . Thanks for clarifying!

PR #560, opened 2021-05-25
chore(@liferay/cli): create project skeleton

izaera:
This is a small PR to find a place for our future beloved `@liferay/cli` before starting its development.

I first tried to put it at `maintenance/js-toolkit` (old toolkit v2 workspace), but we cannot do that because:

1. The old toolkit doesn't conform to the current monorepo standards, so it doesn't seem wise to start a project based on old standards that will need to be tweaked in the future, to move it to its final destination (which is the `js-toolkit` outside of `maintenance`, i.e., the workspace where bundler 3 now lives in).
2. The old toolkit is based on `lerna` and only allows per-worskspace releases. We want to be able to release `@liferay/cli` detached of any other currently existing artifact, so we may not put it there.

Please, provide any feedback you think may be relevant and/or explain what other place could be better before we begin to write it. Thx.

Note that I've also removed some leftovers from the bundler 3 initiative. They are not needed and simply contain copies of what's in `maintenance/js-toolkit` so we may safely remove them. The initial idea was to migrate them once bundler 3 was finished, but since we stopped that, they were left stale.


bryceosterhaus:
I think having it in `projects/js-toolkit/packages/liferay-cli` makes the most sense to me, any other location feels slightly off.

> We want to be able to release @liferay/cli detached of any other currently existing artifact

Yeah this seems important to me, ideally we don't have to release cli and all its dependencies at once.

izaera:
So, I'll merge this to keep going. We can always move it to a different place if we deem it necessary in the future.

Issue #556, opened 2021-05-24
Create a project generator for target platforms


izaera:
> 👀 Please check the [troubleshooting guide](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/How-to-troubleshoot-your-setups.md) before reporting anything. It contains important information on how to fix or diagnose errors.

### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [X] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [X] :two: v2.x
-   [ ] :three: v3.x

### Description
We need to decide if we are going to add target platforms to the current Liferay JS Generator, or create the new `@liferay/cli` tool and start migrating things from Liferay JS Generator to it.

Note that, we may also want to stop using Yeoman in `@liferay/cli`, which seems logical. Given that the current Liferay JS Generator was written to avoid excessive coupling with Yeoman, this shouldn't be very difficult.

nhpatt:
I would create a new @liferay/cli instead of adding it.

-- comments link to this same issue

bryceosterhaus:
I'm fine with either migrating the js generator to a new package, or we could also just use the js generator from our new package. The benefit of migrating it over would be less publishing and versions to worry about and since it is a generator, I don't think people are really going to be leveraging it in a custom way like they would the bundler.

izaera:
What do you think is better?

1. A single command, like `liferay new my-app`, to create a new project with sensible defaults (for now we would only use the one and only platform we have)
2. One command per different target, like `liferay new react my-app`, as our generator currently does

I would go with 1 and then -maybe- provide another command in `@liferay/cli` to switch target platforms, so that the user can, for example, say:

```sh
$ liferay new my-app
...
Created a new JS project for target platform 7.4-GA1

$ liferay target 7.3
Switched target platform to 7.3

$ liferay target react
Switched target platform to React
```

However, I don't know how difficult it would be to switch from -say- `7.4-GA1` to `react` because the configuration of the project can be quite different.

Also, it maybe doesn't make too much sense to create a project for `7.4-GA1`, then switch it to `react`, then switch it to `angular`.

An option that comes to mind is to simply refuse to switch to an invalid/incompatible target platform, or something like that. But still we have the case where the user may want to use Angular instead of React and `liferay new` is creating a React based project.

izaera:
[...]

PR #562, opened 2021-05-27
feat(liferay-cli): support for "liferay new" command to generate new projects

izaera:
This is the PR for #556. See commits for more information.

This is fully functional for a Liferay Platform project based on [portal-7.4-ga1 platform](https://github.com/liferay/liferay-frontend-projects/tree/master/target-platforms/packages/portal-7.4-ga1), though it needs manual steps to be tested since we haven't released anything to npmjs.com yet. 

Once we all agree this is the way we want to go, I will migrate the rest of the projects from the old generator to the new CLI tool so that we can deprecate the old Yeoman generator and continue implementing new features in this CLI tool.

## How to test this locally

If you want to test this locally, you need a local npm registry (eg: [Verdaccio](https://verdaccio.org/)) or publishing the target platform as a pre-release in npmjs.com. Once that's done, you will need to use your local copy of `liferay-cli` to generate a project, and your local copy of the JS Toolkit v2 to build it.

See [this](https://github.com/liferay/liferay-frontend-projects/pull/562/files#diff-34eaddd24b1da53887651a803665379dff158ea8e900f4983fbfdfc5dc60f5e3R119) for more information and/or ask me for help if needed.

## How this looks when tested locally

In case you don't want to setup the local tests, I'm adding an explanation on how this works and pushing the results of the generator to the [fat-jar-sample](https://github.com/izaera/fat-jar-sample) repo, project [my-app](https://github.com/izaera/fat-jar-sample/tree/master/packages/my-app).

> :eyes: Note that you cannot use the `my-app` at `fat-jar-sample` to test the build locally because setting that up in a yarn workspace is even more complicated. So, use `my-app` only as a reference.

```sh
ivan@baco:~ $ liferay new my-app
👋 Welcome to Liferay Project Generator v1.0.0
🛈  Gathering info to create new project 'my-app'...
? What is the human readable description of your project? My App
? What will be your target platform? portal-7.4-ga1
? Do you want to add localization support? Yes
? Under which category should your widget be listed? category.sample
? Do you want to add configuration support?

  💡 Needs Liferay DXP/Portal CE 7.1 with JS Portlet Extender 1.1.0 or
     Liferay DXP/Portal CE 7.2+.

 Yes
🛈  Generating new project 'my-app'...
✅ Successfully generated new project 'my-app'

  You can now run the following commands to build your project:

      $ cd my-app ↩
      $ npm install ↩
      $ npm run build ↩

  This will create a my-app.jar file in your my-app/dist
  folder that you can deploy to your local Liferay server.


ivan@baco:~ $ cd my-app
ivan@baco:~/my-app $ yarn
yarn install v1.22.10
warning package.json: No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Saved lockfile.
Done in 3.47s.
ivan@baco:~/my-app $ yarn build
yarn run v1.22.10
warning package.json: No license field
$ liferay build
Building project for target platform: @liferay/portal-7.4-ga1
ℹ️ Running babel on 2 files...
ℹ️ Running liferay-npm-bundler...
Copying package '/'...
Copied 1 packages
Applying rules to package '/'...
Applied rules to 1 packages
Transforming package '/'...
Babelifying 2 files in package '/'...
Transformed package '/'
Transformed 1 packages
Bundling took 265ms
The build has emitted some warning messages.
✔️ Project successfully built
Done in 1.71s.
ivan@baco:~/my-app $ cp dist/my-app-1.0.0.jar /opt/liferay/osgi/modules
ivan@baco:~/my-app $
```

And this is what you see in the Portal once you add the `my-app` portlet to a page:

image

You can also access the two portlet configuration types: :one: in System Settings, :two: in portlet configuration dialog

I'm attaching the generated JAR file for inspection, too: [my-app-1.0.0.zip](https://github.com/liferay/liferay-frontend-projects/files/6591214/my-app-1.0.0.zip)


bryceosterhaus:
I was only able to give this a cursory glance, I will check it out in more detail later but the overall direction seems good to me.

izaera:
I've pushed more commits for backup purposes, but they are not ready for review yet (they won't be changed much, but I have plans to reorder them and explain them more in depth).

izaera:
So, I've reorganized the commits and pushed everything. This is ready for review. 

I recommend taking a look at the description and read all the commits to get an idea of what is done, before jumping to the code.

izaera:
This PR is huge because it's the first MVP so I couldn't send it in chunks, but the rest should be smaller and iterative :crossed_fingers:

izaera:
The CI is broken because `yarn test` requires `yarn build` in some projects.

However, `yarn build` for the whole repo doesn't work, because it does not honor the dependencies between projects, so it fails to run builds in the proper order.

I'm looking for a solution...

projects/js-toolkit/packages/js-toolkit-core/src/format.ts
```diff
@@ -92,13 +93,13 @@ export function error(
  * ```
  *
  * @remarks
- * Error messages are prepended with a 'ℹ️' emoji.
+ * Error messages are prepended with a '🛈 ' emoji.
```
diegonvs, line 96:
```suggestion
 * Info messages are prepended with a '🛈 ' emoji.
```

and 

Screen Shot 2021-06-03 at 13 08 56

Looks like this emoji don't exists on MacOS Big Sur 11.3.1 (20E241)

projects/js-toolkit/packages/js-toolkit-core/src/format.ts
```diff
@@ -151,13 +152,31 @@ export function print(lines: string | string[], ...rest: string[]): void {
  * ```
  *
  * @remarks
- * Error messages are prepended with a '✔️' emoji.
+ * Error messages are prepended with a '✅' emoji.
```
diegonvs, line 155:
```suggestion
 * Success messages are prepended with a '✅' emoji.
```

projects/js-toolkit/packages/js-toolkit-core/src/format.ts
```diff
@@ -172,13 +191,13 @@ export function success(
  * ```
  *
  * @remarks
- * Error messages are prepended with a '❓' emoji.
+ * Error messages are prepended with a '❔' emoji.
```
diegonvs, line 194:
```suggestion
 * Question messages are prepended with a '❔' emoji.
```

projects/js-toolkit/packages/js-toolkit-core/src/format.ts
izaera: Thanks @diegonvs, that's one of the things I wanted to check, because I have lots of fonts with emojis installed in my system.

I'm not sure if we should remove them, because not everybody has fonts with emoji support and, when they are not available, the display looks quite bad.

I added the possibility to define an environment variable to prevent emojis from being shown, but I still find it a bit complex to configure for a very trivial functionality. On the other hand, the previous emojis we had have been a long time lying around and nobody has complaint, so it may be that they are standard enough...

@bryceosterhaus what do you think? Should we remove them or revert them to the previous ones (uglier but seemingly standard) and wait to see what happens when we release the tool?

izaera: Note that emoji signs make the messages easier to read, so they are not completely useless. But still, it's not a drama if we don't have them, I guess :shrug:

izaera:
Apparently `yarn ci` is not working in master with TS projects, either... I have no idea how it can be green in GH :shrug: 

In any case, I'm going to add a build step (for TS projects) prior to test so that this doesn't happen again. The problem is that `ts-jest` cannot find `js-toolkit-core` from `npm-bundler` if it is not built, which is logical given that the `package.json` of `js-toolkit-core` points to `lib/index.js` but that doesn't exist unless the project is built.

I'm sure there must be a way to avoid this and make `jest` find the source `.ts` files, but I can't find it and I don't think it adds too much value: if we know that building the project before the test fixes it, I think we can live with that.

izaera:
Mmm, I'm now seeing that the failures I'm getting locally are not the same as those in the CI... The CI is failing for bundler v2, the failures related to `ts-jest` are for bundler 3, and we already knew about them. So I think that build before test is already covered by the CI, that's why it is green.

Red herring. My fault :-\

projects/js-toolkit/packages/js-toolkit-core/src/format.ts
izaera: I have pushed-forced without the commit that changed emojis. We will be using the old ones that worked correctly at least in Mac and my Linux box.

bryceosterhaus: Personally, I like it when there are helpful and good looking emojis. However, if there is a case where we might see `[]`, I'd prefer to just keep our old ones, unless there is a way to easily provide a fallback.

izaera: It really depends on what fonts you have installed. There's a project, for example, to create a font of the fly with loads of icons :point_right: https://github.com/sebastiencs/icons-in-terminal#overview.

In general, I've found that the typical ones we use (which usually are 1 char wide and monochrome on Linux) are also available in Mac. However, it's -almost- impossible to know if a certain emoji will be displayed in a terminal, because you need to know the font in use and then check that font to see if the icon is there (unfeasible).

But I would say we are safe with the old ones...

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -384,36 +397,56 @@ export class Project {
...
-
-		if (autoPresets.length > 1) {
+		if (autopresets.length > 1) {
 			throw new Error(
 				'Multiple autopreset dependencies found in project (' +
-					autoPresets +
+					autopresets +
 					'): please remove the invalid ones or ' +
 					'explicitly define the preset to be used in the ' +
 					'.npmbundlerrc file'
```
kresimir-coko, line 407: What's the reason for us not using template literals here? We're using it below on line 437

izaera:
This is almost finished and I assume that we are more or less happy with it, so I'm going to split it into more different PRs to be able to:

1) Merge it easily
2) Release snapshot versions of each artifact to npmjs.com to be able to test it live

The main goal of this is to be able to do a product review without any need to configure anything to test local artifacts. This way, hopefully, @nhpatt and @bryceosterhaus may see how the whole workflow is expected to work and find things to be changed or issues.

I'll set this to draft so we don't accidentally merge it and proceed with the split and merge prior to snapshot releases.

maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/dependencies.js
```diff
@@ -57,12 +57,15 @@ export function addPackageDependencies(
...
-		.filter((dependencyDir) => {
-			return dependencyDir != null;
-		});
+		.filter(
+			(dependency) =>
+				!project.presetInfo.isAutopreset ||
+				dependency !== project.presetInfo.name
+		)
+		.map((dependency) =>
+			resolveDependencyDir(basedirPath, packageJson, dependency)
```
kresimir-coko, line 66: Shouldn't the name be `baseDirPath` instead of `basedirPath`. I know it's preexisting code, but just to bring attention to it.

projects/js-toolkit/packages/js-toolkit-core/src/format.ts
```diff
@@ -160,6 +161,24 @@ export function success(
...
+ * text`
+ *   This is a text message with some ${argument} to show
+ * `
+ * ```
+ */
+export function text(
+	literals: TemplateStringsArray,
+	...values: unknown[]
+): string {
+	return fmt.text(format('', literals, values));
```
kresimir-coko, line 179: `fmt` => `format`? Or is it a keyword?

projects/js-toolkit/packages/js-toolkit-core/src/index.ts
```diff
@@ -87,10 +87,18 @@ export {default as wrapModule} from './transform/js/operation/wrapModule';
 // JSON source transformation
 
 export * from './transform/json';
+export {default as addConfigurationField} from './transform/json/operation/addConfigurationField';
 export {default as addPkgJsonDependencies} from './transform/json/operation/addPkgJsonDependencies';
+export {default as addPkgJsonScripts} from './transform/json/operation/addPkgJsonScripts';
```
kresimir-coko, line 92: With #564 being worked on, shouldn't we avoid abbreviations `pkg` => `package`?

projects/js-toolkit/packages/js-toolkit-core/src/template/Renderer.ts
```diff
@@ -0,0 +1,66 @@
...
+		this._templatesDir = templatesDir;
+		this._outputDir = outputDir;
+	}
+
+	/**
+	 * @param templateRelPosixPath the template path (relative to templates dir) in POSIX format
+	 * @param data the contextual data to render the template
+	 * @return the contents of the rendered template
+	 */
+	render(templateRelPosixPath: string, data: object = {}): Promise<string> {
```
kresimir-coko, line 25: `templateRelPosixPath` => `templateRelativePosixPath`?

kresimir-coko (commented):
Looked over some files, there's so much 😓 

Mainly I noticed some potential SF stuff which I'm not sure I'm correct about.

projects/js-toolkit/packages/js-toolkit-core/src/index.ts
izaera: This is interesting, because bundler 2 and 3 where full of abbreviations like:

- `pkg` instead of `package` (which is a reserved word)
- `pkgJson` instead of `packageJson` because we use `pkg` instead of `package` and because it's easier to type and understandable (IMO)
- The implicit naming convention for relative/absolute paths ([this](https://github.com/liferay/liferay-frontend-projects/pull/562#discussion_r646538403)) which could lead to very long variables like `indexJsTemplateProjectRelativePosixPath`
- I'm sure there are more... :sweat_smile: 

So, we may change the codebase, but it will require a full review of the code and probably breaking the current APIs (for bundler 3 this is not a big problem, but we shouldn't do it for bundler 2).

projects/js-toolkit/packages/js-toolkit-core/src/template/Renderer.ts
izaera: See comment above. I used a naming convention and the `FilePath` class for paths because of how error prone it is path handling in a tool like this :scream: .

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: Probably none :sweat_smile: . Let me review and change it.

maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/dependencies.js
izaera: Yeah, makes sense to use uppercase in word stops. This is probably inconsistent, like [basename](https://nodejs.org/api/path.html#path_path_basename_path_ext) or [dirname](https://nodejs.org/api/path.html#path_path_dirname_path)...

projects/js-toolkit/packages/liferay-cli/src/new/index.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export default async function newProject(
+	name: string,
+	batch?: boolean
+): Promise<void> {
+	/* eslint-disable-next-line @typescript-eslint/no-var-requires */
+	const {version} = require('../../package.json');
+
+	print(title`|👋 |Welcome to Liferay Project Generator v${version}`);
```
bryceosterhaus, line 37: I didn't realize this was even a possible syntax. Very cool
```
fn`some string`
```

projects/js-toolkit/packages/liferay-cli/src/new/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+	const target: Target = targetLiferay;
```
bryceosterhaus, line 39: Is this type not inferred?

projects/js-toolkit/packages/liferay-cli/src/new/target-liferay/index.ts
```diff
@@ -0,0 +1,178 @@
...
+
+import * as facetConfiguration from '../facet-configuration';
+import * as facetLocalization from '../facet-localization';
+import * as facetPortlet from '../facet-portlet';
+import * as facetProject from '../facet-project';
+import prompt from '../util/prompt';
+
+import type {Options} from '..';
+
+export const name = 'Liferay Platform Project';
```
bryceosterhaus, line 26: We typically put constants in all caps, `name` => `NAME`

projects/js-toolkit/packages/liferay-cli/src/new/target-liferay/index.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export async function processOptions(options: Options): Promise<Options> {
+	options = await facetProject.processOptions(options);
+
+	options = await prompt(options, [
+		{
+			choices: ['portal-7.4-ga1'],
```
bryceosterhaus, line 33: nit: Would it make it easier to maintain if we imported all the possible choices from a config file or constant at the top of the file?

projects/js-toolkit/packages/liferay-cli/src/new/target-liferay/templates/src/AppComponent.js.ejs
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+export default class extends React.Component {
```
bryceosterhaus, line 4: I'd probably opt for encouraging use of non-class components of react. 

```jsx
export default function AppComponent() {
    return //....
}

File: projects/js-toolkit/packages/liferay-cli/src/new/util/HumanError.ts [hunk 14]
```diff
@@ -0,0 +1,10 @@
+/**
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+export default class HumanError extends Error {
+	constructor(message) {
+		super(message);
+	}
+}
```
bryceosterhaus, line 10: Whats the purpose of this class instead of using just Error?

projects/js-toolkit/packages/liferay-cli/src/new/util/getPortletName.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param pkgJsonName
+ */
+export default function getPortletName(pkgJsonName: string): string {
+	let portletName = pkgJsonName;
+
+	portletName = portletName.replace(/-/g, '');
+	portletName = portletName.replace(/[^A-Za-z0-9]/g, '_');
+
+	return portletName;
+}
```
bryceosterhaus, line 18: Is there any benefit to separating these into different lines?

```suggestion
export default function getPortletName(pkgJsonName: string): string {
  return pkgJsonName.replace(/-/g, "").replace(/[^A-Za-z0-9]/g, "_");
}

```

projects/js-toolkit/packages/liferay-cli/src/new/util/prompt.ts
```diff
@@ -0,0 +1,47 @@
...
+			if (typeof answer === 'function') {
+				answer = answer(options);
+			}
+
+			print(question`${prompt.message} {${answer}}`);
+
+			options[prompt.name] = answer;
+
+			return options;
+		}, options);
```
bryceosterhaus, line 40: I don't think you actually need a `reduce` here since you are just modifying the original object passed in.

```suggestion
		return (prompts as ReduceablePrompt[]).forEach((prompt) => {
			let answer = prompt.default;

			if (typeof answer === 'function') {
				answer = answer(options);
			}

			print(question`${prompt.message} {${answer}}`);

			options[prompt.name] = answer;
		});
```

bryceosterhaus (commented):
Looking good so far! I didn't go through the flow of the command yet and will review that high-level portion when you break it up into PRs like you mentioned.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -384,36 +397,56 @@ export class Project {
...
 			);
 		}
-		else if (autoPresets.length) {
-			presetFilePath = this.resolve(autoPresets[0]);
+
+		return autopresets.length ? autopresets[0] : undefined;
```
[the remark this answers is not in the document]
izaera: I'm leaning more towards `undefined` for these things because, well, the preset it's truly `not defined` :sweat_smile:, but I agree that JavaScript, being one of the only languages with that distinction, makes it a bit harder to know when to use which.

So, my rule of thumb is to use `null` when something is explicitly disabled/missing, and `undefined` when it is simply missing (we know it's not there, but nobody told us explicitly).

So, `null` is a strong hint that a configuration value is disabled, while `undefined` is missing and it can be replaced by a default value or a fallback, for example.

Look, [this stackoverflow answer](https://stackoverflow.com/questions/12606306/es5-when-to-use-null-and-when-to-use-undefined#12609664) makes it even more difficult :joy: . It says:

> The common sense is the following: undefined are variables which never have been assigned and non-existing properties.
> null however is a state of a variable or property which indicates that it has no value assigned.

So, what I'm saying, more or less. But then, there's a plot twist for function returns:

> Some methods like getElement... explicitely return null to indicate that the resultset is empty. If your function has no return statement, implicitely undefined is returned instead.
> In general, always assign null and never undefined.

Which I admit makes sense, since "void" functions return `undefined`.

So, I would say that my reasoning makes sense for variable assignments and properties, but not for function returns, as you say. 

Does that make sense? :thinking:

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -384,36 +397,56 @@ export class Project {
...
+			? readJsonSync(npmbundlerrcPath)
+			: {};
+
+		// Apply preset if necessary
+
+		let presetFilePath;
+
+		const autopreset = this._getAutopreset();
+
+		if (config.preset === undefined && autopreset) {
```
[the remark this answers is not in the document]
izaera: Well, the check `typeof x === 'undefined'` was used because in early ES versions you could assign a value to `undefined` and change it during the execution of the program as many times as you wanted, making checks like `x === undefined` fail. But that was fixed long ago, so there's no need to do it any more, AFAIK.

And because `x === undefined` is like `x === null` or `x === 1`, we use that for coherence.

Also, I think the linters, TypeScript, etc. would complain if you assigned anything to `undefined`, so I think it's safe to use this construct.

projects/js-toolkit/packages/js-toolkit-core/src/transform/json/operation/addPortletProperties.ts
```diff
@@ -0,0 +1,30 @@
+/**
```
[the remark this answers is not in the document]
izaera: That's one doubt I had too, because for text I use one single operation (`addLines`) for everything and we don't have transformers for CSS and .properties, for example... I didn't make them because they are complex and we don't really need them yet, because we only add (and don't modify or delete).

Regarding your quesiton, I think I prefer having multiple operations, one for each logical thing we do, but I don't know yet if that's the best and I'm not strongly convinced about it.

Rationale: the `transform` API is a powerful way to modify project files from any devtool we make. Basically we have transformers for JSON, text, and JavaScript and the idea that made me come up with this pattern was that 90% of the times what our tools do is simply modifying/mapping files (when transpiling, modifying configuration, writing new code, ...).

So, the idea is to have a set of most common operations in `js-toolkit-core` (that will be our API library) and then some more exotic operations inside the tools (f.e: `npm-bundler` has some transformation operations that are specific to its use case and are not meant to be shared with any other tool).

But now, some doubts arise: 

1) Do we really want to have a big set of operations in our API? Given that we don't use any namespacing at all (everything is exported from `index.js`) we may arrive to a point in the future where 90% of our API are transform operations without any categorization, some of them even deprecated/not used any more.
2) Should we use any namespacing/categorization to hold these transformations (i.e: put the text ones inside a `text` property of the API, the JSON ones inside `json`, and so on, for example)?
3) Should we instead move all transformations to a different place and don't make them part of our API? (probably a different npm package, but then there's the release and dependency handling problem if we take this approach)
4) Should we aim at very generic transformations or give them a proper name that specifies exactly what they do (your original question) ?

I don't think we have enough data yet to know what is best :thinking: 

My intuition tells me that having transformations with proper names is better than having too generic ones. This is because:

1) In the end, you could have a single transformation for everything, which would totally defeat the purpose of the API (reductio ad absurdum)
2) If we have a catalog of transformations, the API is much easier to use/browse, and we can know what things we touch in the configuration by simply looking at the API.

For example, if you have `addConfigurationField.ts`, `addPkgJsonScripts.ts` and `addPortletProperties.ts` you know without looking at more code that our tools deal with configurations fields, which are inside JSON files, modify npm scripts in some places, and modify portlet properties in others. 

Also, because you know what the user is trying to do when he invokes `addConfigurationField` instead of `addPkgJsonScripts`, you can impose more restrictions on (and document) the structure with TypeScript or make more validations.

Finally, when you use this API, having this type of transformations looks better, because you would read:

```javascript
transformJson(
    packageJson,
    addConfigurationField(x),
    addPkgJsonScript(y),
    addPortletProperties(z)
)
```

Which documents itself.

However, this doesn't mean that I like having an API with myriad unrelated transformations, so not sure what to do about it, yet... I would say we should probably namespace operations to avoid polluting the root namespace of `js-toolkit-core` and see case by case if they make sense as a general purpose transformation or we prefer to generalize it or hide it inside the tools...

projects/js-toolkit/packages/liferay-cli/src/build.ts
```diff
@@ -0,0 +1,73 @@
...
+			if (
+				packageJson.dependencies &&
+				packageJson.dependencies['liferay-npm-bundler']
+			) {
+				return true;
+			}
+
+			return false;
+		}
+	);
```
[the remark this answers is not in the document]
izaera: Yeah, that's [what we agreed in the issue](https://github.com/liferay/liferay-frontend-projects/issues/542#issuecomment-844109982), but we can still change it.

I think it makes sense, though, because having a target platform without the bundler wouldn't work because you wouldn't be able to build the project. And there's no other scenario in which pulling the bundler with a preset makes sense, because prior to target platforms, it was the bundler who used the preset, so we are doing an inversion of control now, thus there won't be instances of this inversion of control in the past.

In any case, we can expand this logic in the future if we need to. For example, we can use keywords to tag target platforms that don't include the bundler, use a npm scope, or anything else we may define, ...

So, unless we can think of a case where this detection logic would pose a problem, I think we can go with it for now, as we agreed.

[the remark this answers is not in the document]
izaera: I agree that looking for `liferay-npm-bundler` it's not the cleanest check we can make, but on the other hand, `liferay-npm-bundler` is not going to change (only for bundler v3, but that's incompatible with this presets) and it's a npm package name, so something (almost) set in stone, as a URL or a UUID.

Think of it as a tagging mechanism. And if we need to make target platforms that don't include `liferay-npm-bundler` as a dependency in the future, we can expand this logic, as I said...

projects/js-toolkit/packages/liferay-cli/src/new/index.ts
izaera: ES6 :-)

projects/js-toolkit/packages/liferay-cli/src/new/util/HumanError.ts
izaera: `HumanError` is used to distinguish build failure errors (things that failed because the user did something wrong which he needs to fix) from unhandled errors (our fault when programming the tool).

So, in the global `catch` I can check for `HumanError` vs `Error` and, when it is a `HumanError` show a pretty error message to the user hinting him what to do, whereas when it is a plain `Error` I show the stack trace of the code and blame it on use, so that the user can file an issue in our repo :roll_eyes: .

If it not clear, we may give it a more meaningful name, perhaps... `BuildError` or something like that?

izaera: I probably meant `HumanReadableError` when I came up with that name. But I admit the naming was a bit unfortunate :-(

projects/js-toolkit/packages/liferay-cli/src/new/util/getPortletName.ts
izaera: Simply readability. But I'm OK with chaining too :shrug:

projects/js-toolkit/packages/liferay-cli/src/new/util/prompt.ts
izaera: Not really, because you need to return the options object and [forEach doesn't return anything](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#return_value).

However, we can turn the `prompt` function into an in-place modification function. So, instead of:

```javascript
function prompt(options: Options) : Options
```

Make it:

```javascript 
function prompt(options: Options): void
```

Though I really prefer the first option, because it looks more natural to receive the current options and return the new ones.

And if we want to be more FP, I suppose we can introduce this change in the code:

```javascript
return (prompts /*...*/).reduce((options, prompt) => { /*...*/ }, {...options});
```

To make it explicit that we are not transforming the input options in any way.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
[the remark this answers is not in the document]
izaera: Addressed in https://github.com/liferay/liferay-frontend-projects/pull/576

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: Addressed in https://github.com/liferay/liferay-frontend-projects/pull/576

maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/dependencies.js
izaera: Addressed in https://github.com/liferay/liferay-frontend-projects/pull/576

projects/js-toolkit/packages/js-toolkit-core/src/transform/json/operation/addPortletProperties.ts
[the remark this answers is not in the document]
izaera: Addressed in https://github.com/liferay/liferay-frontend-projects/pull/578

projects/js-toolkit/packages/liferay-cli/src/build.ts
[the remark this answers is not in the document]
izaera: Related to https://github.com/liferay/liferay-frontend-projects/pull/576

projects/js-toolkit/packages/js-toolkit-core/src/index.ts
izaera: I'm leaving this as unchanged for now. Once #564 is in place we may decide what to do with bundler v3 (before releasing a following version).

maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/dependencies.js
```diff
@@ -57,12 +57,15 @@ export function addPackageDependencies(
...
-		.map((dependency) => {
-			return resolveDependencyDir(basedirPath, packageJson, dependency);
-		})
-		.filter((dependencyDir) => {
-			return dependencyDir != null;
-		});
+		.filter(
+			(dependency) =>
+				!project.presetInfo.isAutopreset ||
+				dependency !== project.presetInfo.name
```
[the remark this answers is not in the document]
izaera: Not really, because the filtering is operating in different types of objects (because of the `map()` in the middle).

In any case, filtering several times shouldn't matter too much as long it's one after another, because the conceptual complexity is the same, i.e:

```javascript
let a = '';

for (let i=0; i<100; i++) {
  a += 'x';
}

for (let i=0; i<100; i++) {
  a += 'x';
}
```

is equivalent to:

```javascript
let a = '';

for (let i=0; i<100; i++) {
  a += 'x';
  a += 'x';
}
```

Both in result, execution times, and resource consumption.

There may be a small benefit in doing just one loop because you don't need to allocate the `i` variable twice and because you don't need halve the number of increments to the `i` variable, but those should be insignificant compared to the body of the loop (under normal conditions). Also, the compiler/interpreter should be smart enough to optimize it :crossed_fingers: .

In fact, we have lots of [similar constructs in our codebase](https://github.com/liferay/liferay-frontend-projects/search?q=filter%28Boolean%29).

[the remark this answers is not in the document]
izaera: all over the place

projects/js-toolkit/packages/liferay-cli/src/new/target-liferay/index.ts
izaera: Yes. I put that in-place because we only have one, but as soon as the second appeared, I would have moved it to a constant too :sweat_smile:

projects/js-toolkit/packages/liferay-cli/src/new/index.ts
izaera: I guess it should, but I added it because my "IDE" (vim+CoC) was not recognizing it. However, looking at the code now, it seems like it should have recognized it :shrug: . I will check again when I prepare the chunk PR for this.

izaera:
I'm closing this since I've already addressed all feedback in a new PR I'm preparing.

Issue #437, opened 2021-03-10
Add lint to catch common abbreviations


wincent:
This kind of thing comes up in code review occasionally ([example](https://github.com/liferay-frontend/liferay-portal/pull/865#discussion_r590185100)) and we could actually catch some of these things via linting, at least for the common cases.

See the table "Examples of bad abbreviations" in [our guidelines](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/naming.md#avoid-abbreviations). At least some of those could safely be flagged by a linter without much risk of false-positives.

PR #564, opened 2021-06-01
feat(eslint-config): Add lint to catch common abbreviations

ambrinchaudhary:
I added a lint rule to avoid common abbreviations in identifiers, as described in our [guidelines](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/naming.md#avoid-abbreviations). 
Also, should I add a test for each abbreviation? 

Fixes https://github.com/liferay/liferay-frontend-projects/issues/437


izaera:
I think it's OK :+1: 

I would run it over this whole repo (and probably liferay-portal too), before merging, just to know if it is going to generate a lot of error messages :sweat_smile: 

> Also, should I add a test for each abbreviation?

I don't think that adds anything. You are already covering all the code with the two tests you've written.

bryceosterhaus:
Looks good so far. I like Ivan's idea of running it over this repo and portal just to get an idea of what we are up against.

ambrinchaudhary:
Just for update... I run this lint rule over this repo only, and this was the result  😱  : 
`2125 problems (1795 errors, 330 warnings)`

Although, there are errors caught by this rule that shouldn't, like: 
```
 31:2   error  Avoid constructor abbreviation, preferred alternative is function Object() { [native code] }  @liferay/liferay/no-abbreviations
220:25  error  Avoid toString abbreviation, preferred alternative is function toString() { [native code] }  @liferay/liferay/no-abbreviations
```

I am now looking for a fix for these cases... but... still... we have a lot of lint errors in the project... not sure if I want to test it in portal ;-)

bryceosterhaus:
My guess is that there are a lot of false positives due to methods being on the `ABBREVIATIONS_REPLACEMENTS` object. This is why `toString` shows up because it is an Identifier and it exists on `ABBREVIATIONS_REPLACEMENTS`. It will still pass as true when you check for the key even if its not a value.

You might try using a Map, so it would look like

```js
const ABBREVIATIONS_REPLACEMENTS = new Map([
  ["arr", ["array"]],
  ["btn", ["button"]],
  ["cb", ["callback"]],
  ["desc", ["description"]],
  ["e", ["event", "error"]],
  ["el", ["element"]],
  ["err", ["error"]],
  ["evt", ["event"]],
  ["fm", ["form"]],
  ["fmt", ["format"]],
  ["idx", ["index"]],
  ["img", ["image"]],
  ["lg", ["large"]],
  ["md", ["medium"]],
  ["obj", ["object"]],
  ["opts", ["options"]],
  ["prj", ["project"]],
  ["sm", ["small"]]
]);

ABBREVIATIONS_REPLACEMENTS.has('err') // true
ABBREVIATIONS_REPLACEMENTS.get('err') // ["error"]
ABBREVIATIONS_REPLACEMENTS.has('toString') // false

```

projects/eslint-config/plugins/liferay/lib/rules/no-abbreviations.js
```diff
@@ -0,0 +1,67 @@
...
+
+	meta: {
+		docs: {
+			category: 'Best Practices',
+			description: 'Avoid abbreviations',
+			recommended: false,
+			url:
+				'https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/naming.md#avoid-abbreviations',
+		},
+		fixable: 'code',
```
bryceosterhaus, line 63: Since we aren't providing the automatic fix for them, we can set this as `null`

projects/eslint-config/plugins/liferay/tests/lib/rules/no-abbreviations.js
```diff
@@ -0,0 +1,48 @@
...
+				{
+					message:
+						'Avoid btn abbreviation, preferred alternative is button',
+					type: 'Identifier',
+				},
+			],
+		},
+	],
+
+	valid: [],
```
bryceosterhaus, line 47: Let's add a sanity check for regressions here, even just something like `(error) => error.toString()`

projects/eslint-config/plugins/liferay/lib/rules/no-abbreviations.js
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const formatMessage = function (abbreviation, replacements) {
+	if (replacements.length > 1) {
+		return `Avoid ${abbreviation} abbreviation, preferred alternatives are [${replacements.join(
+			','
+		)}]`;
+	}
+	else {
+		return `Avoid ${abbreviation} abbreviation, preferred alternative is ${replacements}`;
```
bryceosterhaus, line 34: Since the abbreviation and replacement is dynamic based on the error, let's wrap it in quotes so that it stands out to the reader.

```suggestion
		return `Avoid "${abbreviation}" abbreviation, preferred alternative is "${replacements}"`;
```

projects/eslint-config/plugins/liferay/lib/rules/no-abbreviations.js
```diff
@@ -0,0 +1,67 @@
...
+	opts: ['options'],
+	prj: ['project'],
+	sm: ['small'],
+};
+
+const formatMessage = function (abbreviation, replacements) {
+	if (replacements.length > 1) {
+		return `Avoid ${abbreviation} abbreviation, preferred alternatives are [${replacements.join(
+			','
+		)}]`;
```
bryceosterhaus, line 31:
```suggestion
return `Avoid "${abbreviation}" abbreviation, preferred alternatives are [${replacements
  .map((replacement) => `"${replacement}"`)
  .join(",")}]`;

```

projects/eslint-config/plugins/liferay/lib/rules/no-abbreviations.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+const ABBREVIATIONS_REPLACEMENTS = {
```
diegonvs, line 6: I found [this repo](https://github.com/kisvegabor/abbreviations-in-code) here containing some common abbreviations that might look interesting. I found some missing abbreviations that I caught myself using like:
```
...
cfg: ['configuration']
cmd: ['command']
col: ['column']
coord: ['coordinate']
ctx: ['context']
func: ['function']
len: ['length']
msg: ['message']
param: ['parameter']
pos: ['position']
res: ['result']
src: ['source']
...
```

Does make sense including on the Map?

izaera: I would say we should keep adding abbreviations as we find them in reviews. Ideally it should settle some day into a list containing team's biases :sweat_smile:

ambrinchaudhary: Makes sense to me... although for now I just added the ones written in our [guidelines](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/naming.md#avoid-abbreviations). 
Some of them (`cfg, col, msg, res`...) could be very common to find in portal (I guess)... so what is your opinion, should we add them now or wait to find them in portal? Personally, I would add them now and update our guidelines...

izaera: > Personally, I would add them now and update our guidelines...

I agree, though I don't know how much work that is. So maybe we don't want to add (now) those that appear in myriad places. Especially since fixing format issues here is easy, but that's not the case for `liferay-portal` :-(

projects/eslint-config/plugins/liferay/lib/rules/no-abbreviations.js
ambrinchaudhary: Well, I think it is good if we provide a fix for the simplest cases (which are the ones that have only one possible replacement - most of them-), and I did it in this [commit](https://github.com/liferay/liferay-frontend-projects/pull/564/commits/bf9a821bf1bf3aaa65802031489d731bf96cec68). Maybe before merging we should fix all the abbreviations in the current project... or... would be better if we avoid this?

ambrinchaudhary:
Thanks @bryceosterhaus, yes, using a Map removes the false positives that I was getting. So now.... as you can see in the current lint check, we have:

`✖ 1799 problems (1469 errors, 330 warnings)`

What do you think, should we try to fix them? I created a fixer and I guess most of the cases were covered (just the `['e', ['event', 'error']]` will need a manual fix if present)

projects/eslint-config/plugins/liferay/lib/rules/no-abbreviations.js
izaera: I'm a bit worried about that fix :thinking: I usually assume that SF and `lint --fix` are safe operations that can be blindly trusted, but this rename of identifiers may break things IINM. For example, this code:

```javascript
const x = new YourPreferredAPI();

x.cb = 1;
```

would be transformed to:

```javascript
const x = new YourPreferredAPI();

x.callback = 1;
```

izaera: Seeing that :point_up: I'm realizing that the fix should be OK as long assuming the detection of the problem is safe. So maybe, it's not the fix what needs to be changed, but the logic checking the problem.

In other words, it doesn't seem enough to check for `Identifier` only, but we need to make sure that the `Identifier` is referring to a local variable or parameter, for example :thinking: 

Babel has tools to check the scope of Identifiers, I assume that eslint can do the same...

ambrinchaudhary: Good catch @izaera, you are rigth, that example you wrote would be transformed by the current code. I agree with you that is in the detection phase where that `.cb` should not be detected as a lint error. I will work on this 👍

izaera: Yeah, I suffered it quite a lot of times when coding the bundler :sweat_smile:

ambrinchaudhary: So... I spent all the morning trying to find a solution for this case but I couldn't... I don't see any way to distinguish the `Identifier` type that fits in this particular case... 

Anyway, I was thinking about it and maybe is it not an error... in the example, `x.cb = 1`; will be catched and fixed but also the `cb` declaration. The only thing that could cause errors is when `YourPreferredAPI` is an external library... but it is not common that other libraries exposes methods or attributes with these abbreviations that we are trying to catch... but I know, I'm assuming things. 

In addition, I found found [project](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/prevent-abbreviations.js) doing something similar, and this kind of code [`this.err = 1`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/prevent-abbreviations.mjs#L318) is also catched as an error... 

Or another solution could be not to provide a fix (but use it before to fix oour current linting errors), forcing people to manually fix them....

bryceosterhaus: IMO, I think for fixable rules we should distinguish between rewriting vs reorganizing someones code. I would prefer a user updates their own code if it is rewriting. Reorganizing, such as sort/order, we can be more confident in making those changes with autofix. I think the potential bugs with autofixing abbreviations outweigh the gain of having them autofix. They aren't hard changes to make manually and users will learn the pattern overtime.

So I would probably not worry about making this rule automatically fixable

bryceosterhaus: > They aren't hard changes to make manually and users will learn the pattern overtime.

This applies to new code after the rule is added. Fixing 1k+ warnings isn't super easy lol

bryceosterhaus:
@ambrinchaudhary we could probably silence the warning for our maintenance folder for now. It's probably not a great idea to rewrite a bunch of code in those projects. We might be able to do this by adding a eslintrc in the maintenance folder and disable this rule.

projects/eslint-config/plugins/liferay/lib/rules/no-abbreviations.js
bryceosterhaus: Also, I will just add, if the fix seems to work well and we don't have any edge cases, I am fine to add it. My concern was mainly having to maintain something complex or something that has a lot of edge cases where we could potentially introduce regressions.

But if it seems to work well, then let's go for it.

izaera: [This](https://github.com/liferay/liferay-frontend-projects/blob/1cf26ad3a5a2750a5d5d64d9671e126ad02de177/maintenance/projects/js-toolkit/packages/babel-plugin-namespace-amd-define/src/index.js#L42) is how you check scopes in Babel.

However, I don't know if `eslint` provides that `scope` field or not, because it is not part of standard ASTs, so it may be a Babel-only thing...

Regarding the rest, I agree that I would prefer not having an automatic fix for this (the truth is there aren't many fixes available in general, so why taking the risk...). For the `maintenance` folder, we already need to disable a lot of warning for Senna, so we can do the same with this, I guess.

ambrinchaudhary:
> @ambrinchaudhary we could probably silence the warning for our maintenance folder for now. It's probably not a great idea to rewrite a bunch of code in those projects. We might be able to do this by adding a eslintrc in the maintenance folder and disable this rule.

Hey @bryceosterhaus and @izaera, I'm not sure why are you only refering to the `maintenance` folder to disable the errors... maybe it is just a mistake because this new lint rule fires errors in the whole project. So there are two different options here: 

1. Disable the rule in the whole project. But then... should we do this in `portal` as well? And...how we avoid abbreviations in the future code? I don't think is worth to create a rule and disable it, so if this is the case, maybe it is better to not merge this pull and discard the current issue 🤷‍♀️ 
2. Use the fixer to automatically fix the abbreviations in this project (I ignored the [maintenance projects](https://github.com/liferay/liferay-frontend-projects/pull/564/commits/96bd31cb8bb9b61e52222deb495dcde49e694dc7) as you suggested) and avoid lint errors, and keep the rule for the future code. But, do not provide the fixer, users will have to do it manually. I did this in the last two commits (https://github.com/liferay/liferay-frontend-projects/pull/564/commits/53058579a1955cb6f39ea60e3c85ba7a78d9621f and https://github.com/liferay/liferay-frontend-projects/pull/564/commits/585a134c6cbe213aabe684e2a4197f526e180686) and I personally think this is the best option. I just need to review all the [automatic](https://github.com/liferay/liferay-frontend-projects/pull/564/commits/53058579a1955cb6f39ea60e3c85ba7a78d9621f) fixes to check if there are no regressions (I'm on it)

Of course, if you are not agree with 2) I can remove the commits, but then I'm not sure what we should do with this pull ;-)

projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/r2/index.js
```diff
@@ -19,9 +19,9 @@ const PluginError = gutil.PluginError;
 const PLUGIN_NAME = 'gulp-R2-css';
 
 function gulpR2() {
-	const task = through.obj(function (file, enc, cb) {
+	const task = through.object(function (file, enc, callback) {
```
bryceosterhaus, line 22: This is `obj` to `object` is likely going to cause issues. `through.obj` is a method and so we can't change this.

Are we able to add a test case for this so that we don't emit errors for abbreviations like this?

projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/build/index.js
```diff
@@ -34,7 +34,7 @@ function injectJS() {
 		'/liferay-frontend-theme-unstyled/templates/portal_normal\\.(ftl|vm)$'
 	);
 
-	return through.obj(function (file, encoding, callback) {
+	return through.object(function (file, encoding, callback) {
```
bryceosterhaus, line 37: Same issue here. We should make sure the identifier we are checking is not a key on an object

projects/js-themes-toolkit/packages/liferay-theme-tasks/theme/tasks/build/sass.js
```diff
@@ -18,23 +18,25 @@ function replaceExtension(filePath, newExt) {
 }
 
 module.exports = (options) =>
-	through.obj((file, enc, cb) => {
+	through.object((file, enc, callback) => {
```
bryceosterhaus, line 21: ^same as above

projects/js-toolkit/packages/js-toolkit-scripts/src/scripts/create-jar/xml.ts
```diff
@@ -22,27 +22,27 @@ const TYPES = {
 export function addMetatypeAttr(
 	metatype: XmlObject,
 	id: string,
-	desc: ConfigurationJsonField
+	description
```
bryceosterhaus, line 25: Looks like we accidentally removed the type declaration for this re-write. This shouldn't be a future issue for other people if we don't add autofix, but we need to fix this here.

projects/npm-tools/packages/changelog-generator/src/yarn.js
```diff
@@ -24,9 +24,9 @@ function yarn(command) {
 
 		if (status !== 0) {
 			// eslint-disable-next-line no-console
-			error(
+			log(
```
bryceosterhaus, line 27: Was this intentional to go from error to log?

projects/npm-tools/packages/npm-scripts/src/utils/deepMerge.js
```diff
@@ -14,17 +14,18 @@ const clone = (value, options) => merge(emptyTarget(value), value, options);
 function combineMerge(target, source, options) {
 	const destination = target.slice();
 
-	source.forEach((e, i) => {
+	source.forEach((element, i) => {
```
bryceosterhaus, line 17: `element` is a little odd to use here since we aren't talking about an html element. Could we rename this to `item`, which I think is generic enough for a "merge" function?

target-platforms/packages/create-plaform/assets/build.js
```diff
@@ -84,9 +84,9 @@ function runBabel() {
 				'utf8'
 			);
 		}
-		catch (err) {
+		catch (error) {
```
bryceosterhaus, line 87: The GH action error above looks suspicious. You may need to rename this as `babelError` or something if this file is already using `error` as a variable in the file.

target-platforms/packages/portal-7.4-ga1/build.js
```diff
@@ -84,9 +84,9 @@ function runBabel() {
 				'utf8'
 			);
 		}
-		catch (err) {
+		catch (error) {
```
bryceosterhaus, line 87: See GH action error above.

bryceosterhaus (changes requested):
Hey @ambrinchaudhary, this is looking good and I am surprised how many issues it is already catching, so this looks promising to add. I answered your comments below and I also commented on a few of the changes that we should look into fixing. 

You may need to be more specific in the AST check, so instead of looking for just Identifier, you may need to look for only certain types of Identifiers, such as `err.someMethod` instead of `someObject.err`. Changing `err` to `error` in the second case would cause logic issues and the former shouldn't

As always, let me know if you have any questions!

-------
>  I'm not sure why are you only refering to the maintenance folder to disable the errors

The reason we want to disable for just the maintenance folder is because we don't intend for active development for those projects. By disabling the rule for this folder, we reduce the possibility of adding regressions to projects that aren't in active development. We still want this rule for our actively developed projects though.


> 2.  Use the fixer to automatically fix the abbreviations in this project (I ignored the maintenance projects as you suggested) and avoid lint errors, and keep the rule for the future code. But, do not provide the fixer, users will have to do it manually. I did this in the last two commits and I personally think this is the best option. I just need to review all the automatic fixes to check if there are no regressions (I'm on it)

This sounds good to me. I like the idea of using the fixer just for this PR so that we can easily get to a state of no lint errors, but then going forward we don't auto fix it so that users will have to manually address.

projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/r2/index.js
bryceosterhaus: You might be able to check if its a property by doing something like

```js
    Identifier(node) {
        const isProperty = node === node.parent.property;
```

projects/js-toolkit/packages/js-toolkit-scripts/src/scripts/create-jar/xml.ts
izaera: Interesting case: because the fix doesn't take into account TypeScript it fails to maintain the type declaration :scream: . This is one of the more difficult issues of dealing with ASTs: that you never know in advance how many constructs there may be, unless you read the full spec from line 1 to last (and even in that case you could forget one unwittingly :-( ).

projects/eslint-config/plugins/liferay/lib/rules/no-abbreviations.js
izaera: I think [this is how you check scopes in eslint](https://github.com/liferay/liferay-frontend-projects/pull/575/files#diff-0e30dffdb2f9f531b511c9f9998453ee58834d5077ea7dc78701839c16640672R14)...

projects/npm-tools/packages/changelog-generator/src/yarn.js
ambrinchaudhary: yeah, I changed this intentionally. There were to objects called "error": 
```
const {error} = require('./console');

const {error} = child_process.spawnSync('yarn', [command], {stdio: 'inherit'});
```

So I decided to change the console message to a `log` instead than an error, I think it is not a bug deal...

target-platforms/packages/create-plaform/assets/build.js
ambrinchaudhary: Actually, the `error` variable declared before was not being used (and `yarn lint` was complaining), so I'm removing that one.

ambrinchaudhary:
@bryceosterhaus @izaera, I fixed the lint errors that were replaced wrongly. Also, thank you very much for your tip [here](https://github.com/liferay/liferay-frontend-projects/pull/564#discussion_r646819516). Ivan pointed out the same issue and I was not able to find a solution for this case 🙈 .

I guess everything is fixed and completed, but if not and you prefer to take the pull and do the changes needed, feel free! I won't complain ;-)

projects/js-themes-toolkit/packages/generator-liferay-theme/lib/generation/theme-based.js
```diff
@@ -107,11 +107,11 @@ async function downloadThemeFiles(project, themeName, themeVersion) {
 
 		fs.unlinkSync(warFile);
 	}
-	catch (err) {
+	catch (error) {
 		print(error`
 			Error downloading and extracting Liferay's ${themeName} theme:
 
-			${err.message}
+			${error.message}
```
izaera, line 114: This one is wrong (there's an `error` already in scope)

projects/js-themes-toolkit/packages/generator-liferay-theme/lib/generation/theme-based.js
```diff
@@ -157,12 +157,12 @@ async function getThemeVersions(themeName) {
 
 		return themeVersions;
 	}
-	catch (err) {
+	catch (error) {
```
izaera, line 160: This one is wrong (there's an error already in scope)

projects/js-toolkit/packages/js-toolkit-scripts/src/index.ts
```diff
@@ -69,8 +69,8 @@ export default async function main(): Promise<void> {
 
 			await scriptFunction();
 		}
-		catch (err) {
-			print(error`Script 'js-toolkit ${script}' failed:\n${err.stack}`);
+		catch (error) {
```
izaera, line 72: This one is wrong. There's already an `error` in scope.

projects/js-toolkit/packages/js-toolkit-scripts/src/scripts/copy-assets.ts
```diff
@@ -28,9 +28,9 @@ export default function (): Promise<void> {
 				filter: (path) => !/\/\.placeholder$/.test(path),
 				overwrite: true,
 			},
-			(err) => {
-				if (err && err.message !== 'No files to copy') {
-					print(error`${err}`);
+			(error) => {
```
izaera, line 31: This one is wrong. There's already an `error` in scope.

projects/js-toolkit/packages/js-toolkit-scripts/src/scripts/copy-sources.ts
```diff
@@ -24,9 +24,9 @@ export default function (): Promise<void> {
 			project.sourceDir.asNative,
 			outputDir.asNative,
 			{confirm: true, overwrite: true},
-			(err) => {
-				if (err) {
-					print(error`${err}`);
+			(error) => {
```
izaera, line 27: This one is wrong. There's already an `error` in scope.

projects/js-toolkit/packages/js-toolkit-scripts/src/scripts/create-jar/xml.ts
```diff
@@ -22,27 +22,27 @@ const TYPES = {
 export function addMetatypeAttr(
 	metatype: XmlObject,
 	id: string,
-	desc: ConfigurationJsonField
+	description: ConfigurationJsonField
```
izaera, line 25: Since this is of type `ConfigurationJsonField` I would call it `field` of `fieldDescription` better...

It's ironic that I was the one naming it `desc` and now I request a more proper name :muscle: :joy: 

However it has an explanation: this was ported from JS to TS and, in TS, there was no `ConfigurationJsonField` type :-)

projects/js-toolkit/packages/js-toolkit-scripts/src/scripts/translate.ts
```diff
@@ -100,11 +100,11 @@ export default function (): void {
 
 			print(success`Finished`);
 		})
-		.catch((err) => {
+		.catch((error) => {
```
izaera, line 103: This one is wrong. There's already an `error` in scope.

projects/npm-tools/packages/changelog-generator/src/yarn.js
```diff
@@ -6,7 +6,7 @@
 const child_process = require('child_process');
 const fs = require('fs');
 
-const {error, info} = require('./console');
+const {info, log} = require('./console');
```
izaera, line 9: I don't like this one (I saw it yesterday, but out of context, and didn't provide any feedback because of that). Usually `log` is much less than `error` in trace systems. So, something like: `critical` > `error` > `warning` > `info` > `log`/`debug`.

In this case we rename `error` to `log` and now it may look like `info` > `log` (or at least it is not clear which is higher up in the hierarchy).

So I would:

- leave it as `error` and rename the exception variables
- rename it as `fail`, `abort`, or something like that...

target-platforms/packages/create-plaform/assets/build.js
```diff
@@ -84,9 +83,9 @@ function runBabel() {
...
 			/* eslint-disable-next-line no-console */
-			console.log(err.toString());
+			console.log(error.toString());
 			print(error`Build failed`);
```
izaera, line 89: This one is wrong. There's already an `error` in scope.

target-platforms/packages/create-plaform/assets/build.js
izaera: Not really. It is used in line 89. The renaming is incorrect.

target-platforms/packages/portal-7.4-ga1/build.js
```diff
@@ -84,9 +83,9 @@ function runBabel() {
...
```
izaera, line 86: This one is wrong. There's already an `error` in scope.

izaera (changes requested):
Overall, looks good. However I've left comments about wrong renames (most of them related to the `error` method in the `format` facility vs `error` variables).

Seeing how many of them there are, and what the expected usage of the format API was, which leverages ES6 templates to write constructs like:

```javascript
print(error`....`);
print(warn`....`);
print(info`....`);
```
We may need to redesign that API to rename the`error` template function to some other thing, like `fail`, for example.

Not being convinced myself about not using abbreviations for `callback` and `error` I decide to test my belief and did a search in GH to see occurrences (in code) of:

- `cb` vs `callback` and it was a 156M/339M ratio
- `err` vs `error` and it was 667M/1B ratio

So I guess people tends to avoid abbreviations more often than not. 

* Note that the search I made does not take into account if they are identifiers, comments, or output messages, thus there will always be a bias against abbreviations. However, the difference is so big, that it seems a sensible conclusion. In fact, node API uses `callback` and `error` in its documentation and source code.

target-platforms/packages/create-plaform/assets/build.js
ambrinchaudhary: ops, you are right

target-platforms/packages/portal-7.4-ga1/build.js
ambrinchaudhary: right! fixing...

projects/npm-tools/packages/changelog-generator/src/yarn.js
ambrinchaudhary: Got you! ok, I finally renamed the exception variable 👍

ambrinchaudhary:
Thanks a lot for your double check @izaera, I missed all those replacements 🤦‍♀️

bryceosterhaus:
I caught a few more issues where the variable was already in scope and renamed them. I push a commit to your branch. Once this passes CI I will merge. Next step is merging to portal and fixing the lint issues there 😨 

Thanks for your work on this, @ambrinchaudhary!

izaera:
> Thanks for your work on this, @ambrinchaudhary!

:clap: 

> Next step is merging to portal and fixing the lint issues there :fearful: 

looking for a job

PR #567, opened 2021-06-02
feat(eslint-config): Enable spaced-comment rule

### LINKED ISSUE #225: Ensure space between "//" and comment content
As [mentioned here](https://github.com/brianchandotcom/liferay-portal/pull/95673#issuecomment-722598862).

See [example fix](https://github.com/brianchandotcom/liferay-portal/commit/6677adf7cddf20860ab0da8f110ed921d1ac4b95).

Good:

```js
// Yo. Sup?
```

Bad:

```js
//Yo. Sup?
```

kresimir-coko:
Fixes #225 

After trying to figure out how to write an ESLint rule for this, I stumbled upon the rule from ESLint that IMO tackles our problem. The rule can be found at https://eslint.org/docs/rules/spaced-comment.

I've added the `exception: ['/']` just so that it doesn't scream when running `yarn ci` and encountering [these long comment lines](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/index.test.js#L13).

❓ I'm not sure how to proceed exactly if this is the way to do this.
- Publish a new version of the package?
- Test if it works in `liferay-portal`?
- Something else?


bryceosterhaus:
@kresimir-coko that rule looks great and seems to fit what we need.

As for the `exceptions`, I think instead of allowing that we should update those portions of code with outlier comment blocks. @izaera, I noticed you added a few comment blocks with `///////`, are you opposed to removing those and making it more uniform with our other comment blocks?

> Publish a new version of the package?

Let's wait a bit before publishing a new version. We have a few eslint tickets being worked on and so we can wait till they are all added. I can ping you when they all get merged and have you try cutting a release

> Test if it works in liferay-portal?

Go ahead and test this in portal by updating the eslintrc that we have in dxp, `modules/.eslintrc.js`.

izaera:
> are you opposed to removing those and making it more uniform with our other comment blocks?

Of course not. We can change them to whatever makes it simpler :+1:

ambrinchaudhary (approved):

bryceosterhaus:
@kresimir-coko go ahead and remove the exceptions part and then in a separate commit, update our existing comments that use non-standard syntax.

For example:
```js
////////////////////////////////////////////////////////////////////////////////
// Tests by property
////////////////////////////////////////////////////////////////////////////////

update to 

/**
 * Tests by property
 */

```

kresimir-coko:
Sounds good, I'll test in DXP and remove the exception. Glad that this rule works for us! 😄

kresimir-coko:
Seems to work as intended in DXP 🍾 

image

kresimir-coko:
🤔 not sure why `format / check (12.x) (pull_request)` is failing. I did make a mistake initially but then force-pushed a fix, which shouldn't have any formatting mistakes, I think. Can someone give it a look 👀

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/__tests__/index.test.js
```diff
@@ -10,9 +10,9 @@ import path from 'path';
 import PkgDesc from '../../pkg-desc';
 import {Project} from '../index';
 
-////////////////////////////////////////////////////////////////////////////////
-// Tests by property
-////////////////////////////////////////////////////////////////////////////////
+/**
+* Tests by property
```
izaera, line 14: This is missing a space to the left. Maybe that's the cause for the SF failure...

kresimir-coko: 🤦

bryceosterhaus:
@kresimir-coko did you get an idea of how many errors would be in portal? It'd be nice to see the total amount of errors before we merge, just in case it is a crazy amount.

kresimir-coko:
> @bryceosterhaus  did you get an idea of how many errors would be in portal?

Here we go, it's not a lot, and not all of these are related to this rule, some are shorthands and unused args. I'd say maximum of 20 errors (most likely between 10 and 15) are related to the comments.

```
✖ 30 problems (30 errors, 0 warnings)
  22 errors and 0 warnings potentially fixable with the `--fix` option.
```

bryceosterhaus:
@kresimir-coko good to know! Thanks for checking that. I'll merge this and likely release/merge to DXP after the createPortal and abbreviation rules are added

PR #569, opened 2021-06-04
fix(js-toolkit): support spread operator when bundling modules

### LINKED ISSUE #565: Liferay Bundler/Babel doesn't understand the spread-operator
> 👀 Please check the [troubleshooting guide](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/How-to-troubleshoot-your-setups.md) before reporting anything. It contains important information on how to fix or diagnose errors.

-   [ ] :thinking: Question
-   [x] :bug: Bug report

### Version (mark with `x`)

-   [x] :two: v2.x
-   [ ] :three: v3.x

### Description
I'm trying to use [Recoil ](https://recoiljs.org/docs/introduction/getting-started) for State-management. Now recoil uses the spread operator and babel fails then when it tries to add the Liferay.bundler.define around it, because it doesn't know the token.

**Desired behavior:**
The liferay-bundler can parse files with the spread-operator

**Current behavior:**
The liferay-bundler ignores these files

**Repro instructions (if applicable):**

1. install recoil
2. import it
3. build it
4. check the builded files. there should be no Liferay.Loader.define around it.

Liferay 7.3.6
es6-fail

Maybe there are more ES6 features that would also be nice that babel understands it, but at the moment I think the spread-operator is the important one.

izaera:
Fixes #565

Tested locally with the same project I used to reproduce it (an empty shared bundle with just `recoil` dependency)

### DIFF
--- a/maintenance/projects/js-toolkit/packages/babel-preset-liferay-standard/package.json
+++ b/maintenance/projects/js-toolkit/packages/babel-preset-liferay-standard/package.json
@@ -9,6 +9,7 @@
 		"babel-plugin-namespace-modules": "2.24.3",
 		"babel-plugin-normalize-requires": "2.24.3",
 		"babel-plugin-transform-node-env-inline": "0.4.3",
+		"babel-plugin-transform-object-rest-spread": "6.26.0",
 		"babel-plugin-wrap-modules-amd": "2.24.3"
 	},
 	"description": "Babel preset for bundling standard Liferay projects.",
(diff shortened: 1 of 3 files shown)


bryceosterhaus:
LGTM!

Are there any other common syntaxes in modern browsers that we might want to add support for? I didn't realize some popular packages like this were starting to not transpile spread operators.

izaera:
:thinking: Not sure. Probably fat arrow... In general, AST parsers are lenient with most of the new features because they fit in the older syntax scheme, but some newer things that are not backward compatible (like this one) may make it fail.

I will investigate fat arrow too :+1:

izaera:
Fat arrow works, so going to merge because I cannot think of any other risky construct now :thinking:

PR #574, opened 2021-06-07
fix(changelog-generator): don't fail when no version is present

izaera:
Found when trying to make a release of JS Toolkit v2.

----

When the version field is not present in package.json and --interactive is
used the tools fails without a meaningful error message.

This fix makes it show a proper message.

### DIFF
--- a/projects/npm-tools/packages/changelog-generator/src/index.js
+++ b/projects/npm-tools/packages/changelog-generator/src/index.js
@@ -238,14 +238,18 @@ async function getVersion(options) {
 		let currentVersion;
 
 		try {
-			({version: currentVersion} = JSON.parse(
-				await readFileAsync('package.json', 'utf8')
-			));
+			const packageJson = JSON.parse(
+				fs.readFileSync('package.json', 'utf8')
+			);
+
+			currentVersion = packageJson.version;
 
 			if (!currentVersion && fs.existsSync('lerna.json')) {
-				({version: currentVersion} = JSON.parse(
-					await readFileAsync('lerna.json', 'utf8')
-				));
+				const lernaJson = JSON.parse(
+					fs.readFileSync('lerna.json', 'utf8')
+				);
+
+				currentVersion = lernaJson.version;
 			}
 		}
 		catch (_error) {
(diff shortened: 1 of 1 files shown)


izaera:
I've added another commit with an enhancement for lerna monorepos (JS Toolkit v2 is a lerna monorepo)

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -245,11 +245,21 @@ async function getVersion(options) {
...
+				({version: currentVersion} = JSON.parse(
+					await readFileAsync('lerna.json', 'utf8')
+				));
+			}
 		}
 		catch (_error) {
-			currentVersion = '';
+
+			// Ignore
+
```
kresimir-coko, line 258: Any specific reason for the empty rows here? @izaera is it just SF things?

bryceosterhaus: Yeah, I believe we require line breaks before and after comments. I believe its a rule that comes from DXP

projects/npm-tools/packages/changelog-generator/src/index.js
```diff
@@ -245,11 +245,21 @@ async function getVersion(options) {
 			({version: currentVersion} = JSON.parse(
 				await readFileAsync('package.json', 'utf8')
 			));
+
+			if (!currentVersion && fs.existsSync('lerna.json')) {
+				({version: currentVersion} = JSON.parse(
+					await readFileAsync('lerna.json', 'utf8')
+				));
```
bryceosterhaus, line 252: Not really sure if this is just me (likely is just me...), but this syntax seems hard to read. Would it make sense to update it to this?

```suggestion
				currentVersion = JSON.parse(
					await readFileAsync('lerna.json', 'utf8')
				).version;
```

projects/npm-tools/packages/changelog-generator/src/index.js
izaera: Yep. It's the SF, that adds empty lines around comments.

projects/npm-tools/packages/changelog-generator/src/index.js
izaera: I copied it from above, but I don't like it either, unless it is prependend with a `let` or a `const`, in which case, it is more understable because you have a context for destructuring.

So (for me):

````javascript
const {version: currentVersion} = json;  :+1: 

({version: currentVersion) = json;) :-1: 
```

Do we all agree? :thinking:

izaera:
Pushed changes related to feedback.

bryceosterhaus:
LGTM!

Issue #553, opened 2021-05-21
Add eslint rule to warn against using `createPortal` directly from React


bryceosterhaus:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [x] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Description

We added a new ReactPortal component to DXP and we want to make sure developers use that instead of createPortal. This new component gives us a bit more control for making sure developers aren't breaking out of the document flow when it's not necessary.

PR #575, opened 2021-06-07
feat(@liferay/eslint-config): add rule to prevent developers from using createPortal directly from react-dom.

bryceosterhaus:
Fixes https://github.com/liferay/liferay-frontend-projects/issues/553

Aim of this is to aim people towards using our component in DXP. 

This feature shouldn't be added to DXP until https://github.com/liferay-frontend/liferay-portal/pull/1091 is merged

### DIFF
--- a/projects/eslint-config/plugins/portal/lib/rules/no-react-dom-create-portal.js
+++ b/projects/eslint-config/plugins/portal/lib/rules/no-react-dom-create-portal.js
@@ -4,9 +4,9 @@
  */
 
 const DESCRIPTION =
-	'Direct use of ReactDOM.render is discouraged; instead, use ' +
-	'the <react:component /> JSP taglib, or do: ' +
-	`import {render} from 'frontend-js-react-web';`;
+	'Direct use of ReactDOM.createPortal is discouraged; instead, use ' +
+	'the <ReactPortal /> component: ' +
+	`import {ReactPortal} from 'frontend-js-react-web';`;
 
 module.exports = {
 	create(context) {
@@ -47,7 +47,7 @@ module.exports = {
 
 		const report = (node) =>
 			context.report({
-				messageId: 'noReactDOMRender',
+				messageId: 'noReactDOMCreatePortal',
 				node,
 			});
 
@@ -73,10 +73,10 @@ module.exports = {
 						if (
 							node.callee.object.name === namespace.name &&
 							node.callee.property.type === 'Identifier' &&
-							node.callee.property.name === 'render'
+							node.callee.property.name === 'createPortal'
 						) {
 
-							// eg. Foo.render()
+							// eg. Foo.createPortal()
 
 							if (isSame(node.callee.object, namespace)) {
 
@@ -144,7 +144,7 @@ module.exports = {
 									variable.defs[0].node &&
 									variable.defs[0].node.imported &&
 									variable.defs[0].node.imported.name ===
-										'render'
+										'createPortal'
 								);
 							})
 					);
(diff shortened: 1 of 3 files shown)


projects/eslint-config/plugins/portal/lib/rules/no-react-dom-create-portal.js
```diff
@@ -0,0 +1,213 @@
...
+					add(
+						foundBindings,
+						context
+							.getDeclaredVariables(node)
+							.filter((variable) => {
+								return (
+									variable.defs[0] &&
+									variable.defs[0].node &&
+									variable.defs[0].node.imported &&
+									variable.defs[0].node.imported.name ===
```
kresimir-coko, line 146: Perhaps we can give a friendly name to the `variable.defs[0]` and that way store it only once?

projects/eslint-config/plugins/portal/lib/rules/no-react-dom-create-portal.js
```diff
@@ -0,0 +1,213 @@
...
+					node.init &&
+					node.init.type === 'CallExpression' &&
+					node.init.callee.type === 'Identifier' &&
+					node.init.callee.name === 'require' &&
+					node.init.arguments.length > 0 &&
+					node.init.arguments[0].type === 'Literal' &&
+					node.init.arguments[0].value === 'react-dom'
+				) {
+					const variables = context.getDeclaredVariables(node);
+					if (node.id.type === 'Identifier') {
```
kresimir-coko, line 165: Empty line in-between perhaps?

projects/eslint-config/plugins/portal/tests/lib/rules/no-react-dom-create-portal.js
```diff
@@ -0,0 +1,184 @@
+/**
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+const MultiTester = require('../../../../../scripts/MultiTester');
+const rule = require('../../../lib/rules/no-react-dom-create-portal');
+
+const parserOptions = {
+	ecmaFeatures: {
```
kresimir-coko, line 10: Isn't ecma an abbreviation? Wouldn't `ECMA` be more appropriate?

kresimir-coko (commented):
Out of my depth here so I just provided some minor things that I could notice.

projects/eslint-config/plugins/portal/lib/rules/no-react-dom-create-portal.js
bryceosterhaus: This is copied from `no-react-dom-render.js` so I think I will just leave it as is for now so that we don't have to re-write the other. This code is used to check for `import {x as y} from 'foo-bar';`. We take the definition (`y`) and check the name of its import (`x`). This way people can't circumnavigate defining something like `import {createPortal as fooCreatePortal} from 'react-dom';`

projects/eslint-config/plugins/portal/tests/lib/rules/no-react-dom-create-portal.js
bryceosterhaus: This is an option passed to babel-eslint, so we can't update this key

izaera: In any case, when it is at the start of the word, we usually lowercase it (at least in portal...)

bryceosterhaus:
gonna merge but wait on new eslint release until GA2 is out

PR #576, opened 2021-06-08
feat(js-toolkit): automatically exclude transitive dependencies from autopresets (target platforms)

### LINKED ISSUE #566: Find a way to automatically exclude transitive dependencies from autopresets (target platforms)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report

### Version (mark with `x`)

-   [X] :two: v2.x
-   [ ] :three: v3.x

### Description

With the target platform feature we are making projects depend on a single dependency that transitively brings the whole portal and the devtools needed for the build (babel, bundler, etc.).

Because we add those as `dependencies`, not `devDependencies` (because otherwise the project wouldn't install them in the `node_modules` folder) the bundler thinks they may be needed at runtime and tries to bundle them, leading to huge build times.

The way to avoid this is to use the [exclude](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/.npmbundlerrc-file-reference.md#exclude) config option. However, because we may use semver expressions for devtools, it is possible that the list of exclusion we write today doesn't filter all the deps brought tomorrow (for example if a new version of babel appears that pulls more deps) so we need a way to dinaymically filter the devtool deps brought by the target platform.

So, the best option seems to use a bundler copy plugin (smilar to [liferay-npm-bundler-plugin-exclude-imports](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit/packages/liferay-npm-bundler-plugin-exclude-imports)) that is smart enough to find devtools deps and exclude them from the build (unless the user has explicitly added them as deps to the project's package.json).

izaera:
This is chunk 1 of https://github.com/liferay/liferay-frontend-projects/pull/562

Fixes #566

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
@@ -25,6 +25,12 @@ import {VersionInfo} from './types';
 /** A package manager */
 export type PkgManager = 'npm' | 'yarn' | null;
 
+/** Information on the preset being used */
+export interface PresetInfo {
+	name: string;
+	isAutopreset: boolean;
+}
+
 /**
  * Describes a standard JS Toolkit project.
  */
@@ -175,6 +181,13 @@ export class Project {
 		}
 
 		return this._pkgManager;
+	}
+
+	/**
+	 * Get information about the preset in use
+	 */
+	get presetInfo(): PresetInfo | undefined {
+		return this._presetInfo;
 	}
 
 	/**
@@ -362,10 +375,10 @@ export class Project {
 		}
 	}
 
-	_findAutopresets(): string[] {
+	_getAutopreset(): string | undefined {
 		const {dependencies = {}, devDependencies = {}} = this._pkgJson;
 
-		return Object.keys({
+		const autopresets = Object.keys({
 			...dependencies,
 			...devDependencies,
 		}).reduce((autoPresets, pkgName) => {
@@ -384,47 +397,18 @@ export class Project {
 
 			return autoPresets;
 		}, []);
-	}
 
-	_loadDefaultPreset(): string {
-		let presetFilePath;
-
-		const autoPresets = this._findAutopresets();
-
-		if (autoPresets.length > 1) {
+		if (autopresets.length > 1) {
 			throw new Error(
 				'Multiple autopreset dependencies found in project (' +
-					autoPresets +
+					autopresets +
 					'): please remove the invalid ones or ' +
 					'explicitly define the preset to be used in the ' +
 					'.npmbundlerrc file'
 			);
 		}
-		else if (autoPresets.length) {
-			presetFilePath = this.resolve(autoPresets[0]);
 
-			this._toolsDir = new FilePath(
-				path.dirname(this.resolve(autoPresets[0] + '/package.json'))
-			);
-		}
-		else {
-
-			// If no preset was found, use the default one
-
-			presetFilePath = require.resolve(
-				'liferay-npm-bundler-preset-standard'
-			);
-
-			this._toolsDir = new FilePath(
-				path.dirname(
-					require.resolve(
-						'liferay-npm-bundler-preset-standard/package.json'
-					)
-				)
-			);
-		}
-
-		return presetFilePath;
+		return autopresets.length ? autopresets[0] : undefined;
 	}
 
 	_loadNpmbundlerrc(): void {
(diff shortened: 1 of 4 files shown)


izaera:
Once this is merged, we may release a new version of the JS toolkit to be used for target platforms.

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -362,18 +375,18 @@ export class Project {
...
 	}
 
-	_findAutopresets(): string[] {
+	_getAutopreset(): string | null {
 		const {dependencies = {}, devDependencies = {}} = this._pkgJson;
 
-		return Object.keys({
+		const autopresets = Object.keys({
 			...dependencies,
 			...devDependencies,
```
kresimir-coko, line 383: Are these the autopresets?
```
Object.keys({
	...dependencies,
	...devDependencies,
)};
```
Or are the reduced object keys the autopresets, just nitpicking here to perhaps reduce chaining idk 🤷

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -382,38 +395,57 @@ export class Project {
...
-		if (autoPresets.length > 1) {
+		if (autopresets.length > 1) {
 			throw new Error(
-				'Multiple autopreset dependencies found in project (' +
-					autoPresets +
-					'): please remove the invalid ones or ' +
+				'Multiple autopreset dependencies found in project ' +
+					`(${autopresets}): please remove the invalid ones or ` +
 					'explicitly define the preset to be used in the ' +
 					'.npmbundlerrc file'
```
kresimir-coko, line 406: You didn't add template literals across the whole string because of the line breaking, right?

kresimir-coko (approved):
LNHTM 👍

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: They are the package names of all dependencies that are autopresets (i.e: have `liferay-npm-bundler` as a dependency).

:thinking: Maybe we should add another criterion to detect an autopreset, namely that it is a valid preset (its main index is a JSON file and it looks like a preset).

I will add it, I think :+1:

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: Yes. We could use them, but then we would need to move new lines to the left to remove the indent and it Looks Horrible :tm:

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -25,6 +25,12 @@ import {VersionInfo} from './types';
 /** A package manager */
 export type PkgManager = 'npm' | 'yarn' | null;
 
+/** Information on the preset being used */
+export interface PresetInfo {
+	name: string;
+	isAutopreset: boolean;
```
bryceosterhaus, line 31: nit: portal would complain about sorting here... I am less concerned about it, but I'll create [this issue](https://github.com/liferay/liferay-frontend-projects/issues/579) to investigate an eslint plugin to sort this for us.

bryceosterhaus (approved):
LGTM! I just had one nitpick but we can address that in the issue I linked

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: No need to investigate anything: linting is disabled in `maintenance/js-toolkit`, that's why it doesn't complain (I noticed it yesterday). I was thinking in fixing as much as I can, send a PR, and activate it. Otherwise, it will keep drifting and get to a point where it's impossible to fix it.

izaera: Created [this issue](https://github.com/liferay/liferay-frontend-projects/issues/580) to activate eslint in toolkit v2

izaera: Not sure about my previous comment. The problem I was seeing was due to be trying to run `eslint` from the project directory when it needs to be run from the root project of `liferay-frontend-projects` because otherwise the configuration are not correctly merged.

However, I still don't understand why this specific rule doesn't kick in. Maybe because it is a type declaration instead of an object? :thinking: I guess it's that.

Sorry for the noise :grimacing:

bryceosterhaus: Yeah I think it must be because its a type delcaration. I noticed something similar in Clay awhile back but forgot about it till now

PR #577, opened 2021-06-08
docs(gh-jira-sync): update exalate scripts to remove Ambrin

brunofernandezg:
Update the user map because Ambrin is leaving FI


brunofernandezg:
Not sure if it's correct. At first I forgot to fetch the origin, the result should be alphabetically ordered and without Ambrin. Not sure why in files changed appear as if I was reordering again because master should be already sorted.

brunofernandezg:
Great, thanks Bryce!

PR #581, opened 2021-06-09
chore: let `eslint` be executed from inner project folders

izaera:
Turns out that `.eslintignore` files are not aggregated the same as
`.eslintrc.js` files, according to the docs:

- https://eslint.org/docs/user-guide/configuring/ignoring-code#ignorepatterns-in-config-files
- https://eslint.org/docs/user-guide/configuring/ignoring-code#the-eslintignore-file

This makes impossible running `eslint` from any other folder but the root
of `liferay-frontend-projects` because it fails to load some ignores and
that causes false positives. We don't notice this under normal conditions
because we always run `yarn lint`, which causes the whole repo to be linted.

However, I -being an irrational person- wanted to integrate the linting
into `vim` by means of a custom small plugin I'm writing. This is because
it will let me run `eslint` from `vim` and jump to the errors directly,
hopefully leading to an increase in performance.

This custom plugin runs `eslint` for single files as well as projects or
workspaces, so that I can save time, instead of waiting several seconds for
`yarn lint` to finish, so I need `eslint` to be run from any location (not
only the workspace folder), thus I'm pushing this change.

The configuration of ignores inside `.eslintrc.js` is equivalent to that of
`.eslintignore`, but it can be inherited, so I think that, even in the
absence of my plugin, the change is a step forward.

In fact, we probably shouldn't be using `.eslintignore` at all, but
`ignorePatterns` inside config files...


izaera:
Lint is green, so it should be safe to merge. If the change were incorrect, we would get a bunch of linting errors.

bryceosterhaus:
> Lint is green, so it should be safe to merge. If changes were incorrect, we would get a bunch of linting errors.

PR #583, opened 2021-06-10
docs(guidelines): Add docs on extracting JS from JSPs

### LINKED ISSUE #344: Document best practices of extracting js from the jsp
We have a lot of js inside the JPS's, this has a lot of downsides (code is not transpiled,  logic is mixed with jsp stuff making the code difficult to follow, etc.)

It would be nice to document the best practices in this regard, if you are using a visual component the option of using the soy:component-renderer is clear, but what is the option in the case you are not using javascript to render the page, how this cases can be extracted to a different file?

Thanks!

kresimir-coko:
Fixes #344 

Let me know if I missed something or misrepresented some information!

### DIFF
--- a/guidelines/dxp/extracting_js_from_jsps.md
+++ b/guidelines/dxp/extracting_js_from_jsps.md
@@ -1,0 +1,56 @@
+# Extracting JS from JSPs
+
+As we're modernizing our codebase across DXP we are starting to encounter cases where we are touching JS code inside JSPs and often times we have to remind ourselves how best to tackle that.
+
+We have two options when extracting JS from JSPs:
+
+-   `<liferay-frontend:component>` - good old JS that doesn't render UI
+-   `<react:component>`- markup with interactive behaviour
+
+The `context` and `props` arguments passed to the tag are of type `Map<String, Object>` built using the `HashMapBuilder.<String, Object>`.
+
+## Using `<liferay-frontend:component>`
+
+Whenever you have some JS code inside the `<aui:script>` you most likely want to use this tag. It allows you to easily send data (`context`) into the scope of your JS function, it separates concerns and leads to cleaner code.
+
+### JSP
+
+```javascript
+<liferay-frontend:component
+	componentId="Foo" // If your JS code is in form of a Class, this is where you specify the Class name
+	context="<%= context %>" // Data you want to send from the JSP to the JS
+	module="foo-bar/js/foo.js" // Relative path to your new JS file
+/>
+```
+
+## Using `<react:component>`
+
+If the JS you want to extract is rendering some HTML, you should use a React component and the `<react:component>` tag. Similar to the `<liferay-frontend:component>` tag, it allows you to send data (`props`) and it gives you the ability to create a React component inside the file you specify as value of the `module` argument.
+
+### JSP
+
+```javascript
+<react:component
+	module="foo-bar/js/FooBar" // Relative path to the React component
+	props="<%= props %>" // Data you want to send from the JSP to the JS
+/>
+```
+
+## Javascript
+
+Both the React and non-React component accept the `context` in the same manner, by destructuring the object properties as seen in the example below.
+
+```javascript
+export default function ({
+	contextPropertyFoo,
+	contextPropertyBar,
+	contextPropertyFooBar,
+}) {
+	// Code has access to the properties you passed in through context
+}
+```
+
+## Possible pitfalls
+
+-   Make sure to follow the file structure across the codebase when adding the new file
+-   When you have multiple functions that you want to extract, use your best judgement when determining if you should split them into separate files or export all of them inside one file


guidelines/dxp/extracting_js_from_jsps.md
```diff
@@ -0,0 +1,41 @@
+# Extracting JS from JSPs
+
+As we're modernizing our codebase across DXP we are starting to encounter cases where we are touching JS code inside JSPs and often times we have to remind ourselves how best to tackle that.
+
+We have two options when extracting JS from JSPs:
+
+-   `<liferay-frontend:component>` - good ol' JS that doesn't render UI
```
izaera, line 7: I would use `old` here, per the same reason we don't use abbreviations in source code: some people's English knowledge may be reduced and they may not understand what it means.

izaera: So, keep it boring :joy:

izaera:
LNHTM

guidelines/dxp/extracting_js_from_jsps.md
kresimir-coko: True, my inner Tolkien came through there

kresimir-coko:
I've force-pushed the typo change!

guidelines/dxp/extracting_js_from_jsps.md
izaera: incredible

guidelines/dxp/extracting_js_from_jsps.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Code Example
+
+```javascript
+<liferay-frontend:component
+	componentId="Foo" // If your JS code is in form of a Class, this is where you specify the Class name
+	context="<%= context %>" // Data you want to send from the JSP to the JS
+	module="foo-bar/js/foo.es" // Relative path to your new JS file
+/>
+```
```
bryceosterhaus, line 24: I'd also include an example of what foo.js contains and how it uses the context

guidelines/dxp/extracting_js_from_jsps.md
```diff
@@ -0,0 +1,41 @@
...
+
+Whenever you have some simpler JS code inside the `<aui:script>` you most likely want to use this tag. It allows you to easily send backend data (`context`) into the scope of your JS function, it separates concerns and leads to cleaner code.
+
+### Code Example
+
+```javascript
+<liferay-frontend:component
+	componentId="Foo" // If your JS code is in form of a Class, this is where you specify the Class name
+	context="<%= context %>" // Data you want to send from the JSP to the JS
+	module="foo-bar/js/foo.es" // Relative path to your new JS file
```
bryceosterhaus, line 22: nit: we are moving away from `.es` as they aren't necessary anymore and so it'd probably be best to have `.js` here

guidelines/dxp/extracting_js_from_jsps.md
```diff
@@ -0,0 +1,41 @@
...
+We have two options when extracting JS from JSPs:
+
+-   `<liferay-frontend:component>` - good old JS that doesn't render UI
+-   `<react:component>`- markup with interactive behaviour
+
+The `context` and `props` arguments passed to the tag are of type `Map<String, Object>` built using the `HashMapBuilder.<String, Object>`.
+
+## Using `<liferay-frontend:component>`
+
+Whenever you have some simpler JS code inside the `<aui:script>` you most likely want to use this tag. It allows you to easily send backend data (`context`) into the scope of your JS function, it separates concerns and leads to cleaner code.
```
bryceosterhaus, line 14: Whether it is simple or not, it doesn't really matter for using this taglib.

```suggestion
Whenever you have some JS code inside the `<aui:script>` you most likely want to use this tag. It allows you to easily send backend data (`context`) into the scope of your JS function, it separates concerns and leads to cleaner code.
```

guidelines/dxp/extracting_js_from_jsps.md
```diff
@@ -0,0 +1,41 @@
...
+<liferay-frontend:component
+	componentId="Foo" // If your JS code is in form of a Class, this is where you specify the Class name
+	context="<%= context %>" // Data you want to send from the JSP to the JS
+	module="foo-bar/js/foo.es" // Relative path to your new JS file
+/>
+```
+
+## Using `<react:component>`
+
+If the JS you want to extract is rendering some HTML that's a sign that you should use the `<react:component>` tag. Similar to the `<liferay-frontend:component>` tag, it allows you to send backend data (`props`) and it gives you the ability to create a React component inside the file you specify as value of the `module` argument.
```
bryceosterhaus, line 28: It doesn't have to be "backend" data. You can also just pass a string as prop.

```suggestion
If the JS you want to extract is rendering some HTML, you should use a React component and the `<react:component>` tag. Similar to the `<liferay-frontend:component>` tag, it allows you to send data (`props`) and it gives you the ability to create a React component inside the file you specify as value of the `module` argument.
```

guidelines/dxp/extracting_js_from_jsps.md
```diff
@@ -0,0 +1,41 @@
...
+```javascript
+<react:component
+	module="foo-bar/js/FooBar" // Relative path to the React component
+	props="<%= props %>" // Data you want to send from the JSP to the JS
+/>
+```
+
+## Possible pitfalls
+
+-   Make sure you have modernized the JS that you extracted. Like using `const` instead of `var`
```
bryceosterhaus, line 39: I don't think I would consider this a pitfall and can probably move it out of this file. It's likely that our linting and formatting tools will catch this for them.

bryceosterhaus (commented):
Looks good! I just had a few notes

kresimir-coko:
I messed up and force-pushed onto the same commit after sorting out all the notes you left @bryceosterhaus. Should be fine tho 👍

bryceosterhaus:
Looks good!

bryceosterhaus:
"Comment and Close" is a terrible button

PR #594, opened 2021-06-16
fix(eslint-config): add case for abbreviation being a key on an object

bryceosterhaus:
Caught this issue while trying to get this merged to portal


izaera:
Makes sense, I guess.

Time will tell if this eslint rule is worth the effort... :roll_eyes:

izaera (approved):

bryceosterhaus:
> Time will tell if this eslint rule is worth the effort...

Issue #591, opened 2021-06-15
Add `deploy` support to `@liferay/cli` projects


izaera:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [X] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [X] :two: v2.x
-   [X] :three: v3.x

### Description
We need some way to deploy projects for target platforms as we did before with `liferay-npm-build-support` for the other project types.

My initial proposal is to put this auxiliary code inside `@liferay/js-toolkit-core` and keep it as simple as possible to avoid the problems we had in previous versions of the JS Toolkit (we even had code to translate labels which AFAIK, nobody ever used and could have been extracted to a different tool).

I suggest using `@liferay/js-toolkit-core` because it will be our toolkit library/API so all shared code should go there. We attempted doing these things in a modular fashion in JS Toolkit v2 but it never proved more benefitial than the monolythic approach, so this time we may start putting everything there and extract afterwards if necessary (for example, for deprecation purposes). In any case, things in `@liferay/js-toolkit-core` should be cohesive (that rules out the possibility to include translation for labels) so that it makes sense as an API and shared dependency.

We will need to find a balance.

izaera:
As per [this conversation](https://github.com/liferay/liferay-frontend-projects/pull/584#discussion_r651324868) it is possible that we may need to move the `build` process to the same place, too.

izaera:
Turns out that we also had [js-toolkit-scripts](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/js-toolkit/packages/js-toolkit-scripts) that was an attempt to migrate [liferay-npm-build-support](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit/packages/liferay-npm-build-support) to Toolkit v3.

Thus, we now have:

1) `@liferay/cli`
2) target platform embedded scripts (`build`, `deploy`, ...)
3) `@liferay/js-toolkit-core`
4) `@liferay/js-toolkit-scripts`
5) `liferay-npm-build-support`

Which all serve the same purpose :roll_eyes: 

It would be great to ~add six or seven more parallel projects~ leave just one, but while we shift from v2 to v3 (whatever it might be) and from the yeoman generator to `@liferay/cli` we will need to have at least:

3) `@liferay/js-toolkit-core`
5) `liferay-npm-build-support`

[...]

PR #596, opened 2021-06-17
feat: add `deploy` support to `@liferay/cli` projects 

izaera:
This PR adds the `deploy` command to platform projects created with `@liferay/cli` and reorganizes the code to make future maintenance easier.

I've tested this locally using my local version of `@liferay/cli` to generate a new project, then running:

1) `yarn && yarn link @liferay/portal-7.4-ga1`
2) `yarn build && yarn deploy`

And testing that everything was OK in my liferay-portal instance.

I've also reviewed the generated files manually and made sure that gitignores were correct.


izaera:
Closes #591

target-platforms/packages/portal-base/package.json
```diff
@@ -0,0 +1,23 @@
...
+		"@babel/preset-react": "^7.13.13",
+		"babel-preset-liferay-standard": "^2.26.0",
+		"liferay-npm-bundler-loader-copy-loader": "^2.26.0",
+		"liferay-npm-bundler-loader-babel-loader": "^2.26.0",
+		"liferay-npm-bundler-plugin-exclude-imports": "^2.26.0",
+		"liferay-npm-bundler-plugin-inject-imports-dependencies": "^2.26.0",
+		"liferay-npm-bundler-plugin-inject-peer-dependencies": "^2.26.0",
+		"liferay-npm-bundler-plugin-namespace-packages": "^2.26.0",
+		"liferay-npm-bundler-plugin-replace-browser-modules": "^2.26.0",
+		"liferay-npm-bundler-plugin-resolve-linked-dependencies": "^2.26.0"
```
bryceosterhaus: What if we change a version of these dependencies only for a single target platform? Wouldn't we want target platforms to not share?

It seems like portal-base would be tied to each individual version of DXP but I may be misunderstanding the purpose of this package.

target-platforms/packages/portal-base/src/configuration.js
```diff
@@ -0,0 +1,58 @@
+/**
```
bryceosterhaus, line 1: Just to make sure, would we ever have both `.npmbuildrc` and `.liferay.json` in the same project? I know we want to reduce the number of configs we could have in a project.

Additionally, I think we might want this config to be a js file to be more in line with `npmscripts.config.js`.

target-platforms/packages/portal-base/package.json
bryceosterhaus: I think I am just slightly confused on the need/purpose of portal-base. It seems like it might be more of a hassle to publish for each new target platform and then making sure old target platforms don't rely on a newer portal-base

izaera: > What if we change a version of these dependencies only for a single target platform? Wouldn't we want target platforms to not share?

If that case happens we will need another refactor to split up again. The truth is that we will need to evolve the architecture of target platforms as time passes by. But since they are like a black box for the user and and their only point of contact is the `liferay` npm binary we can change them as much as we want/need.

> It seems like portal-base would be tied to each individual version of DXP but I may be misunderstanding the purpose of this package.

It depends. As with `npm-scripts` we can have one pseudo/real branch per DXP version or -better- try to rule them all with one single package (as we wanted to do in the future with `npm-scripts`). I would say we will be able to use the same package for every DXP version if we account for that possibility. In the end, if we upgrade the build to do more things or use newer versions of devtools, that doesn't mean we cannot target older DXP versions. See the bundler: people have been always able to use the latest version for any DXP version.

> I think I am just slightly confused on the need/purpose of portal-base. It seems like it might be more of a hassle to publish for each new target platform and then making sure old target platforms don't rely on a newer portal-base

In fact, I started target platforms with the opposite model in mind (each platform has its own code), but since Javi said Brian would need to publish platforms (not us) we cannot rely on that model because we have split responsibilities:

1) Brian will define when the platform is released and will run a script we provide to release them
2) We need to have a way to fix problems in released platforms without having to wait days for someone to release something

Per 1, we want to provide a `create-platform` script that is fast, simple and fail-safe. Something we are far from with the first `create-platform` script I created.

Per 2, we need to be able to publish new versions of portal-base, then tell the customers to upgrade their `yarn.lock` files to update their platform (without any need to release a new platform).

To sum up:

- The target platform package only has the configuration to define how a project for a given DXP version is built and what the imports are (immutable, at least the imports part)
- The portal-base package implements all the logic and can be upgraded without touching target platforms. We "simply" need to take care of making it backwards and forward compatible (something that I think won't be so difficult).

izaera: In any case, we are making a lot of assumptions here, upfront, and some may prove wrong in the future. This is an invariant in all tools development :roll_eyes:, but the good news are that, as long as you define a coherent model and control the surface of the tools (not exposing too much API) to be able to change underlying things it usually works and is maintainable. At least that has been my experience with the bundler and the AMD loader...

target-platforms/packages/portal-base/src/configuration.js
izaera: > Just to make sure, would we ever have both .npmbuildrc and .liferay.json in the same project? I know we want to reduce the number of configs we could have in a project.

Nope. I want to get rid of `.npmbuildrc` files :sweat_smile: . Old projects (those created with the current yeoman generator) will have it, but if people want to "upgrade" them, they will need to either remove the `.npmbuildrc` file or migrate it to `.liferay.json`.

> Additionally, I think we might want this config to be a js file to be more in line with npmscripts.config.js.

That's the first thing I tried, but:

1) It's impossible, because it's our tools what modify this file, and we cannot modify a `.js` programatically because we cannot guarantee that we don't break it. So, `.js` config files are very good for read-only configurations (from the tool POV) that are written by the user. But this is not the case, as I said. It's the `liferay` bin what reads and writes this file.
2) Don't think of `.liferay.json` as a config file, but more as a settings repository (similar to `.yo-repository`). The use won't interact with this file directly, but from the tools (unless needing to tweak something very low-level).

Now thinking about 2 and seeing that the only setting we have now is the Liferay deploy directory, maybe we can move this to the user's home folder (to `~/.liferay.json` or something like that). We are saying this file is for local settings that don't need to be pushed to the VCS and that's usually dirs and/or credentials, which usually will fit in the HOME directory. :thinking: 

We can maybe do a frontend similar to `git config` named `liferay config` and manipulate these files from it.

target-platforms/packages/portal-base/package.json
bryceosterhaus: Ah I see, thanks for the explanation, that helps clarify and gives good context.

> Brian would need to publish platforms (not us) we cannot rely on that model because we have split responsibilities

I hadn't totally thought about this side of things, so you are right in that it helps give us a bit more freedom to publish fixes to portal-base without having to rely bchan publishing.



> In any case, we are making a lot of assumptions here, upfront, and some may prove wrong in the future. This is an invariant in all tools development 🙄, but the good news are that, as long as you define a coherent model and control the surface of the tools (not exposing too much API), to be able to change underlying things, it usually works and is maintainable.

Yep, totally makes sense!

target-platforms/packages/portal-base/src/configuration.js
bryceosterhaus: > But this is not the case, as I said. It's the liferay bin what reads and writes this file.

Oh I see, yeah that makes sense I didn't realize users wouldn't interact with this file much and that it likely a non-committed file. I was thinking of it as analogous to the npm scripts config.

> maybe we can move this to the user's home folder

Personally, I'm not a huge fan of this option. Whenever a tool moves files outside (upwards) of the directory I am in, I find it hard to track and get annoyed. I could be convinced though if you think it is the best option.

> We can maybe do a frontend similar to git config named liferay config and manipulate these files from it...

I like this idea as long as it is not a headache to maintain. Having to remember to add options to the config tool might be annoying and easily forgettable.

izaera: OK, let's leave this as it is now and we can implement the `liferay config` tool in the future, when everything is clearer.

PR #597, opened 2021-06-17
feat(bs3-bs4-compat): add new package for bootstrap 3 to 4 compatibility layer

bryceosterhaus:
Not totally sure if this is the right place to put this package, but at least we can review it hear and decide where to put it later.

This is to help address https://issues.liferay.com/browse/LPS-125918

I yanked out the compat layer we previously had in 7.3 and now locate all that scss in this package. I don't think we can provide a 100% solution for this, but this compat layer has always been meant as a bandaid and not a permanent fix. By providing this, we at least give a place for customers to look at css they might need or they can add the entire layer by following the README directions.

Below is two screen shots comparing 7.3 (has compatibility layer) and 7.4 + this package. They are not identical but I think they are similar enough (not a css guy for the record). I don't actually think we can get a 100% replica without writing and maintaining more code, which I don't think we want to do since this is just a temporary solution for customers.

Let me know what you think! cc @pat270 @marcoscv-work

PS these screenshots are based on the markup from [here](https://liferay.github.io/lexiconcss/content/test-assorted/)

**7.3**
with-compat-7 3

**7.4 + this package**
7 4-bs3-compat

### DIFF
--- a/projects/js-themes-toolkit/README.md
+++ b/projects/js-themes-toolkit/README.md
@@ -4,6 +4,7 @@
 
 This repo contains the source for a [set of NPM packages](packages) designed to help create, update, and maintain Liferay Themes. The packages that you are most likely to interact with directly are:
 
+-   [bs3-bs4-compat](packages/bs3-bs4-compat): Bootstrap 3 to 4 Compatibility Layer for Liferay DXP. This layer is added to your theme when you run the "upgrade" task for DXP 7.4.
 -   [generator-liferay-theme](packages/generator-liferay-theme): A [Yeoman](https://yeoman.io/) generator for creating new themes, themelets, and layout templates.
 -   [liferay-theme-tasks](packages/liferay-theme-tasks): A set of [Gulp](https://gulpjs.com/) tasks for building and deploying themes.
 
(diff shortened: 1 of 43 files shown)


dsanz:
Hi @bryceosterhaus , I've put together some ideas about how could we achieve this [here](https://liferay.atlassian.net/wiki/spaces/ENGFRONTENDINFRA/pages/1569718980/7.4+Theming#Upgrade-path-proposal), please let me know your opinion.

bryceosterhaus:
Added step in upgrade task to ask user if they want to include the compat layer.

Next steps here would be having some of our css gurus to make sure the compat layer css is working as expected...

bryceosterhaus:
Next steps for this.
- Validate that this scss covers all compat use cases we were covering in 7.3
- Publish v1 of compat layer
- Publish new theme tasks

bryceosterhaus:
cc @pat270, @marcoscv-work. Would one of you guys be able to validate the css here for the compat layer? I would try, but I really don't have the proper context of what we had previously to put my stamp of approval on this being satisfactory for a compatibility layer. In order to test it in our styled and classic theme, [check out the README](https://github.com/liferay/liferay-frontend-projects/pull/597/files#diff-bdef52a56e1f458644856b9df23f6cf65b4b9749a1c9a32151d121fd9816d5f5R36-R60).

marcoscv-work:
Reviewing.. 

For reference, this is how it look like without the layer:
image

marcoscv-work (approved):
LGTM !!!

Keep in mind that the compatibility layer does not cover all the old Clay markup, but the markup that was used in portal

bryceosterhaus:
gonna merge and release the v1 compat layer

PR #605, opened 2021-07-05
docs(gh-jira-sync): update exalate scripts to remove Jon

brunofernandezg:
Only removed Jon Mak from the Exalate user map.


nhpatt:
I think it's safe to merge 😂

nhpatt (approved):

PR #607, opened 2021-07-06
feat: add Sass support to target platform projects

### LINKED ISSUE #589: Add SASS support to target platform projects

-   [ ] :thinking: Question
-   [ ] :bug: Bug report

### Version (mark with `x`)

-   [X] :two: v2.x
-   [] :three: v3.x

### Description
According to this [conversation](https://github.com/liferay/liferay-frontend-projects/pull/584#discussion_r651127821) make it possible to use SCSS in target platform projects.

izaera:
This PR adds support for Sass compilation. It will look for any file with extension `.scss` under `src` and generate a compiled version of it in `build`.

Things to be determined before merging:

1) Should all `.scss` files be rendered or only a subset of them (for example: skip files with names starting with `_`)?
2) Should we also process `.sass` and `.css` files?

This PR also supports `@import`ing Sass files from npm dependencies. The syntax is the same used by [sass-loader](https://www.npmjs.com/package/sass-loader),so for example, you could do:

```
@import '@clayui/css/src/scss/atlas';
```

If you wanted to import some scss file of Clay.

Note, however, that this differs from `liferay-portal`'s syntax, where we would write:

```
@import 'clay/atlas';
```
 
This is because `npm-scripts` [passes some include paths to `sass`](https://github.com/liferay/liferay-frontend-projects/blob/47c1411f6d85b4d419476309dcc7525600360c50/projects/npm-tools/packages/npm-scripts/src/config/npmscripts.config.js#L167). However, I think that following the standard `sass-loader` syntax is much better, since it is what people should expect and since it is an spec that avoids collisions (as opposed to the one we use in `liferay-portal`).

----

Tested locally generating a new project and deploying it to a portal instance.

Closes #589


izaera:
Set to draft because I won't merge this until points :one: and :two: of the description are clear.

izaera:
Given that `liferay-npm-scripts` [renders every `.scss` file in the project except those with names starting with `_`](https://github.com/liferay/liferay-frontend-projects/blob/47c1411f6d85b4d419476309dcc7525600360c50/projects/npm-tools/packages/npm-scripts/src/sass/build.js#L31-L49) I think we should do the same here as a starting point.

In the case of webpack, the entry points can be inferred by traversing the JavaScript files and looking for `import`s, but we can't do that because:

1) We don't traverse the dependency graph using `.js` files (i.e: we export **everything** by default, as opposed to webpack)
2) We may need css files in more places than JavaScript code (for example: portlet headers, or an external OSGi bundle)

So, I will add a commit to skip files starting with `_` and leave the rest as is.

target-platforms/packages/portal-base/src/build.js
```diff
@@ -105,7 +127,59 @@ function runBundler() {
...
+		abort(`{liferay-npm-bundler} received signal: ${signal}`);
+	}
+
+	if (status !== 0) {
+		abort(`{liferay-npm-bundler} exited with status: ${status}`);
+	}
+}
+
+function runSass() {
+	const scssFiles = findFiles(srcDir, (dirent) => {
```
bryceosterhaus, line 152: I'm not super familiar with the word `dirent`. Is this a thing or a typo?

bryceosterhaus (commented):
One question regarding this logic, and we may have discussed this previously but I have forgotten. Do we plan to consolidate or share logic between `portal-base` and `npm-scripts`? We have very similar logic for compiling sass in our [build script](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/packages/npm-scripts/src/sass/build.js).

I can see us either
- leveraging `portal-base` in npm-scripts
- Sharing logic in a central place between the two
- Leveraging npm-scripts from portal-base (although I think portal-base would likely be the more generic one)

target-platforms/packages/portal-base/src/build.js
izaera: Nope, [it does exist](https://nodejs.org/api/fs.html#fs_class_fs_dirent) (I had to read the docs when giving it a name; I usually named it `item`, but this is apparently a better name :sweat_smile: ).

izaera:
IIRC, we decided to coalesce both features in a common place. Probably a shared generic library like [js-toolkit-core](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/js-toolkit/packages/js-toolkit-core) would do it.

However I think we preferred to first develop things separately, then merge them as we see it fits, given that we don't really know how similar they will be in the end. Remember that portal's build has always been different from customers' builds because of their lifecycles and contexts.

So, I would say, ~90% of the code will be the same, but we will still need some thin layers on top of that shared code to differentiate between `npm-scripts` and target platforms...

bryceosterhaus:
> However I think we preferred to first develop things separately, then merge them as we see it fits, given that we don't really know how similar they will be in the end

👍  👍  thats what I thought and remember as well, but just wanted to make sure.

This PR looks good to me, my only suggestion would be to add some tests. You could even just copy most of them from [here](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/npm-tools/packages/npm-scripts/test/sass/build.js) and then add some tweaks

izaera:
> my only suggestion would be to add some tests. 

I think I'll add end to end tests like the ones we have in JS Toolkit, so that we can test the whole build process. They were more useful in JS Toolkit, because they serve as samples, sandboxes, and for QA.

I'll try to push them today so that you can see how it would look.

izaera:
I have added these 3 new commits:

- https://github.com/liferay/liferay-frontend-projects/pull/607/commits/deeaac4f1d05c3c20b7e42d8b097f49b04f39162
- https://github.com/liferay/liferay-frontend-projects/pull/607/commits/b02599bea2d4a695e5b23b8d513af5c67c35ac8b
- https://github.com/liferay/liferay-frontend-projects/pull/607/commits/765b4e80d13fe0c78e7e7124f7ef83c3f64c8b54

Can you please @bryceosterhaus tell me what you think?

The idea is to be able to run `yarn qa` and have a project built with target platforms, then check that the output is exactly the expected one. In my experience this is more useful than unit tests, when it comes to devtools, given how difficult they are to unit test, and even you unit test them, they tend to fail because of the final integration...

I'm pasting a typical (correct) `yarn qa` run for reference:

image

target-platforms/scripts/qa.js
```diff
@@ -0,0 +1,83 @@
...
+				.digest('hex');
+			const actualHash = createHash('sha256')
+				.update(fs.readFileSync(actualFile))
+				.digest('hex');
+
+			if (expectedHash !== actualHash) {
+				somethingChanged = true;
+				console.log('*', path.join(actualDir, expectedItem));
+			}
+		}
```
bryceosterhaus, line 50: I hadn't ever thought about using `createHash` for this, very nice!

bryceosterhaus (commented):
Looks pretty good. Nothing major needs fixing but I had a few notes

I cloned this locally and was getting failures when I ran `yarn qa` although I wasn't exactly sure why it was failing. Is is possible to show an actual diff of the failing file? I only see the message that they differ but it doesn't help me know _why_ it is failing.

target-platforms/scripts/qa.js
```diff
@@ -0,0 +1,83 @@
...
+
+run('yarn');
+
+run('yarn', 'clean');
+
+run('yarn', 'build');
+
+console.log('\n>>> diff', EXPECTED_BUILD_DIR, ACTUAL_BUILD_DIR);
+if (diff(EXPECTED_BUILD_DIR, ACTUAL_BUILD_DIR)) {
+	console.log('Builds differ :-(');
```
bryceosterhaus, line 78: Could we wrap this in newlines so that it is a bit more prominent on failure?

Without new lines
Screen Shot 2021-08-12 at 10 01 13 AM


With new lines
Screen Shot 2021-08-12 at 10 02 51 AM

target-platforms/scripts/qa.js
```diff
@@ -0,0 +1,83 @@
...
+
+const ACTUAL_BUILD_DIR = path.resolve(__dirname, '../qa/test-project/build');
+const EXPECTED_BUILD_DIR = path.resolve(
+	__dirname,
+	'../qa/test-project/build.expected'
+);
+const TEST_PROJECT_DIR = path.resolve(__dirname, '../qa/test-project');
+
+function diff(expectedDir, actualDir) {
+	var somethingChanged = false;
```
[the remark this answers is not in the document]
izaera: my god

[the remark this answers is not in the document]
diegonvs: https://media.giphy.com/media/U3ON51cwBlj4EMXKYx/giphy.gif

izaera:
> Is is possible to show an actual diff of the failing file? I only see the message that they differ but it doesn't help me know why it is failing.

I thought about that too, but then realized we needed to add more exotic devDependencies and more code that needs maintainance. So I thought that we can simply use an external tool (like [meld](https://meldmerge.org/), for example) to see the differences, once the script detects the changes. That's why I used a hash instead of a binary comparison.

izaera:
> I cloned this locally and was getting failures when I ran `yarn qa`

It would be interesting to know what can be happening before commiting all this. It is possible that the behaviour depends on the machine, the date, or something I haven't thought of. In that case, we may need to make the script smarter...

target-platforms/scripts/qa.js
izaera: :+1: done

target-platforms/scripts/qa.js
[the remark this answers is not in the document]
izaera: done :+1:

bryceosterhaus:
@izaera it looks like it has to do with the css.map.

```diff
-{"version":3,"sourceRoot":"","sources":["file:///home/ivan/Liferay/Javascript/liferay-frontend-projects/target-platforms/qa/test-project/src/css/styles.scss"],"names":[],"mappings":"AAAA;EACC;EACA;;;AAGD;EACC;EACA;;;AAGD;EACC","file":"styles.css"}
+{"version":3,"sourceRoot":"","sources":["file:///Users/bryceosterhaus/repos/liferay-frontend-projects/target-platforms/qa/test-project/src/css/styles.scss"],"names":[],"mappings":"AAAA;EACC;EACA;;;AAGD;EACC;EACA;;;AAGD;EACC","file":"styles.css"}
```

PR #614, opened 2021-08-09
chore: move devDependency to the project it is used in

bryceosterhaus:
Updates dependency version and moves the dependency to where the package is used. 

context: https://github.com/liferay/liferay-frontend-projects/pull/612


nhpatt (approved):
LGTM :)

PR #619, opened 2021-08-16
feat(npm-scripts): Adds missing Liferay global mocks for jest from `data-engine-taglib`

diegonvs:
In order to remove mocks defined in `data-engine-taglib` jest.config.js file, I decided to bring some of them to here

See https://github.com/liferay/liferay-portal/blob/5f33f077a3f08ca5c2e9872d6f9d42c234e47a37/modules/apps/data-engine/data-engine-taglib/jest-setup.config.js#L15

In order to prevent commits and errors like this https://github.com/liferay-frontend/liferay-portal/pull/1359

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/jest/mocks/AUI.js
+++ b/projects/npm-tools/packages/npm-scripts/src/jest/mocks/AUI.js
@@ -1,0 +1,14 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/* eslint-env jest */
+
+const AUI = () => ({
+	use: (key, callback) => callback(key),
+});
+
+module.exports = {
+	...AUI,
+};
--- a/projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
+++ b/projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
@@ -8,12 +8,27 @@
 const authToken = 'default-mocked-auth-token';
 
 /**
+ * https://github.com/liferay/liferay-portal/blob/21d88b8a5cbbf6bfcbb752b5312b2be4fb106761/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/component.es.js#L203
+ */
+const component = jest.fn(() => {});
+
+/**
+ * https://github.com/liferay/liferay-portal/blob/21d88b8a5cbbf6bfcbb752b5312b2be4fb106761/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/component.es.js#L305
+ */
+const destroyComponent = jest.fn();
+
+/**
  * Event support APIs on the `Liferay` object inherited from `A.Attributes`
  *
  * https://github.com/liferay/liferay-portal/blob/31073fb75fb0d3b309f9e0f921cb7a469aa2703d/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/events.js#L66
  * https://yuilibrary.com/yui/docs/api/classes/Attribute.html
  */
 const events = {
+
+	/**
+	 * https://clarle.github.io/yui3/yui/docs/api/files/event-custom_js_event-target.js.html#l850
+	 */
+	after: jest.fn(),
 
 	/**
 	 * https://yuilibrary.com/yui/docs/api/files/event-custom_js_event-target.js.html#l372
@@ -49,9 +64,38 @@ const events = {
 const Language = {
 
 	/**
+	 * https://github.com/liferay/liferay-portal/blob/f5bc2504a4f666241363a30975b6de5d57a6f627/portal-web/docroot/html/common/themes/top_js.jspf#L153
+	 */
+	available: {
+		ar_SA: 'Arabic (Saudi Arabia)',
+		ca_ES: 'Catalan (Spain)',
+		de_DE: 'German (Germany)',
+		en_US: 'English (United States)',
+		es_ES: 'Spanish (Spain)',
+		fi_FI: 'Finnish (Finland)',
+		fr_FR: 'French (France)',
+		hu_HU: 'Hungarian (Hungary)',
+		ja_JP: 'Japanese (Japan)',
+		nl_NL: 'Dutch (Netherlands)',
+		pt_BR: 'Portuguese (Brazil)',
+		sv_SE: 'Swedish (Sweden)',
+		zh_CN: 'Chinese (China)',
+	},
+
+	/**
 	 * https://github.com/liferay/liferay-portal/blob/31073fb75fb0d3b309f9e0f921cb7a469aa2703d/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/language.js#L18
 	 */
 	get: jest.fn((key) => key),
+};
+
+/**
+ * https://github.com/liferay/liferay-portal/blob/f5bc2504a4f666241363a30975b6de5d57a6f627/portal-web/docroot/html/common/themes/top_js.jspf#L161
+ */
+const PortletKeys = {
+	DOCUMENT_LIBRARY: 'DOCUMENT_LIBRARY',
+	DYNAMIC_DATA_MAPPING:
+		'com_liferay_dynamic_data_mapping_web_portlet_DDMPortlet',
+	ITEM_SELECTOR: 'ITEM_SELECTOR',
 };
 
 /**
@@ -96,6 +140,11 @@ const ThemeDisplay = {
 	getBCP47LanguageId: jest.fn(() => 'en-US'),
 
 	/**
+	 * https://github.com/liferay/liferay-portal/blob/a4866af62eb89c69ee00d0e69dbe7ff092b50048/portal-web/docroot/html/common/themes/top_js.jspf#L214
+	 */
+	getDefaultLanguageId: jest.fn(() => 'en-US'),
+
+	/**
 	 * https://github.com/liferay/liferay-portal/blob/31073fb75fb0d3b309f9e0f921cb7a469aa2703d/portal-web/docroot/html/common/themes/top_js.jspf#L217
 	 */
 	getDoAsUserIdEncoded: jest.fn(() => 'default-mocked-do-as-user-id'),
@@ -124,6 +173,11 @@ const ThemeDisplay = {
 	 * https://github.com/liferay/liferay-portal/blob/31073fb75fb0d3b309f9e0f921cb7a469aa2703d/portal-web/docroot/html/common/themes/top_js.jspf#L247
 	 */
 	getPortalURL: jest.fn(() => 'http://localhost:8080'),
+
+	/**
+	 * https://github.com/liferay/liferay-portal/blob/f5bc2504a4f666241363a30975b6de5d57a6f627/portal-web/docroot/html/common/themes/top_js.jspf#L282
+	 */
+	getScopeGroupId: jest.fn(() => '2012'),
 };
 
 /**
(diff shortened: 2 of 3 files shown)


diegonvs:
When testing in `data-engine-taglib` with an empty `jest-setup.config.js`:

Screen Shot 2021-08-16 at 18 13 31


it's working my dears

diegonvs:
moving back to "ready to review"

projects/npm-tools/packages/npm-scripts/src/jest/mocks/AUI.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * SPDX-FileCopyrightText: © 2019 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/* eslint-env jest */
+
+const AUI = () => ({
```
matuzalemsteles, line 8: 👀

projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
```diff
@@ -158,14 +223,22 @@ const Util = {
 	 * https://github.com/liferay/liferay-portal/blob/31073fb75fb0d3b309f9e0f921cb7a469aa2703d/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/util.js#L999
 	 */
 	sub: jest.fn(),
```
matuzalemsteles, line 225: It would be great to return the value it gets too, so I can remove my custom mock in a11y 🙂. Otherwise, we won't have any text being rendered.

https://github.com/liferay/liferay-portal/blob/5f33f077a3f08ca5c2e9872d6f9d42c234e47a37/modules/apps/frontend-js/frontend-js-a11y-web/jest-setup.config.js

projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
```diff
@@ -124,6 +173,11 @@ const ThemeDisplay = {
 	 * https://github.com/liferay/liferay-portal/blob/31073fb75fb0d3b309f9e0f921cb7a469aa2703d/portal-web/docroot/html/common/themes/top_js.jspf#L247
 	 */
 	getPortalURL: jest.fn(() => 'http://localhost:8080'),
+
+	/**
+	 * https://github.com/liferay/liferay-portal/blob/f5bc2504a4f666241363a30975b6de5d57a6f627/portal-web/docroot/html/common/themes/top_js.jspf#L282
+	 */
+	getScopeGroupId: jest.fn(() => '2012'),
```
izaera, line 180: I don't know if it is the case, but keep in mind that `npm-scripts` should be usable from every portal branch (7.2, 7.3, 7.4, etc.) so we are coupling this mocks to the portal itself when we use very specific values.

I'm not radically against these mocks, as long as they are generic and can be used in any portal version. For example, imagine we mock some API we have added in 7.3: that wouldn't affect 7.2 since it won't be using it.

However, if we are too specific when mocking, we may end up with a `npm-scripts` version that cannot work for two different branches of liferay-portal, going backwards in our intention of having a single npm-scripts for any portal version (note that we don't have this now, but would like to have it).

So, let's keep adding mocks for now, but keep them generic, and keep in mind that we may need to design a way to offload these mocks to liferay-portal source tree, like we did with bundler imports some time ago.

/CC @nhpatt @bryceosterhaus @kresimir-coko

nhpatt: The values are all fake, right? (maybe 2012 is not the best "fake value") because scopeGroupId changes randomly following SQL insertion order.

>So, let's keep adding mocks for now, but keep them generic, and keep in mind that we may need to design a way to offload these mocks to liferay-portal source tree, like we did with bundler imports some time ago.

👏

matuzalemsteles: > However, if we are too specific when mocking, we may end up with a npm-scripts version that cannot work for two different branches of liferay-portal, going backwards in our intention of having a single npm-scripts for any portal version (note that we don't have this now, but would like to have it).

Hey @izaera this could not become a long-term problem as we have several branches of DXP being released year-over-year, theoretically, we would always have to deal with backward compatibility, at some point we can bloat the code with too many cases. Wouldn't we also be against the idea of versioning?

projects/npm-tools/packages/npm-scripts/src/jest/mocks/AUI.js
diegonvs: 👀

diegonvs: This is unnecessary. I tried to see usages of AUI in `data-engine-taglib` but I didn't found

matuzalemsteles: Do they still use it? did you get to test without it?

projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
diegonvs: > I don't know if it is the case, but keep in mind that npm-scripts should be usable from every portal branch (7.2, 7.3, 7.4, etc.) so we are coupling these mocks to the portal itself when we use very specific values.

I didn't remember it could be used through past branches.

> However, if we are too specific when mocking, we may end up with a npm-scripts version that cannot work for two different branches of liferay-portal, going backwards in our intention of having a single npm-scripts for any portal version (note that we don't have this now, but would like to have it).

Makes sense!

> So, let's keep adding mocks for now, but keep them generic, and keep in mind that we may need to design a way to offload these mocks to liferay-portal source tree, like we did with bundler imports some time ago.

Ok :)

projects/npm-tools/packages/npm-scripts/src/jest/mocks/AUI.js
diegonvs: > Do they still use it? 

Nope

> did you get to test without it?

Yes, it's working without AUI.
Methodology:
1. Checkout master
2. Remove AUI in `jest-setup.config.js`
3. `yarn test`
4. See it working without any fail

projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
diegonvs:
```suggestion
	sub: jest.fn((str) => str),
```

Looks okay?

matuzalemsteles: Yes, it would be good to just test to see if this won't break some Portal test that wasn't expecting the return of the `sub`, in case it breaks if it's just a snapshot I think we could just update.

projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
izaera: > Hey @izaera this could not become a long-term problem as we have several branches of DXP being released year-over-year, theoretically, we would always have to deal with backward compatibility, at some point we can bloat the code with too many cases. Wouldn't we also be against the idea of versioning?

Yes to everything :-). Maybe it's time to begin extracting portal-specific things to `liferay-portal` source tree, like we did with imports, and leave only what's generic in `npm-scripts`...

I don't think we have plans for that, but the more we delay it, the more difficult it will come. Still, we may not want to have a generic `npm-scripts` but prefer to maintain pseudo-branches (it might be cheaper, maybe :shrug:)...  I would like to know what others think about it ( especially @bryceosterhaus, @nhpatt, and maybe @jbalsas).

izaera: Oh, I read "this would become a long-term problem" :-( (too early here, sorry). Why wouldn't it become a long-term problem?

The mocks are tightly coupled with the API in liferay-portal and that evolves in time, so it is subject to change. I agree that, being an API, it's probable that the mocks can be generic for all versions, but at the very least, we could end up having mocks for things that have been removed from liferay-portal but still we must have them here.

What it's clear to me is that liferay-portal specific things should go in liferay-portal source tree, not here, because that's cheating. It has been working because it's not that bad and you can always work around things, but I find it difficult to reason about a code that applies to many liferay-portal versions that are built differently without a clear abstraction.

Imagine if eslint had a rule called `use-indentation-as-in-any-liferay-portal-version-even-if-it-changes-accross-time` instead of two rules `use-indentation-with-N-spaces` and `use-indentation-with-N-tabs`, which is then used from liferay-portal and configured there. What would you prefer to maintain as an eslint developer?

matuzalemsteles: @izaera sorry for the confusion I think I ended up making it more confused but your first understanding is right that this could become a long term problem, in the end, I was wanting to ask a question 😅

Yes, I totally agree with you that this will become a problem I think my point is more related to backward compatibility with versioning.

bryceosterhaus: > I don't think we have plans for that, but the more we delay it, the more difficult it will come. Still, we may not want to have a generic npm-scripts but prefer to maintain pseudo-branches (it might be cheaper, maybe 🤷)

This is probably one of those grey areas that we might want some DXP specifics to live in npm-scripts and some specifics to be in DXP. `npm-scripts` as a whole is built to specifically address DXP and modules that live in DXP source, so it can't really be too generic. However, I do think for this specific circumstance (test mocks), it does make the most sense to keep it as close as possible to the API it is mocking. So **I think keeping the mocks with the DXP source code makes the most sense.**



One tangent thought on this, have we ever considered or discussed moving npm-scripts to DXP itself and referencing the bin directly? (I'm not necessarily advocating for it)

diegonvs: > So I think keeping the mocks with the DXP source code makes the most sense.

I think We can provide a way to tell product teams that: 

Hey, We have some predefined mocks and when using `jest-setup.config.js` in this way(We can provide a doc) you guys can extend.

jbalsas: > [...] So I think keeping the mocks with the DXP source code makes the most sense [...]

One side-effect of this is that `liferay-npm-scripts` isn't directly usable by others (Liferay Workspace, IDE, 3rd party devs) to develop following Liferay conventions. That can of course be solved in a different way and you might have already discussed that in the context of the `liferay-cli` or something else. 

Just putting it out here because we've always said we wanted to provide a good authoring experience following what we do internally. We need to be careful when moving things into the `liferay-portal` repo so we they don't get buried in there with no publication strategy.

izaera: I also thought about moving npm-script to liferay-portal, but there are two things that doesn't quite fit:

1. As Chema says, we wanted to promote npm-scripts to being a public tool that everybody can use, not only liferay-portal
2. If we put it inside liferay-portal it's going to be reviewed more by the backend teams and that can interfere in its development. I wouldn't expect it to be so bad, but definitely less agile than having it here. 

Regarding 1, it may be already fixed by target-platforms (aka liferay-cli) because we said we would move common code (for both liferay-portal and external projects) to a toolkit lib shared between the two tools.

So, maybe, once we progress in that task, we may move the leftovers of npm-scripts that don't fit in the shared lib to liferay-portal and remove it from here...

bryceosterhaus: > That can of course be solved in a different way and you might have already discussed that in the context of the liferay-cli or something else.

Yeah I think the hope is for liferay-cli and target platforms to address code not in liferay-portal source.

> So, maybe, once we progress in that task, we may move the leftovers of npm-scripts that don't fit in the shared lib to liferay-portal and remove it from here...

This sounds good to me. I like the idea of this being a "slow and steady" process of bringing publicly available stuff to liferay-cli and the liferay-portal specific stuff into liferay-portal itself. I think we are getting there and it will continue to get better and better.

izaera:
Hey @diegonvs . What are we going to do with this PR? Should we close it? Merge?

kresimir-coko:
Closed because it's a duplicate of https://github.com/liferay/liferay-frontend-projects/pull/948

PR #620, opened 2021-08-17
feat(eslint-config): Add lint to catch common AUI utilities

kresimir-coko:
Hey @bryceosterhaus I've added the initial code for the no-aui eslint rule.

I tried testing it in DXP but I didn't succeed. I asked @izaera about it, and he suggested copy/pasting the source folder into the `dxp/modules/node_modules/@liferay`, I did that and ran `gradlew formatSource` and `yarn formatSource` from an arbitrary folder that had usage of `A.one` but it didn't react.

So I'm wondering 1st of all is the code in the rule correct, I think it is. And the other thing I'm interested in, how to test it in DXP?

### DIFF
--- a/projects/eslint-config/aui.js
+++ b/projects/eslint-config/aui.js
@@ -5,18 +5,21 @@
 
 'use strict';
 
+const local = require('./utils/local');
+
 const config = {
+	plugins: [local('@liferay/aui')],
 	rules: {
-		'@liferay/aui/no-all': 'off',
-		'@liferay/aui/no-array': 'off',
-		'@liferay/aui/no-each': 'off',
-		'@liferay/aui/no-get-body': 'off',
-		'@liferay/aui/no-io': 'off',
-		'@liferay/aui/no-merge': 'off',
-		'@liferay/aui/no-modal': 'off',
-		'@liferay/aui/no-node': 'off',
-		'@liferay/aui/no-object': 'off',
-		'@liferay/aui/no-one': 'off',
+		'@liferay/aui/no-all': 'error',
+		'@liferay/aui/no-array': 'error',
+		'@liferay/aui/no-each': 'error',
+		'@liferay/aui/no-get-body': 'error',
+		'@liferay/aui/no-io': 'error',
+		'@liferay/aui/no-merge': 'error',
+		'@liferay/aui/no-modal': 'error',
+		'@liferay/aui/no-node': 'error',
+		'@liferay/aui/no-object': 'error',
+		'@liferay/aui/no-one': 'error',
 	},
 };
 
--- a/projects/eslint-config/aui.js
+++ b/projects/eslint-config/aui.js
@@ -1,0 +1,27 @@
+/**
+ * SPDX-FileCopyrightText: © 2021 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+'use strict';
+
+const local = require('./utils/local');
+
+const config = {
+	extends: [require.resolve('./react')],
+	plugins: [local('@liferay/portal')],
+	rules: {
+		'@liferay/aui/no-all': 'error',
+		'@liferay/aui/no-array': 'error',
+		'@liferay/aui/no-each': 'error',
+		'@liferay/aui/no-get-body': 'error',
+		'@liferay/aui/no-io': 'error',
+		'@liferay/aui/no-merge': 'error',
+		'@liferay/aui/no-modal': 'error',
+		'@liferay/aui/no-node': 'error',
+		'@liferay/aui/no-object': 'error',
+		'@liferay/aui/no-one': 'error',
+	},
+};
+
+module.exports = config;
--- a/projects/eslint-config/aui.js
+++ b/projects/eslint-config/aui.js
@@ -5,22 +5,18 @@
 
 'use strict';
 
-const local = require('./utils/local');
-
 const config = {
-	extends: [require.resolve('./react')],
-	plugins: [local('@liferay/portal')],
 	rules: {
-		'@liferay/aui/no-all': 'error',
-		'@liferay/aui/no-array': 'error',
-		'@liferay/aui/no-each': 'error',
-		'@liferay/aui/no-get-body': 'error',
-		'@liferay/aui/no-io': 'error',
-		'@liferay/aui/no-merge': 'error',
-		'@liferay/aui/no-modal': 'error',
-		'@liferay/aui/no-node': 'error',
-		'@liferay/aui/no-object': 'error',
-		'@liferay/aui/no-one': 'error',
+		'@liferay/aui/no-all': 'off',
+		'@liferay/aui/no-array': 'off',
+		'@liferay/aui/no-each': 'off',
+		'@liferay/aui/no-get-body': 'off',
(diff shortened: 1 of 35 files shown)


bryceosterhaus:
@kresimir-coko how were you testing your code? Did you use [ASTExplorer](https://astexplorer.net/)?

These are the options you need to have to use it for eslint
Screen Shot 2021-08-17 at 11 08 03 AM

**ASTExplorer panels**
Top left corner is your "source code"
Bottom left corner is your "eslint code"
Top right corner is the AST
Bottom left corner is the output of eslint checking the source code

projects/eslint-config/plugins/liferay/lib/rules/no-aui.js
```diff
@@ -0,0 +1,81 @@
...
+							break;
+						case 'Node':
+							preciseMessage = MESSAGE_NODE;
+							break;
+						case 'Object':
+							preciseMessage = MESSAGE_OBJECT;
+							break;
+						default:
+							preciseMessage = message;
+					}
```
bryceosterhaus, line 59: We shouldn't disallow all uses of AUI since there are cases that we don't have a replacement for yet.

projects/eslint-config/plugins/liferay/tests/lib/rules/no-aui.js
```diff
@@ -0,0 +1,41 @@
...
+const parserOptions = {
+	parserOptions: {
+		ecmaVersion: 6,
+		sourceType: 'module',
+	},
+};
+
+const ruleTester = new MultiTester(parserOptions);
+
+ruleTester.run('no-aui', rule, {
```
bryceosterhaus, line 18: Lets make sure to add a test case for each scenario.

bryceosterhaus (commented):
Hey @kresimir-coko, this is starting to look good. One thing we need to think about is how to apply this rule to only _new code_. If we merged this as is, portal would emit more warnings than we could handle. I'm not totally sure how we should scope it to only new code, but maybe you could investigate that a bit.

You could also run this on existing portal and get an idea of how many errors we would have, maybe it isn't as bad as I think it would be.

kresimir-coko:
> You could also run this on existing portal and get an idea of how many errors we would have, maybe it isn't as bad as I think it would be.

That's how I was testing, there are 740 errors right now.

kresimir-coko:
Idk why all these checks are failing, but it's happening on master as well as on my branch `Cannot find module '@liferay/js-toolkit-core' or its corresponding type declarations.` seems to be the culprit, but not sure how to fix it. It mentioned running `npm install` so I did that, and then that required running `yarn`and it still throws the same error, but without the advice for running `npm install`.

bryceosterhaus:
> there are 740 errors right now.



As for checking against new code only... I don't know if it is possible to do without only running it on the diff between HEAD and current branch. Since that isn't ideal, I think it might be ideal to add one at a time and make the changes that way. Did any of the rules have significantly less errors than the others? For example, we could probably update all the A.getBody calls since I imagine there aren't a crazy amount of those. I am trying to be cautious that this rule doesn't turn into a "remove AUI completely" epic 😂 

What do you think?




> Idk why all these checks are failing, but it's happening on master as well as on my branch Cannot find module '@liferay/js-toolkit-core' or its corresponding type declarations.

If you try running `yarn build` and then `yarn test` locally, does that fix it? I remember seeing this in the past and thought that solved it

bryceosterhaus:
And from a code structure standpoint, could we move this to be a plugin and then separate each use case into its own rule? This will allow us to enable/disable individual rules

plugins -> https://github.com/liferay/liferay-frontend-projects/tree/master/projects/eslint-config/plugins

projects/eslint-config/plugins/liferay/lib/rules/no-aui.js
```diff
@@ -0,0 +1,122 @@
...
+const MESSAGE_SELECTOR = 'use document.querySelector instead';
+
+module.exports = {
+	create(context) {
+		return {
+			MemberExpression(node) {
+				const nodeObjectName = node.object.name;
+				const nodePropertyName = node.property.name;
+
+				if (
```
diegonvs, line 26: Hey Kreso, What do you think in using an conditional just checking `nodeObjectName === 'A'` and inside it, create a switch block checking `nodePropertyName` values?

kresimir-coko: I had that initially @diegonvs, but that required me to have a `default` case, which is bad because we don't want to check against all usages of `A`, just these specific ones.

If I removed the `default` case, then lint would complain about not having an error message for all the cases.

So I decided to go with this ugly implementation, because the `switch case` essentially required the `default` case.

kresimir-coko:
> @bryceosterhaus I am trying to be cautious that this rule doesn't turn into a "remove AUI completely" epic 😂

Yea,  it's starting to be very close to it 😅

kresimir-coko:
@bryceosterhaus I've reworked these rules, and fixed some inconsistencies with them, with the goal of changing the structure to a plugin, like you suggested. Due to this being a big, structural change, **I force-pushed the changes** and essentially deleted the old commits, as it would just be confusing since it has no real relation to the current variant.

projects/eslint-config/aui.js
```diff
@@ -0,0 +1,27 @@
...
+		'@liferay/aui/no-all': 'error',
+		'@liferay/aui/no-array': 'error',
+		'@liferay/aui/no-each': 'error',
+		'@liferay/aui/no-get-body': 'error',
+		'@liferay/aui/no-io': 'error',
+		'@liferay/aui/no-merge': 'error',
+		'@liferay/aui/no-modal': 'error',
+		'@liferay/aui/no-node': 'error',
+		'@liferay/aui/no-object': 'error',
+		'@liferay/aui/no-one': 'error',
```
bryceosterhaus, line 23: Let's disable these by default. Once we start to enable them in DXP and update the code, we can slowly enable them by default

projects/eslint-config/aui.js
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+'use strict';
+
+const local = require('./utils/local');
+
+const config = {
+	extends: [require.resolve('./react')],
+	plugins: [local('@liferay/portal')],
```
bryceosterhaus, line 12: Are these two configs necessary?

bryceosterhaus (changes requested):
Looks pretty good! Only change before merging would be disabling these rules by default and make it an opt-in rule only for now. This allows us to enable it in DXP when we we actually update the DXP source

projects/eslint-config/aui.js
kresimir-coko: Probably not, I don't remember adding these 🤔

kresimir-coko:
@bryceosterhaus I've removed the 2 lines as you recommended, added `aui.js` to the list of files inside `package.json`, and turned the rules off.

I tried testing them (before turning them off) inside DXP by copy-pasting the `eslint-config` folder into `modules`, as I did with the previous iteration of the `no-aui` rule, but it doesn't seem to activate when running `yarn checkFormat`. Did I mess up the config perhaps?

PR #622, opened 2021-08-24
feat(eslint-config): add an ESlint rule that adds an empty line between sibling elements

kresimir-coko:
## Overview
With ASTExplorer finally working, I had the chance to make this rule actually work... kinda.

For code like this, it throws out an error, as it's supposed to, on lines where the `div` and `small` are opening:
```js
<label className="form-group-autofit">
	<span className="form-group-item">
		{label}
		{subTitle && (
			<span className="font-weight-normal form-text">
				{subTitle}
			</span>
		)}
	</span>
	<div className="flex-row form-group-item">
		<ClaySlider
			className="w-100"
			max={max}
			min={min}
			onValueChange={onValueChange}
			showTooltip={false}
			value={value}
		/>
		<small className="font-weight-normal form-text ml-3">
			{value + '%'}
		</small>
	</div>
</label>
```

Now, the problem I encountered was the indentation when adding a new line inside the `fix: (fixer) => ...`. I tried to add the amount of tabs equalling the column that the node was at, but I don't think it worked, as I'm getting wonky output when running `yarn test`.

## DXP

There are **697** errors found by this rule in `liferay-portal`. 

## Future

- Did we want other newline scenarios to be handled with this rule? Should I investigate further and perhaps split the task into multiple newline rules, if the investigation deems it necessary?
- I need to fix the indentation, not sure how exactly, but I can spend more time figuring it out
- I need to make sure the `fixer` is doing its job and automatically fixing these errors automatically
- Gonna need to write documentation for it

### DIFF
--- a/projects/eslint-config/index.js
+++ b/projects/eslint-config/index.js
@@ -62,6 +62,7 @@ const config = {
 	rules: {
 		'@liferay/liferay/array-is-array': 'error',
 		'@liferay/liferay/destructure-requires': 'error',
+		'@liferay/liferay/empty-line-between-elements': 'error',
 		'@liferay/liferay/group-imports': 'error',
 		'@liferay/liferay/import-extensions': 'error',
 		'@liferay/liferay/imports-first': 'error',
--- a/projects/eslint-config/plugins/liferay/index.js
+++ b/projects/eslint-config/plugins/liferay/index.js
@@ -7,6 +7,7 @@ module.exports = {
 	rules: {
 		'array-is-array': require('./lib/rules/array-is-array'),
 		'destructure-requires': require('./lib/rules/destructure-requires'),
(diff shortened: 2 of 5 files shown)


bryceosterhaus:
> I tried to add the amount of tabs equalling the column that the node was at, but I don't think it worked

I don't think we would want to add tabs on the new line, that should be empty.


> Did we want other newline scenarios to be handled with this rule?

Let's just stick with elements for now. I think other cases might be a bit harder to ensure accuracy

kresimir-coko:
Hey @bryceosterhaus I've pushed a couple small commits. I messed about with the errors `yarn test` is throwing relating to this rule, but I can't figure it out. Not sure where to look, or how to approach the investigation without brute-forcing it.

PR #635, opened 2021-08-31
[POC] Naive Liferay Dev Server

jbalsas:
As the title indicates, this is a quick and naive attempt at building a Liferay Dev Server that I hope can serve as inspiration for the team to pick this up and build a real one!!

https://user-images.githubusercontent.com/905006/131479123-5b8d2108-2bcf-484e-988a-2a1c390eb7ab.mov

## Features
- Command line
- Sets up a server at `http://localhost:3000`
- Proxies all unkonwn requests to a given Liferay Portal instance (`http://localhost:8080` by default)
- Attempts to map requests of the form `/o/moduleName/fileName` or `/o/js/resolved-module/moduleName/fileName` to their output on disk
- Watches for changes and triggers `yarn build` on the specific projects
- Sets up a live-reload mechanism to reload automatically when the build finishes

## To-Do
- Handle SCSS
- Check other possible load patterns and verify modules are mapped correctly to their build output
- Apparently it only works with Chrome Devtools open... figure out why? 😂 
- Improve Build performance

I think last point is important. Had we had this in place, we would've surely noticed that **our build step is not as fast as we'd like it to be**. We put a lot of pressure on gradle and deployment, but the fact is that building some modules takes almost a minute. This should be our cue to start looking at that. 

One thing I had in mind was to trigger some type of incremental builds, by just putting in a virtual folder the files that changed inside a module rather than forcing us to recompile the whole thing. I'll leave it there for another experimentation :) 

Anyways, this is here for your consideration and enjoyment... and hopefully so someone can build this for realz!! 👊 

PS: @nhpatt, I know I said I wouldn't do it, but I did

notsorry


bryceosterhaus:
Sorry, I thought I responded to this but I must have just in my head. Glad you opened this though and starting to move the conversation. I think my main concern is rolling our own implementation of this. I think we should aim to utilize existing technologies that already have done so much work in this space, such as webpack. We may need to write some helpers specifically for deploying resources to the right place, but I think we will have better performance if we use something that already exists. I assume you have already looked into that a bit, but any idea what some of the main hurdles of that would be?

I remember back with Loop and AC, we rolled our own sort of dev server so that we didn't have to deploy every time but we had made a very specific took for our use case. We later switched to using webpack and that seemed to clean it up quite a bit. However, our use case didn't leverage the AMD loader, so we would need to make sure we are building in a similar manner that we are when we actually deploy the module.

jbalsas:
Hey @bryceosterhaus!

> I think we should aim to utilize existing technologies that already have done so much work in this space, such as webpack

I wholeheartedly agree. 

In this case, however, with 571 lines (just 278 of code) this feels pretty lean and close to something that can provide real immediate value to our developers.

> We may need to write some helpers specifically for deploying resources to the right place, but I think we will have better performance if we use something that already exists. I assume you have already looked into that a bit, but any idea what some of the main hurdles of that would be?

Using webpack directly was always my to-go approach. 5+ years later, we still find ourselves without anything at all for what it's really a simple proxy server with some custom resolution logic.

When I thought about this again, I got some bundler v3 vibes... and we all know where that took us :( 

I'd be happy to explore the webpack approach in more detail, but it is my expectation that we'd end up jumping through hoops, trying to bend webpack to work with our setup, or worse, having to rewrite our whole build process so we could benefit from its features. This last part is not bad in itself. We could choose to change our approach so we could reap some of the benefits. My main concern is that we will be only delivering yet another promise to our developers.

I tried to start migrating modules to webpack some years ago with [frontend-js-web](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-web/webpack.config.js) but that eventually didn't pay off and DXP has certainly moved in a different direction.

Long term, I think our bet should be a combination of [module scripts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) and [import maps](https://github.com/WICG/import-maps). I haven't verified this, but I think we could reduce our build pipeline and generate different import maps for prod/dev to reduce the need for deployments. 

Long story short. I think we should look at how we can deliver value to our developers right away. And then, start planning how we want Frontend Developer Experience within Liferay to look like. Doing the latter before is a risk I don't think we should take.

jbalsas:
> [...] but I think we will have better performance if we use something that already exists [...]

Just as a side note, this performs pretty well and does the absolute minimum. The main thing that doesn't perform is our build process itself. 

I played around with incremental builds and an execution of `npm-scripts build` has around 5-6 seconds of overhead. For a configuration with 0 files to process, it still takes `npm-scripts build` 6 seconds to complete. That's right now the main bottleneck, but it is in any case:
- Under our control... we should be able to improve it!
- Already using custom tooling (babel, tsc, webpack...)

I've seen webpack builds taking quite a bit of time too, so I don't think we're that far off.

bryceosterhaus:
> Long story short. I think we should look at how we can deliver value to our developers right away. And then, start planning how we want Frontend Developer Experience within Liferay to look like. Doing the latter before is a risk I don't think we should take.

I like this idea of getting something to devs even if it isn't our perfect ideal. We can always iterate and so long as we expose it via some simple CLI, we can always change the internals of it to speed up.


> Just as a side note, this performs pretty well and does the absolute minimum. The main thing that doesn't perform is our build process itself.

Oh good to know. I haven't gotten a chance to test this yet but you're right that it'll probably help us identify some bottlenecks and speed things up. I wasn't fully aware of the difficulties we had with testing this in webpack previously but it sounds like we have enough experience with it to know that it isn't going to be a quick win.


I'll test this out a bit later this week or next and can provide some more feedback. If we really want to get this out quickly and gather feedback, we could potentially try out some sort of "experimental" track, similarly to how FB has [this organization](https://github.com/facebookexperimental) for projects they experiment with.

matuzalemsteles:
One of the features that significantly improve the Webpack build process is that it knows what should be compiled and uses a lot of caching to skip steps, it also has a lot of parallelisms if you configure it, the same for babel you can configure local cache, you can reduce a lot of time spent on build just setting this even for production builds. But even if you have the best configuration and use the latest version a production build will still take a few seconds, but for the development environment, you can build in a few ms if you combine chunks.

I think if we want to reduce this bottleneck there are many strategies that we could attack, multithreading is an option but it increases the complexity in code, surely incremental builds would be the fastest strategy maybe. But even if we implement everything we will still have bottlenecks ah unless we adopt the [swc](https://github.com/swc-project/swc) which will have much higher gains, there is a community movement that is starting to adopt this in some big projects like Next.js for example. Unfortunately, even though we've improved a lot, Node.js/JS has limitations.

jbalsas:
> But even if you have the best configuration and use the latest version a production build will still take a few seconds [...]

Funny thing is, our build process still takes ~5 seconds even if there aren't any files to process (on a quick test) :)

> But even if we implement everything we will still have bottlenecks ah unless we adopt the swc which will have much higher gains [...]

Moving to [swc](https://github.com/swc-project/swc) or [esbuild](https://github.com/evanw/esbuild) (or other rust-based tooling) should be for sure in our radar for the future.

> I'll test this out a bit later this week or next and can provide some more feedback. If we really want to get this out quickly and gather feedback, we could potentially try out some sort of "experimental" track, similarly to how FB has this organization for projects they experiment with.

We can even make it simpler and just place this directly in `liferay-portal` so it can be invoked from `ant frontend-dev-server` or something like that. That way we can simply have some of our devs test it directly (I've heard @markocikos and @carloslancha complain lately about how hard it is to work with `frontend-taglib-clay` and maybe we can use that feedback to get something that helps them and others.

matuzalemsteles:
> Funny thing is, our build process still takes ~5 seconds even if there aren't any files to process (on a quick test) :)

🙈 Yeah, I think this is very much related to running the separate processes for each phase of the build. We would be able to shorten it by invoking the Node API of each package, instead of running the `babel` and `tsc` CLI and maybe the `bundler` if it has, we could shorten about 1~2 seconds of that time, or more.

> Moving to swc or esbuild (or other rust-based tooling) should be for sure in our radar for the future.

Yes, not that we need to migrate everything now but we can keep an eye on how they are evolving very fast in terms of supporting more features.

bryceosterhaus:
> We can even make it simpler and just place this directly in liferay-portal

Oh I like this. It would allow us to quickly iterate and see how it goes too. We could simply put it in there but not announce it for teams other than our own. We can test it out within our team first and then go from there.

I would opt for making it invoked via node and possibly in our package.json at the modules level. `yarn run liferay-dev`.

jbalsas:
> So testing the dev-server today I ran into a few issues, neither are major

I'm sure there are even more 😂 

> * Unable to sign in on port 3000. Had to go to 8080 first and sign in and then back to 3000

Yeah, this is something I need to investigate... for some reason the sign-in redirect doesn't work well. I thought it might have to do with the backend so I need to debug it a little bit.

> * Liferay lang keys didn't seem to work consistently but this probably makes sense because the file isn't actually deployed

Yeah, I wouldn't expect them to work at all, though it should be pretty straightforward to implement since:
- We have access to the local filesystem, thus to the `Language.properties` files map
- We are proxying the files, so we can do the rewrite on the fly

> * Feels a little slow or it seems like I need to also manually reload sometimes. Again, not a major issue because its only a few seconds.

Yeah, I think there's some kind of interaction with the loader or something... that's why I mentioned as a ToDo that apparently it only works with DevTools open... did you by any chance experience that?

> * The reload alerts can be a little annoying, a few times I get 4-5 stacked.

Hahaha... yeah... might want to get some UX experts to improve that workflow, feel free to tweak it anyway you want!! 🤗 

> One simple idea to improve build time is to add some cli flags for liferay-npm-scripts build that dev-server could use to speed it up. Like --sass-only or --js-only.

Nice one! That definitely helps!

What I was trying (but didn't keep it around for some reason) was:
- Backup config files (`.npmscripts.config.js`...)
- Make special configuration so that `src`, `includes`... only targets the modified file
- Trigger build (should only affect that 1 file)
- Restore original config files

If I remember correctly I kinda made it work for JS, but the CSS build script wouldn't allow properly configuring the globs to point to a single file. We can revisit that approach if we want to push this further. In any case, I think the bottleneck was more on bundler v2 still processing `node_modules` than in the processing of the source files, so maybe it's a moot point.

> Overall though, for only ~400 lines of code, this works surprisingly well and is a big improvement of deploying.

Yeah... ~400 + the lines of the dependencies + the million of lines in Node.js... but that's more or less like everyone else 😂 

> For next steps, I think we just go with it and label is something more experimental and we can gather data on its usefulness. I think we publish this as-is under @liferay/dev-server at 1.0.0-experimental and we can have people use it via npx. Steps to use would be Navigate to DXP and run liferay-dev-server.
>
> Any opposed to just going ahead with this? Eventually I think we end up putting this under @liferay/cli and you can run it with liferay dev

I thought you wanted to move it into DXP and make a more quiet rollout. I'm fine either way, this is for you all to decide what to do with it considering you're going to be the main maintainers. You can even dump it and start from scratch if that works better for you!

--- 

I think @izaera's comments would require further exploration in different threads (especially about which direction to take bundler and portal frontend).

The main point I think is:

> [...]To recap: I think we can proceed with the dev server already (we should have proceeded with it long ago, but never did it :-( )[...]

So I'd suggest to get something out there that covers some part of the frontend development process and then slowly see what else needs to be added avoiding unnecessary features. It's not that often that a frontend is changing configurations or dependencies so it could be okay to go through a full reload for that case, just need to make the expectations clear.

bryceosterhaus:
Added a handful of commits cleaning up the code to be more inline with existing projects.

I also moved it to `js-toolkit` since it is more js than npm related

projects/js-toolkit/packages/dev-server/bin/liferay-dev-server.js
```diff
@@ -5,7 +5,7 @@
  * SPDX-License-Identifier: LGPL-3.0-or-later
  */
 
-import server from '../lib/server';
+const {default: server} = require('../lib/server');
```
jbalsas, line 8: Ha, good catch!

I forgot I was testing this with `NODE_OPTIONS='--experimental-specifier-resolution=node'` just for fun

hallelluja

PR #638, opened 2021-09-01
chore(eslint-config): add no-aui rules to global eslint-config file

kresimir-coko:
Hey @bryceosterhaus, I've figured out what was missing in order to get the `no-aui` rules working inside DXP! Here it is. You can draft a new version if you don't wanna wait for me to do it tomorrow. I assume this would be a minor version?


bryceosterhaus:
@kresimir-coko technically this would be a major version since it introduces breaking changes by turning on this feature by default. Ideally this would have been added in with 27.0.0 (which we thought). However, since we didn't realize it wasn't enabled by default, this is a quick patch version now since 27.0.0 didn't actually break anything yet. Not sure if that made sense...

kresimir-coko:
It did @bryceosterhaus thank you 👍

PR #640, opened 2021-09-01
Notes on liferay-cli & target-platforms

bryceosterhaus:
@izaera, I tried to review the fat jar epic and here is my findings below...

I ran into issues trying to test target platforms and liferay-cli. I suspect that it isn't working properly because Ivan is running a local npm registry and I am not. Below is a list of a few errors I encountered but they may not be issues once these are published, so I am going to wait for Ivan before trying to fix any of these.

- `yarn qa` fails because the `copyAssets` method is copying over the ungenerated scss file to the build folder
- Created a new project with `@liferay/portal-7.4-ga1` and had errors complaining about not having an `/assets` directory.
- Sass was also not building, but I suspect that was because this package wasn't updated to latest
- I tried linking `@liferay/cli` and had issues creating new projects.
```
packages/portal-base/src/deploy.js:22
const outputFile = outputDir.join(outputFilename);
                             ^

TypeError: Cannot read property 'join' of undefined
``` 

Again, I suspect all these issues stem from using `yarn link` and non-released packages so I wanted to wait before trying to fix any of these. Perhaps we should start with an initial alpha release for everything and test without using a local npm registry. I will wait for @izaera on these though.

### DIFF
--- a/target-platforms/qa/test-project/build.expected/css/styles.scss
+++ b/target-platforms/qa/test-project/build.expected/css/styles.scss
@@ -1,0 +1,13 @@
+.pre {
+	font-family: monospace;
+	white-space: pre;
+}
+
+.tag {
+	font-weight: bold;
+	margin-right: 1em;
+}
+
+.value {
+	font-family: monospace;
+}


izaera:
Forgot about this PR... In any case, it doesn't apply any more, since I've merged a lot of changes to be able to release preview versions, so closing it.

Thx.

Issue #627, opened 2021-08-27
Update TS version to 4.4


jbalsas:
### Issue type (mark with `x`)

-   [x] 🎁  Feature request

### Description

[TypeScript 4.4](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/) has been officially launched and includes some useful improvements such as:
- [Improved Control Flow Analysis of Aliased Conditions and Discriminants](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#cfa-aliased-conditions)
- [Symbol and Template String Pattern Index Signatures](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#symbol-template-signatures)

We're currently running TS 4.2, so we're additionally also missing some goodies from [TypeScript 4.3](https://devblogs.microsoft.com/typescript/announcing-typescript-4-3) such as:
- [override and the --noImplicitOverride Flag](https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/#override)
- [Template String Type Improvements](https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/#template-string-improvements)

**Desired behavior:**

Running `liferay-npm-scripts build` should use `typescript@4.4`

**Current behavior:**

Running `liferay-npm-scripts build` uses `typescript@4.2.3`

**Repro instructions (if applicable):**

**Other information (environment, versions etc):**

There's a bit of an implicit dependency between the version of TS we run and the version of [typescript-eslint/typescript-eslint](https://github.com/typescript-eslint/typescript-eslint). If you try to run `liferay-npm-scripts format` with a version of `TS>4.3`, you'll get the following warning:

```
=============
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
You may find that it works just fine, or you may not.
SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.3.0
YOUR TYPESCRIPT VERSION: 4.4.2
Please only submit bug reports when using the officially supported version.
=============
```

I assume this will depend on [TypeScript 4.4 Syntax Support](https://github.com/typescript-eslint/typescript-eslint/issues/3606) so we might want to watch for it and wait to update both at the same time to avoid the noise in our output.

PR #642, opened 2021-09-01
chore: update typescript to v4.4.0

bryceosterhaus:
fixes https://github.com/liferay/liferay-frontend-projects/issues/627

Tried this locally and didn't see errors, should be good to go?


projects/npm-tools/packages/npm-scripts/package.json
```diff
@@ -76,7 +76,7 @@
 		"stylelint": "^13.2.0",
 		"terser": "5.3.8",
 		"ts-loader": "^8.0.4",
-		"typescript": "^4.2.3",
+		"typescript": "^4.4.0",
```
jbalsas, line 79: Should we prefer a fixed version `4.4.2` here or at least `~4.4.2`? After all, `^4.2.3` also works for `4.4.2` ;) 

Based on the experience here, I get the feeling we want to be very explicit about which specific version we're using (and shipping). That way, devs in Liferay DXP might have an easier time figuring out which TS options are available to them.

This makes me think that we might also want to start being more vocal about TS support so we get more people on board, but maybe that can wait until we land some bits of https://github.com/liferay-frontend/liferay-portal/pull/1379

jbalsas (approved):
LGTM based on our current setup, but based on the experience on trying to figure out which TS features are available (can't rely just on `package.json` broad version, really), I'd say it would be better if we restricted the version range a bit more.

projects/npm-tools/packages/npm-scripts/package.json
bryceosterhaus: Totally agree. Just pushed an update

Issue #639, opened 2021-09-01
Investigate workspace scripts for version/publish to avoid pushing code without publishing


bryceosterhaus:
We have had pushes happen to the repo and then break on publishing which then can lead to multiple tags and commits for versions on the repo. We need to look into this to make the scripts a little more safe.

PR #643, opened 2021-09-01
fix(js-publish): add message when publishing fails

bryceosterhaus:
fixes https://github.com/liferay/liferay-frontend-projects/issues/639

This is to help developers know what to do next when publishing fails. This will help us avoid people running `yarn version` multiple times and accidentally pushing non-published versions to upstream.

### DIFF
--- a/projects/npm-tools/packages/js-publish/src/index.js
+++ b/projects/npm-tools/packages/js-publish/src/index.js
@@ -223,7 +223,17 @@ async function runYarnPublish(pkg) {
 		args.push('--tag', 'prerelease');
 	}
 
-	run('yarn', 'publish', ...args);
+	try {
+		run('yarn', 'publish', ...args);
+	}
+	catch (error) {
+		printBanner(
+			'Failed to publish package! ❌',
+			'',
+			`You've already pushed to remote.`,
+			`Try publishing with 'yarn publish ${args.join(' ')}'`
+		);
+	}
 }
 
 module.exports = {


projects/npm-tools/packages/js-publish/src/index.js
```diff
@@ -223,7 +223,17 @@ async function runYarnPublish(pkg) {
...
-	run('yarn', 'publish', ...args);
+	try {
+		run('yarn', 'publish', ...args);
+	}
+	catch (error) {
+		printBanner(
+			'Failed to publish package! ❌',
+			'',
+			`You've already pushed to remote.`,
+			`Try publishing with 'yarn publish ${args.join(' ')}'`
```
jbalsas, line 234: Do we get any relevant information as to why publishing failed? I'm assuming in some cases the developer might need to do something before being able to successfully re-run `yarn publish`?

bryceosterhaus: We delegate to `yarn publish` and that should emit whatever error occurred. Most likely it is due to invalid credentials or an expired OTP token. Could even be that npm is unreachable.

Rather than trying to explain the specific reason of failure (because we won't know), I thought it was best to describe what has happened (git push) and then next steps for finishing the task.

jbalsas: > We delegate to yarn publish and that should emit whatever error occurred

Fair enough, then! I do like the fact that we provide clear instructions to retry 👍

jbalsas (approved):
goforit

projects/npm-tools/packages/js-publish/src/index.js
diegonvs: > expired OTP token

I never get the current OTP token at first try when publishing things 😂 

Adding this exception would be good instead of [a long list of instructions](https://github.com/liferay/liferay-frontend-projects/blob/6f839c15a2673c364e7afbbe9e21efae549e4993/projects/npm-tools/packages/js-publish/src/index.js#L69) :)

PR #650, opened 2021-09-08
feat(eslint-config): add rule to ensure JSX expressions guard against rendering 0 on accident

### LINKED ISSUE #648: Add eslint rule to avoid rendering literal `0` when we conditionally render jsx based on length
```jsx
{someItems.length && <div />} // this will literally render 0 if no items exist.

// Fix

{!!someItems.length && <div />}
```

bryceosterhaus:
fixes #648

This doesn't address if a user creates a variable first, but we can add that later if needed


jbalsas (approved):
thumbsup

I wonder how many of this we'll see in `liferay-portal` 😅

kresimir-coko (approved):
LNHTM 👍

PR #662, opened 2021-09-30
feat(eslint-plugin): create eslint-plugin to replace eslint-config

bryceosterhaus:
https://github.com/liferay/liferay-frontend-projects/issues/9

[POC of this in DXP](https://github.com/liferay-frontend/liferay-portal/pull/1522/files)

This should completely replace eslint-config, and we can delete and deprecate that package once we apply this to DXP. We also need to update all of our docs here and other projects using eslint-config. See above issue for checklist.

Primary benefit of this change is that eslint should now work in vscode while working within dxp.


bryceosterhaus:
Gonna just move this forward and we can make changes if need be in future PRs. 90% of this PR is copied straight from the first commit which is just copying the config directory.

liferay-headless:
Hehehe :P LGTM... oh wait.

PR #663, opened 2021-09-30
feat(npm-scripts): use eslint-plugin instead of eslint-config

bryceosterhaus:
Removes eslint-config and updates references to package


bryceosterhaus:
fyi bulk of changes is the removal of the eslint-config directory. I don't imagine us patching anything there since we would opt for just using the plugin instead. If there ever is an emergency circumstance, we can always roll back to this commit and publish from there.

bryceosterhaus:
Let's just keep this moving, and as always, if anything comes up I can update in a future PR

PR #670, opened 2021-10-05
fix(js-toolkit-core): add missing dependency

izaera:
We were missing `ejs` dependency.


izaera:
Merging to release a new version

Issue #668, opened 2021-10-04
Proposal for stabilization plan


izaera:
### Issue type (mark with `x`)

-   [X ] :thinking: Question
-   [ ] :bug: Bug report
-   [ ] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [X] :two: v2.x
-   [X] :three: v3.x

### Motivation

We currently have two versions of the bundler in this repo: [v2](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit) and [v3](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/js-toolkit). The former being the official one, while v3 is experimental and has only been used [in one project in liferay-portal](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-recharts/package.json#L6).

Bundler v3 was almost built to its [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product) phase but its development was then stopped because of the release of the webpack federation feature. We tried to develop a PoC based on webpack federation to get rid of bundler v2 and v3 and almost succeeded, but then needed to revert it because it didn't fulfill all our needs (specifically we couldn't guarantee which OSGi bundle provided which npm package).

Now, we have added [target platforms](https://github.com/liferay/liferay-frontend-projects/tree/master/target-platforms) and [@liferay/cli](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/js-toolkit/packages/liferay-cli) to the mix.

Target platforms are a new build system that couples projects with a certain release of DXP or Portal so that they can leverage the full potential of our platform (i.e: instead of making almost agnostic projects, like we did before with the [yeoman generator](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit/packages/generator-liferay-js), we make very opinionated projects that target a specific Portal/DXP version).

The advantage of using target platforms is that development and configuration is much simpler than before, because the target platform can take care of almost everything for the developer, given that it knows where the project will be deployed. The con is that you are coupling your project to Liferay and, that you won't be free to tailor your build to the last detail.

[...]

PR #671, opened 2021-10-05
refactor: stabilize toolkit versions

izaera:
Make some tweak in JS Toolkit in order to achieve stabilization (#668)


.prettierignore
```diff
@@ -19,6 +19,7 @@
 
 /projects/js-toolkit/**/generators
 /projects/js-toolkit/**/lib
+/projects/js-toolkit/qa/**
```
bryceosterhaus, line 22: Why should we ignore prettier in these files? Does it cause issue?

bryceosterhaus (commented):
> Screen Shot 2021-10-05 at 9 45 48 AM

.prettierignore
izaera: Because it is usually a project generated for customers (with our generator/@liferay-cli), following other coding standards (for example it may suit React or Angular standards), so it doesn't make too much sense to lint them according to our rules.

It's not that we cannot add them, but it is not worth the effort and may give us problems in the future. 

In fact, I've been talking with @magjed4289 about automating this kind of tests in the future, so we may need to move them out of here. It really depends on how we design the QA architecture for the future.

bryceosterhaus:
positive/neutral gif === approval



Rejected PR Gif

Issue #672, opened 2021-10-05
Update themes deps versions to get the latest available 


georgel-pop-lr:
### Issue type (mark with `x`)

-   [x ] :thinking: Question
-   [ x] :bug: Bug report
-   [ ] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [x ] :eight: v8.x
-   [ ] :nine: v9.x
-   [ ] :keycap_ten: v10.x or higher

### Description
With each release of a fix-pack we release a new version of "liferay-frontend-theme-styled" and "liferay-frontend-theme-unstyled" but if we don't update this versions manually in the themes, just regenerating it doesn't get the latest version available. Each time we need to update these dependencies versions on the repo. 

**Desired behavior:**
Get the latest available version for  "liferay-frontend-theme-styled" and "liferay-frontend-theme-unstyled" after each fix-pack release.

**Current behavior:**
You don't get the latest available version for  "liferay-frontend-theme-styled" and "liferay-frontend-theme-unstyled" after each fix-pack release.

**Repro instructions (if applicable):**
The client wants to see the changes of https://issues.liferay.com/browse/LPS-135023 inside their theme, but by only regenerating it, that's not possible. They have to manually change the latest version in their theme node_modules/liferay-theme-deps-7.1/package.json do a npm install and gulp deploy the theme. 

**Other information (environment, versions etc):**

PR #673, opened 2021-10-05
fix(#672): update 7.1 themes deps  versions to get the latest available

georgel-pop-lr:


bryceosterhaus:
@georgel-pop-lr this seems like a reasonable fix to me as we are already doing this with `liferay-theme-deps-7.0` and its dependencies.

Can you rebase this PR and update? I noticed it also has a commit from Ivan that is already in upstream. I'm not totally sure why it is including that here...

georgel-pop-lr:
> @georgel-pop-lr this seems like a reasonable fix to me as we are already doing this with `liferay-theme-deps-7.0` and its dependencies.
> 
> Can you rebase this PR and update? I noticed it also has a commit from Ivan that is already in upstream. I'm not totally sure why it is including that here...

Okay, great, I rebase and push.

But I guess in order to always get the latest version, we will need to always update the yarn.lock file? 

Regards.

bryceosterhaus:
@georgel-pop-lr yeah the customer _may_ need to update their yarn.lock, but it may also automatically do that for them when it sees the `~`.

PR #686, opened 2021-10-11
feat(eslint-plugin): update and add eslint rules and config

### LINKED ISSUE #644: Disable use of `localhost` url references
We shouldn't allow users to reference localhost in code committed to DXP. We can encourage the use of referencing the `window.location` object instead.

### LINKED ISSUE #647: Variable delcarations for a `ref` in react should always have `*Ref` in the name
All ref variables should be suffixed with `Ref`

### LINKED ISSUE #655: Require that React.useState returned arguments has consistent naming pattern
Enforce use of `const [*, set*] =` when using useState hook.

Valid
```
const [value, setValue] = useState();
```

Invalid
```
const [value, updateValue] = useState();
```

### LINKED ISSUE #669: Update empty-line-between-elements to handle more use cases
Should Error
```
<div>
	{someCondition && <div />}
	<div />
</div>
```

Fix
```
<div>
	{someCondition && <div />}

	<div />
</div>
```

bryceosterhaus:
fixes #644
fixes #655
fixes #669
fixes #647

Let me know if any of these don't make sense

### DIFF
--- a/projects/eslint-plugin/configs/general.js
+++ b/projects/eslint-plugin/configs/general.js
@@ -81,8 +81,10 @@ const config = {
 		'@liferay/no-length-jsx-expression': 'error',
 		'@liferay/no-require-and-call': 'error',
 		'@liferay/padded-test-blocks': 'error',
+		'@liferay/ref-name-suffix': 'error',
 		'@liferay/sort-import-destructures': 'error',
 		'@liferay/sort-imports': 'error',
+		'@liferay/use-state-naming-pattern': 'error',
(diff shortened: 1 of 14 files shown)


bryceosterhaus:
Results in DXP

Screen Shot 2021-10-11 at 4 49 23 PM

bryceosterhaus:
Just realized I need to also add some logic for auto-fixing references to variables I am changing. For example, I need to not only autofix the declaration of node -> nodeRef, but also where it is used.

Issue #624, opened 2021-08-25
Adjust no-global-fetch rule to conform with TS migration patterns


kresimir-coko:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [x] :gift: Feature request
-   [ ] :woman_shrugging: Other

Prompted by @jbalsas in https://github.com/liferay-frontend/liferay-portal/pull/1379#discussion_r695720969

PR #689, opened 2021-10-12
chore(eslint-plugin): lint to accept "fetch", not "fetch.es"

kresimir-coko:
Fixes https://github.com/liferay/liferay-frontend-projects/issues/624

Lemme know if I did it properly, I couldn't come up with a different solution.

/cc @jbalsas

### DIFF
--- a/projects/eslint-plugin/rules/portal/lib/rules/no-global-fetch.js
+++ b/projects/eslint-plugin/rules/portal/lib/rules/no-global-fetch.js
@@ -16,10 +16,12 @@ module.exports = {
 
 		const isValidDefaultImport = (node) => {
 			return (
-				node.source.value.endsWith('/fetch') &&
-				node.specifiers.find(
-					(specifier) => specifier.type === 'ImportDefaultSpecifier'
-				)
+				node.source.value.endsWith('/fetch') ||
+				(node.source.value.endsWith('/fetch.es') &&
+					node.specifiers.find(
+						(specifier) =>
+							specifier.type === 'ImportDefaultSpecifier'
+					))
 			);
 		};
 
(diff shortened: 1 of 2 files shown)


projects/eslint-plugin/rules/portal/lib/rules/no-global-fetch.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
 
 		const isValidDefaultImport = (node) => {
 			return (
-				node.source.value.endsWith('/fetch.es') &&
+				node.source.value.endsWith('/fetch') &&
```
bryceosterhaus, line 19: Instead of choosing one or the other, lets allow for both since it is likely we will have a point in DXP where both exist or it may get back ported to older versions that use `.es`
```suggestion
				(node.source.value.endsWith('/fetch') || node.source.value.endsWith('/fetch.es')) &&
```

projects/eslint-plugin/rules/portal/lib/rules/no-global-fetch.js
```diff
@@ -16,10 +16,12 @@ module.exports = {
...
-				node.source.value.endsWith('/fetch.es') &&
-				node.specifiers.find(
-					(specifier) => specifier.type === 'ImportDefaultSpecifier'
-				)
+				node.source.value.endsWith('/fetch') ||
+				(node.source.value.endsWith('/fetch.es') &&
+					node.specifiers.find(
+						(specifier) =>
+							specifier.type === 'ImportDefaultSpecifier'
+					))
```
bryceosterhaus, line 24: I think the parenthesis needs to move here so that the OR is for the endsWith and the AND is for the find

```suggestion
				(node.source.value.endsWith('/fetch') ||
					node.source.value.endsWith('/fetch.es')) &&
				node.specifiers.find(
					(specifier) =>
						specifier.type === 'ImportDefaultSpecifier'
				)
```

bryceosterhaus (changes requested):

projects/eslint-plugin/rules/portal/lib/rules/no-global-fetch.js
kresimir-coko: Good catch, that was prettier that autofixed that in that specific manner

PR #690, opened 2021-10-12
feat(eslint-plugin): enable sorting rule for TS interfaces

### LINKED ISSUE #579: Investigate using a plugin for sorting interface keys
liferay-portal is pretty picky on sorting keys in an object and so we might want to investigate keeping consistency here by adding an eslint plugin for sorting objects and type definitions. Below is the type of code we want to sort.

```ts
interface PresetInfo {
    name: string;
    isAutopreset: boolean; // this should be before "name"
}
```

bryceosterhaus:
fixes #579

cc @izaera, this seem good with you? I figure since we are sorting everywhere else in Liferay, this makes sense to add.


izaera:
Funny, because I thought this rule was already in place :sweat_smile: I try to sort props alphabetically in every place.

PR #691, opened 2021-10-13
feat(eslint-plugin): add one-var rule

kresimir-coko:
With the gentle nudge by @nhpatt I ended up testing and using the [one-var](https://eslint.org/docs/rules/one-var) rule with the `['error', 'never']` settings. I have tested it inside `liferay-portal` and it produces 72 autofixable errors.

~~I found this [one-var-declaration-per-line](https://eslint.org/docs/rules/one-var-declaration-per-line) ESLint rule. When I plugged it in `liferay-portal` it produced 63 errors, all of which are auto-fixable. The autofix is weird as it just pushes the vars to a new line. The autofix issue can be seen in the 2nd commit.~~

### DIFF
--- a/maintenance/projects/js-themes-toolkit-v9-x/packages/liferay-theme-tasks/lib/r2/liferay-r2/css-parse.js
+++ b/maintenance/projects/js-themes-toolkit-v9-x/packages/liferay-theme-tasks/lib/r2/liferay-r2/css-parse.js
@@ -277,7 +277,8 @@ module.exports = function (css, options) {
 
 		// declarations
 
-		var decl, rule;
+		var decl;
+		var rule;
 		while ((decl = declaration()) || (rule = expandedatrule())) {
 			if (decl) {
 				decls.push(decl);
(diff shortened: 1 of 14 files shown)


nhpatt:
Did is not what we want, AFAIK. I would say that this is closer: https://eslint.org/docs/rules/one-var

kresimir-coko:
> Did is not what we want, AFAIK. I would say that this is closer: https://eslint.org/docs/rules/one-var

@nhpatt that crossed my mind, I'll check it out as their docs are unclear, but I thought that we want to declare variables when they're about to be used, not at the start of the function, and as I'm reading this rule, it lints against that. For example this is incorrect code: 
```js
function foo(){
    const bar = false;
    const baz = true;
    let qux;
    let norf;
}
```
But, this is correct code:
```js
function foo(){
    const bar = true,
        baz = false;
    let qux,
        norf;
}
```
I've never seen us comma-separate variable declarations.

nhpatt:
> I thought that we want to declare variables when they're about to be used, not at the start of the function

Yep, that's what we want

> I've never seen us comma-separate variable declarations.

I saw that there are some flags that disallow that and force you to use var/let several times

kresimir-coko:
You were right @nhpatt, sorry for doubting you Sensei.

kresimir-coko:
I've force pushed the new rule @nhpatt, thanks for the feedback

bryceosterhaus:
LGTM!

PR #692, opened 2021-10-14
feat(eslint-plugin): add new custom rule to lint against !Object.keys() in conditionals

kresimir-coko:
I've pushed a complete version of this rule, so it should be ready for review. Let me know if I reduced the duplication as much as possible.

### DIFF
--- a/projects/eslint-plugin/rules/general/docs/rules/no-conditional-object-keys.md
+++ b/projects/eslint-plugin/rules/general/docs/rules/no-conditional-object-keys.md
@@ -1,0 +1,65 @@
+# Don't use negated `!Object.keys()` or double negated `!!Object.keys()` expressions as they always evaluate to `false` and `true`, respectively
+
+This rule enforces that engineers don't use a truthy or falsy value that `Object.keys()` can return. Instead, use an expression that is more precise.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```js
+if (!Object.keys({foo: 'bar'})) {
+	// do your magic
+}
+```
+
+or
+
+```js
+!!Object.keys({foo: 'bar'}) && 'test';
+```
+
+or
+
+```js
+const negatedObjectKeys = !Object.keys({foo: 'bar'});
+```
+
+or
+
+```js
+!Object.keys({foo: 'bar'}) && true;
+```
+
+Examples of **correct** code for this rule:
+
+```js
+if (Object.keys({foo: 'bar'})) {
+	// do your magic
+}
+```
+
+or
+
+```js
+if (Object.keys({foo: 'bar'}).length) {
+	// do your magic
+}
+```
+
+or
+
+```js
+Object.keys({foo: 'bar'}) && 'test';
+```
+
+or
+
+```js
+if (Object.keys({foo: 'bar'}).find((i) => true)) {
+	// do your magic
+}
+```
+
+## Further Reading
+
+-   https://github.com/liferay/liferay-frontend-projects/issues/10
--- a/projects/eslint-plugin/rules/general/docs/rules/no-conditional-object-keys.md
+++ b/projects/eslint-plugin/rules/general/docs/rules/no-conditional-object-keys.md
@@ -7,54 +7,30 @@ This rule enforces that engineers don't use a truthy or falsy value that `Object
 Examples of **incorrect** code for this rule:
 
 ```js
+if (Object.keys({foo: 'bar'})) {
+	// do your magic
+}
+
 if (!Object.keys({foo: 'bar'})) {
 	// do your magic
 }
-```
 
-or
-
-```js
 !!Object.keys({foo: 'bar'}) && 'test';
-```
 
-or
-
-```js
 const negatedObjectKeys = !Object.keys({foo: 'bar'});
-```
 
-or
-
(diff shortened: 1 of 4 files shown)


projects/eslint-plugin/rules/general/lib/rules/no-conditional-object-keys.js
```diff
@@ -0,0 +1,72 @@
...
+			if (objectName === OBJECT && propertyName === KEYS) {
+				context.report({
+					message,
+					node,
+				});
+			}
+		};
+
+		return {
+			IfStatement(node) {
```
bryceosterhaus, line 23: Rather than checking just against `if ()` blocks, we also need to check against other types of conditionals.

For example:
```js
!Object.keys({}) ? false : true

!Object.keys({}) && 'test'
```


However, this also makes me think, do we ever have a case where we would want to allow `!Object.keys(someObj)` on it's own? I don't think so since that will always evaluate to false. So I think a more robust solution would be to check against all cases of `!Object.keys(someObj)`. We wouldn't want to check against cases that  call `.length` or some other key/method on `Object.keys(someObj)`.

projects/eslint-plugin/rules/general/docs/rules/no-conditional-object-keys.md
```diff
@@ -0,0 +1,41 @@
...
+	// do your magic
+}
+```
+
+or
+
+```js
+if (Object.keys({foo: 'bar'}).length) {
+	// do your magic
+}
```
bryceosterhaus, line 20: I think this should be a correct usage

projects/eslint-plugin/rules/general/tests/lib/rules/no-conditional-object-keys.js
```diff
@@ -0,0 +1,111 @@
...
+const parserOptions = {
+	parserOptions: {
+		ecmaVersion: 6,
+		sourceType: 'module',
+	},
+};
+
+const ruleTester = new MultiTester(parserOptions);
+
+const message = "don't use truthy values in conditionals";
```
bryceosterhaus, line 18:
```suggestion
const message = "!Object.keys(anyArg) always evaluates to false";
```

projects/eslint-plugin/rules/general/lib/rules/no-conditional-object-keys.js
```diff
@@ -0,0 +1,57 @@
...
+module.exports = {
+	create(context) {
+		const throwError = (node, objectName, propertyName) => {
+			if (objectName === OBJECT && propertyName === KEYS) {
+				context.report({
+					message,
+					node,
+				});
+			}
+		};
```
bryceosterhaus, line 21: `throwError` is a little strange for the name since it only reports an error if a condition is met. I would call this something like `checkExpression`.

bryceosterhaus (changes requested):
I just tested this against `!!Object.keys()` and get an error running the lint.

You may want to check for `CallExpression` instead of `UnaryExpression` too, I think it might make the logic a little more straight forward for checking both `!` and `!!`

- check for all CallExpression and for nodes that equal Object.keys()
- Check that CallExpression and see if its parent is a `UnaryExpression` (! case)
- Check to see if that `UnaryExpression` has a parent of `UnaryExpression` (!! case)
- Finally, report if it is ! or !!

kresimir-coko:
> @bryceosterhaus I messed around with it a bit to handle a few more cases.

I see what you did, I think I might have taken some of your previous recommendations too much to heart as I assumed `Object.keys({})` was valid and that we don't want to lint against `.length` and other methods.

It looks more comprehensive than my solution, I'll apply the changes and push 👍

kresimir-coko:
@bryceosterhaus I've pushed commits that implement your suggestions. Not sure why new CI/github actions didnt run, but yea, should be ready to get looked into, again 👀

projects/eslint-plugin/rules/general/docs/rules/no-conditional-object-keys.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+Examples of **correct** code for this rule:
+
+```js
+if (Object.keys({foo: 'bar'}).length) {
+	// do your magic
+}
+
+Object.keys({foo: 'bar'}) && 'test';
```
bryceosterhaus, line 32: This would be an incorrect code since it doesn't check for length

projects/eslint-plugin/rules/general/index.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
 	'no-abbreviations': require('./lib/rules/no-abbreviations'),
 	'no-absolute-import': require('./lib/rules/no-absolute-import'),
 	'no-arrow': require('./lib/rules/no-arrow'),
+	'no-conditional-object-keys': require('./lib/rules/no-conditional-object-keys'),
```
bryceosterhaus, line 17: Note that we also need to "turn on" this rule in our general config as well

bryceosterhaus (commented):
I noticed a few more things, but I can just pull down this PR and update it and re-send

bryceosterhaus:
follow up https://github.com/liferay/liferay-frontend-projects/pull/721

PR #693, opened 2021-10-14
feat: migrate adaptations for react, angular and vue to the new js-toolkit

izaera:
Not much to say, as this is simply moving code from one place to the other. I've polished it a bit while moving, but it's mostly that.


izaera:
Going to merge this to be able to continue

ninja

PR #699, opened 2021-10-18
chore: add more lint rules

### LINKED ISSUE #3: Change quote-props rule to "consistent-as-needed"
I just got bitten by this while preparing this PR: https://github.com/liferay/liferay-js-themes-toolkit/pull/247

Basically, given an object containing some version numbers properties:

```
{
  '6.2': true,
  '7.0': true,
  '7.1': true,
}
```

we currently have ESLint configured via "quote-props" to require quotes around property names "as-needed"; producing this:

```
{
  6.2: true,
  '7.0': true,
  7.1: true,
}
```

This looks pretty terrible. I noticed that Prettier didn't have an opinion on this, so I changed them all to skip the quotes:

```
{
  6.2: true,
  7.0: true,
  7.1: true,
}
```

of course, that is wrong too, because `7.0` actually gets stringified to `'7'`, meaning the object is equivalent to:

```
{
  '6.2': true,
  '7': true,
  '7.1': true,
}
```

This breaks checks of the form `if (object['7.0'] === true)`. So I added a lint suppression and put strings back on all the keys, but we should just configure ESLint to use "consistent-as-needed" instead. Via [the docs](https://eslint.org/docs/rules/quote-props):

> requires quotes around all object literal property names if any name strictly requires quotes, otherwise disallows quotes around object property names

### LINKED ISSUE #6: Lint for expect() calls without matchers
I see quite a few instances of test code that looks like this:

```javascript
expect(getByText('report'));
expect(getByRole('button'));
```

These don't actually *assert* anything (other than implicitly, that `getByText()`/`getByRole()`) didn't throw before `expect()` was even called. This would be equivalent:

```javascript
getByText('report');
getByRole('button');
```

Let's add a lint against doing nothing with the return value of `expect()` in tests: it should always be chained with a matcher. In the examples above, that could be:

```javascript
import '@testing-library/jest-dom/extend-expect';

expect(getByText('report')).toBeVisible();
expect(getByRole('button')).toBeVisible();
```

or similar.

### LINKED ISSUE #12: Lint to prevent unhandled promised rejections
Every promise should have a `catch()`, otherwise we risk creating log spew in the browser console that makes us look incompetent for a lack of error handling (and that would be a fair appraisal!).

Example of **incorrect** code:

```js
if (online) {
    client.connect(url).then(result => {
        dispatch({action: 'blah', result});
    });
}
```

(An unrelated nit with that code: if it is in a React component, it is probably missing an `isMounted()` check to guard against setting state of similar on an unmounted component.)

Examples of **correct** code:

```js
client.connect(url)
  .then(result => {
    dispatch({action: 'blah', result});
  })
  .catch(() => {
    // - gracefully recover; or:
    // - retry; or:
    // - log (maybe only when `process.env.NODE_ENV === 'development'`); or, at worst:
    // - swallow
  })
```

Note that it is OK not to `catch()` if you are returning your promise to somebody else: we assume that they do the right thing (even though we might not make an attempt to statically verify that):

```js
return client.connect(url);
```

For similar reasons (ie. that we assume the caller will deal with the problem) it's also ok not to `try`/`catch` when using `async`/`await` (currently we can only do that in the test environment, but [expect to be able to do it soon throughout portal](https://github.com/liferay/liferay-npm-tools/issues/338)):

```js
async function get(item) {
  // We expect you to do local `try`/`catch` if you want to handle the error locally,
  // but if you don't, we assume caller does it.
  const raw = await client.get(item);

  return JSON.parse(raw);
}
```

Basically the rationale of all this is to catch obvious mistakes, but other than that give people the freedom to structure their code as they wish without insisting that they litter it at every level with explicit error handling.

### LINKED ISSUE #311: Standardize data attributes retrieval
Should we standardize how are we retrieving the data attributes? Should we use `getAttribute` for everything, or use `dataset` in the case we are retrieving a data attribute?

bryceosterhaus:
fixes #12
fixes #6
fixes #311
fixes #3

Slowly working through backlog of eslint rules we wanted to add.


bryceosterhaus:
DXP changes needed after this PR. (there were over 1k auto-fixes 😨 )

Screen Shot 2021-10-18 at 2 48 37 PM

Most of these are related to `Expected catch() or return  promise/catch-or-return`

projects/eslint-plugin/rules/general/lib/rules/no-get-data-attribute.js
```diff
@@ -0,0 +1,52 @@
...
+					node.callee.type === 'MemberExpression' &&
+					node.callee.property &&
+					node.callee.property.name === 'getAttribute' &&
+					node.arguments &&
+					node.arguments[0] &&
+					node.arguments[0].value &&
+					node.arguments[0].value.startsWith('data-')
+				) {
+					const argumentValue = node.arguments[0].value;
+					const argumentName = argumentValue.replace('data-', '');
```
kresimir-coko, line 21: Shouldn't there be an empty line inbetween these?

projects/eslint-plugin/rules/general/lib/rules/expect-assert.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * SPDX-FileCopyrightText: © 2021 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+const MESSAGE = 'Every `expect()` should assert something.';
```
kresimir-coko, line 6: Interesting that you went with capitalization here, which prevents you to imply `message` inside `context.report`

kresimir-coko (approved):
Looks great, just a couple nitpicks.

PR #701, opened 2021-10-19
refactor: move common platform code projects to JS Toolkit 

izaera:
This moves base target platforms projects to js-toolkit workspace and makes target platforms as lean as possible.

This has the following benefits:

1. The js-toolkit workspace has TypeScript support
2. The js-toolkit workspace is integrated in the general yarn workspace, allowing to access related projects through yarn link, for example
3. The target-platforms projects is only used as a lean repository of bundler configuration files/marker projects, not as a real development project with lots of source code. This is good because target-platforms is detached from the general yarn workspace (to avoid polluting dependency tree) and has a very limited build support.

I've tested that nothing broke with our fantabulous JS Toolkit QA scripts :tada:

### DIFF
--- a/projects/js-toolkit/scripts/util/spawn.js
+++ b/projects/js-toolkit/scripts/util/spawn.js
@@ -7,9 +7,14 @@ const childProcess = require('child_process');
 
 module.exports = function spawn(cmd, args = [], options = {}) {
 	const result = childProcess.spawnSync(cmd, args, {
+		shell: true,
 		stdio: 'inherit',
 		...options,
 	});
+
+	if (result.error) {
+		throw result.error;
+	}
 
 	if (result.status !== 0) {
 		throw new Error(`Execution failure: ${cmd} ${args.join(' ')}`);


izaera:
Note that, once we merge this, we are ready to do the product review.

However, I'll send a PR with the scripts for local testing with Verdaccio first, so that I can then explain how can this be tested.

projects/js-toolkit/packages/portal-adapt-base/package.json
```diff
@@ -1,23 +1,32 @@
...
 	"dependencies": {
 		"@liferay/portal-base": "^1.0.0",
 		"cross-spawn": "^7.0.0",
 		"escape-string-regexp": "^2.0.0",
 		"globby": "^10.0.1"
 	},
 	"description": "Base library for adaptation autopresets",
-	"main": "./src/index.js",
+	"license": "LGPL-3.0",
+	"main": "lib/index.js",
```
bryceosterhaus, line 11: Just to verify, we will need to re-build this during development anytime we make changes?

projects/js-toolkit/tsconfig.json
```diff
@@ -2,13 +2,13 @@
 	"compileOnSave": true,
 	"compilerOptions": {
 		"esModuleInterop": true,
-		"lib": ["ES2015", "DOM"],
+		"lib": ["ES2015"],
```
bryceosterhaus, line 5: This might be for a different thread, but any thoughts on upgrading to es2021?

izaera: Apparently ES2021 is only [fully supported in node v16](https://stackoverflow.com/questions/67371787/what-typescript-configuration-produces-output-closest-to-node-js-16-capabilities), so we may want to defer this, as we usually say that we use node LTS version as reference (which [is now 14](https://nodejs.org/en/)) in all our tooling...

PR #702, opened 2021-10-19
style(senna): fix linting errors

izaera:
Just that!


maintenance/projects/senna/src/screen/RequestScreen.js
```diff
@@ -274,8 +274,9 @@ class RequestScreen extends Screen {
 						reason.requestError = true;
 						reason.requestPrematureTermination = true;
 						break;
+					default:
+						throw reason;
 				}
-				throw reason;
```
bryceosterhaus, line 278: This seems like slightly different logic. Maybe I am not reading this right but I am seeing that the `throw` 

Previous Code:
- Run switch statement and set timeout/requestError
- Then, `throw reason`

New Code:
- Run switch statement and set timeout/requestError
- OR `throw reason`.

With the new code, we won't throw unless one of the `case`'s are not met

izaera:
> I'm impressed you went through the trouble of updating all the should in tests.

:joy: I made a vim macro for that: remove `should`, move one word right, add an `s`. That made it for 90% of the tests since English behaves as a good language in this case. 

Otherwise I would have changed eslint config, most probably :roll_eyes: 

One funny thing I noticed is that someone, at some moment in time, had to add some `should`s, because of how they were written in some of the tests (incorrectly) :tada:

maintenance/projects/senna/src/screen/RequestScreen.js
izaera: :open_mouth: Good catch

izaera:
Ready to go :rocket: 

I had to forcepush to avoid a "fabricated" `fix` semantic commit that would have appeared in the changelog, maybe.

izaera:
I'll merge this since I've fixed [this issue](https://github.com/liferay/liferay-frontend-projects/pull/702#discussion_r732104667) and the rest was approved.

PR #707, opened 2021-10-20
feat: add eqeqeq eslint rule

### LINKED ISSUE #14: Enforce use of "strict equality"

bryceosterhaus:
fixes #14

@izaera, could you check my changes to make sure I didn't break anything? I updated all cases of `someVal == null` and `someVal != null` to also include a check for undefined since we are enforcing `===`

### DIFF
--- a/maintenance/projects/js-themes-toolkit-v8-x/packages/liferay-theme-tasks/test/util.js
+++ b/maintenance/projects/js-themes-toolkit-v8-x/packages/liferay-theme-tasks/test/util.js
@@ -321,7 +321,7 @@ function cleanTempTheme(themeName, version, component, initCwd) {
 
 	cleanDirectory(tempPath);
 
-	if (initCwd !== null) {
+	if (initCwd !== null && initCwd !== undefined) {
 		process.chdir(initCwd);
 	}
 }
--- a/maintenance/projects/js-themes-toolkit-v8-x/packages/liferay-theme-tasks/test/util.js
+++ b/maintenance/projects/js-themes-toolkit-v8-x/packages/liferay-theme-tasks/test/util.js
@@ -46,7 +46,7 @@ expect.extend({
 					pass = false;
 					message = `Path '${path}' is not a folder`;
 				}
-				else if (fs.readdirSync(path).length == 0) {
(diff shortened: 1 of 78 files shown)


maintenance/projects/js-themes-toolkit-v8-x/packages/liferay-theme-tasks/test/util.js
```diff
@@ -321,7 +321,7 @@ function cleanTempTheme(themeName, version, component, initCwd) {
 
 	cleanDirectory(tempPath);
 
-	if (initCwd != null) {
+	if (initCwd !== null && initCwd !== undefined) {
```
izaera, line 324: Can't we express this as `if (initCwd)` . I find `if (initCwd !== null && initCwd !== undefined)` a bit too complex to use it in every place of the code...

bryceosterhaus: My goal with using `initCwd !== null && initCwd !== undefined` is that it is safer for a wide refactor since `initCwd != null` is equal to `initCwd !== null && initCwd !== undefined`

If I was writing new code, I would likely use `if (initCwd)`, but I can't totally be sure that this bit of code would work with that unless I tested it thoroughly(which I didn't since I touched lots of code).

bryceosterhaus: >  If it is only for this PR, it's OK and we may refactor them when we come back to the code. But if it is going to be our usual pattern, I find it a bit too much for my refined taste

Just saw this from below. Yep, this is just for this PR and not something I intend for us to do. Only made that change for some safety in refactoring.

PR #727, opened 2021-11-04
feat: update liferay-cli to account for changes in Angular architecture

izaera:
We have found that the sample Angular application we provide with the JS Toolkit doesn't work well when component libraries and apps are integrated together.

The originating issue for this task was https://issues.liferay.com/browse/LPP-42949 which is really an integration issue, but the proposed fix will be included in our sample project so that customers can take advantage of it.

See https://issues.liferay.com/browse/LPP-42949?focusedCommentId=2555696&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2555696 for a rough explanation of what needs to be done.


bryceosterhaus (approved):
LGTM!

PR #730, opened 2021-11-04
feat: Add GitHub jira integration project

victorg1991:
This PR add the new tool to synchronized github issues with jira.

Sending this now in order to get some feedback about the code and the project structure :)

It works using webhooks, so for this to work we need to deploy this in some public server, and then configure each project adding a webhook pointing to the deployed app. For now it only creates a jira issue when a github issue is created. it should be very easy/fast to add other handlers.

Note: There is a lot of process.env scattered across the files, I plan to move this to a config file to make this more centralized, but it was easier for now just to use the process.env :) 

Let me know what do you think :)


bryceosterhaus:
Looks good so far!

Instead of deploying to a separate public server, can we just use github actions to automate this process? This would also be nice because it contains all of our logic to just github without worrying about separate deploys to another system.

[I noticed this project from atlassian](https://github.com/atlassian/gajira)

victorg1991:
Hey @bryceosterhaus, I haven't thought about that 🤔. 

It seems that the github actions doesn't offer a persistent storage between runs, we need that to map github issue id to jira issue id, otherwise once we add a comment in github or edit something we cannot know what ticket/comment is referencing. Also we cannot retry actions if they fail (jira may be down).

I think we can overcome this limitation using a 

external storage system, but that means splitting this tool across several system. Do you have any other idea how we can do that?

Since we probably reuse most of the code that handles the event, I'll see if I can have all the events handling done today and then we can decide if we want to move the code to github actions :)

bryceosterhaus:
> github actions doesn't offer a persistent storage between runs, we need that to map github issue id to jira issue id

Oh interesting, yeah I hadn't thought of that and that makes sense. I know you can use [Artifacts](https://github.com/actions/upload-artifact) which "allows you to share data between jobs and store data once a workflow is complete". But using that may be overly complicated than if you just stored this data somewhere else.

support/packages/github-to-jira-synchronization-tool/src/config/config.js
```diff
@@ -0,0 +1,53 @@
...
+
+let config = null;
+
+function loadConfig() {
+	if (config) {
+		return config;
+	}
+
+	try {
+		const json = fs.readFileSync(
```
izaera, line 17: :thinking: I think you can use `require()` for this

izaera: It would save you the readFileSync and the JSON.parse

support/packages/github-to-jira-synchronization-tool/src/constants.js
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+const PASSWORD = process.env.PASSWORD;
+const PORT = process.env.PORT || 5000;
+const PROJECT = process.env.PROJECT || 'IFI';
+const SECRET = process.env.SECRET;
+const USERNAME = process.env.USERNAME;
+
+module.exports = {
```
izaera, line 12: I like hiding this into a module, instead of placing `process.env.xxx`s all over the code :+1:

support/packages/github-to-jira-synchronization-tool/src/event-handlers/issueAssignedHandler.js
```diff
@@ -0,0 +1,53 @@
...
+		) {
+			return jiraClient.transitionIssue({
+				issueId: jiraIssue.key,
+				transition: JIRA_TRANSITIONS.reopen,
+			});
+		}
+	},
+};
+
+module.exports = issueAssignedHandler;
```
izaera, line 53: When there's just one entity exported in a .js file (that should be named as the entity per our frontend guidelines, as you do in this case :+1: ) I usually move the `module.exports` to the top, and do the export and the definition in the same place.

I'm not sure if we all do the same, though, but it seems more natural to me (similar to `export default function x()` in ES)

support/packages/github-to-jira-synchronization-tool/src/event-handlers/issueClosedHandler.js
```diff
@@ -0,0 +1,33 @@
...
+		if (jiraStatusNames !== JIRA_STATUS.closed) {
+			return await jiraClient.transitionIssue({
+				issueId: jiraIssue.key,
+				transition: JIRA_TRANSITIONS.close,
+			});
+		}
+	},
+};
+
+module.exports = issueClosedHandler;
```
izaera, line 33: Same as above

izaera:
LGTM :+1: 

I haven't reviewed the functionality (I assume the workflow is OK) but I like the code and how it is organized. Good job :-).

support/packages/github-to-jira-synchronization-tool/src/config/config.js
victorg1991: I though that using `require` with json files may have some problems. But maybe it is just in case the json is dynamic? and how require caches stuff? 🤔

support/packages/github-to-jira-synchronization-tool/src/event-handlers/issueAssignedHandler.js
victorg1991: Yeah, that makes more sense actually. I'm going to change that :)

victorg1991:
Thanks for the review @izaera, I've pushed the changes :)

support/packages/github-to-jira-synchronization-tool/src/config/config.js
izaera: Yeah, require() caches forever unless you explicitly clean it by hand. In any case, this is static as it is inside the source code so it's all good.

I would say a good criterion is:

1. Use require if the JSON is a static project resource (like in this case)
2. Use readFile in case the JSON is outside of the project and it may change or if there are myriads of JSON files to read and you don't want to cache them
3. Use require if the JSON is outside of the project but it won't change during the life of the process and you want to cache it (for example: I usually read the package.json file from liferay-npm-bundler because there's only one and it doesn't change while liferay-npm-bundler is executed. Also, because I access it from many places in the code, I take advantage of require()'s cache.

PR #738, opened 2021-11-05
feat: migrate shared bundle generation to @liferay/cli 

izaera:
Implementation of https://issues.liferay.com/browse/IFI-2718

This adds a new type of project to the generator so that we don't leave any gap in the previous implementation based on a Yeoman generator.

### DIFF
--- a/projects/js-toolkit/packages/liferay-cli/src/new/facet-shared-bundle/templates/src/index.js.ejs
+++ b/projects/js-toolkit/packages/liferay-cli/src/new/facet-shared-bundle/templates/src/index.js.ejs
@@ -1,0 +1,10 @@
+/**
+ * Place any initialization of this shared bundle in this file. That way, you
+ * can run:
+ *
+ * 		require('<%= name %>');
+ *
+ * in your dependent projects, before using this shared bundle and you will make
+ * sure that this code is executed once.
+ */
+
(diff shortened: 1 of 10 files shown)


projects/js-toolkit/packages/liferay-cli/src/new/facet-shared-bundle/templates/src/index.js.ejs
```diff
@@ -0,0 +1,10 @@
+/**
+ * Place any initialization of this shared bundle in this file. That way, you
+ * can run:
+ *
+ * 		require('<%= name %>');
```
bryceosterhaus, line 5: it may just be github, but these tabs look gigantic

bryceosterhaus (approved):
LGTM!

projects/js-toolkit/packages/liferay-cli/src/new/facet-shared-bundle/templates/src/index.js.ejs
izaera: :thinking: They are real tab characters (not spaces) so I guess it depends on what's displaying them...

izaera: I'll review the .ejs templates in any case, to make sure we use tabs/spaces everywhere.

izaera: Done. We use tabs in .ejs templates (like this one). Sadly prettier do not format these files, nor eslint checks them. They are so few that I don't think they deserve a custom SF process, but if they grow maybe we need to think something.

Issue #700, opened 2021-10-19
Upgrade node-sass to make it work for node LTS (14) and current (16)


izaera:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [X] :bug: Bug report
-   [ ] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [ ] :eight: v8.x
-   [ ] :nine: v9.x
-   [X] :keycap_ten: v10.x or higher

### Description

**Desired behavior:**

We are committed to make tooling work for at least LTS and, when possible, current versions of node.

**Current behavior:**

Right now the Themes Generator is failing to install for node 16 because we depend on node-sass 14.4.1 (determined running `yarn why node-sass` in this repo) which is intended to be used from node 14 (as shown in [the package description](https://www.npmjs.com/package/node-sass)).

**Repro instructions (if applicable):**

See [LPP-42803](https://issues.liferay.com/browse/LPP-42803)

bryceosterhaus:
> we are removing node-sass from liferay-portal

![](https://media.giphy.com/media/3o6Mbolqx8Ses8KXoQ/giphy.gif)

brunofernandezg:
Test comment (ignore)

PR #741, opened 2021-11-09
feat(theme-tasks): upgrade gulp-sass and node-sass for newer node versions

bryceosterhaus:
fixes https://github.com/liferay/liferay-frontend-projects/issues/700

This keeps original functionality that defaults theme tasks to use node-sass, but we upgraded both the gulp-sass and node-sass versions to work with node 14 & 16.

We could potentially remove node-sass and require that as a peer dependency if they don't want to use dart sass.

Any thoughts on if we should move the default to dart-sas?


bryceosterhaus:
In my last commit I moved dart-sass to be default and require anyone who wants to use node-sass to install it themselves

izaera:
LGTM

PR #744, opened 2021-11-10
feat(npm-scripts): add preflight check for yarn.lock validity

### LINKED ISSUE #743: Add preflight check to make sure yarn.lock does not reference local registry
Sometimes when using a local registry, your yarn.lock will reference that instead of the public registry. We need to add a preflight check to verify we don't commit yarn.lock with localhost references

bryceosterhaus:
fixes #743

This adds a check to make sure there is no reference to localhost or 127.0.0.1 in our yarn.lock


izaera:
Awesome! :raised_hands: 

LGTM

PR #748, opened 2021-11-11
chore: update engines field to match our node lts

bryceosterhaus:
Updating our engines specification to match node lts

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -19,7 +19,7 @@ jobs:
 
         strategy:
             matrix:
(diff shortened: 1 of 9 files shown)


izaera:
LGTM

Node LTS is now 16.3.0 :grimacing:

bryceosterhaus:
Screen Shot 2021-11-15 at 11 27 00 AM
🤦 

I'll update to 16 lol

bryceosterhaus:
this is why you don't make updates right at an LTS transision

PR #755, opened 2021-11-16
docs(target-platforms): update README.md

izaera:


izaera:
Going to merge this in 3, 2, 1....

bryceosterhaus:
@izaera what is with the huge yarn lock changes?

izaera:
@bryceosterhaus It's not the global yarn.lock, but the one at target-platforms. That's why I didn't include target-platforms in the main yarn structure of liferay-frontend-projects: because we need to include the deps from every portal version in the target platforms, we create an explosion of duplicated dependencies and version numbers. This is unavoidable, but at least it is mitigated by placing target-platforms in its own isolated yarn workspace and by keeping its code to a bare minimum (the workhorse is in the js-toolkit, projects portal-base and portal-adapt-base).

bryceosterhaus:
Oh gotcha, that makes sense. I don't think I realized it was nested in the target-platforms directory.

izaera:
Some (or all, I don't know for sure) of the projects at `maintenance` are detached from the global yarn workspace too.

PR #756, opened 2021-11-17
feat(liferay-cli): add new published target platforms

izaera:
This adds the target platforms I published yesterday to the list of available targets when `liferay new <project>` is run.

Right now I'm showing this menu:

imagen

I would like to find a more descriptive names, but that's all I could think of.

The main idea we need to convey is that by using a target platform for a specific version you get access to all our APIs through bundler-imports whereas choosing `portal-agnostic` (the `Liferay Portal CE (not using platform's APIs)` you don't. Sadly, expressing that intuitively, and in one line, seems very difficult to me.

### DIFF
--- a/projects/js-toolkit/packages/liferay-cli/bin/liferay.js
+++ b/projects/js-toolkit/packages/liferay-cli/bin/liferay.js
@@ -22,6 +22,7 @@ const {argv} = require('yargs')
 	)
 	.command('build', 'Build a project created with @liferay/cli')
 	.command('clean', 'Remove output directories')
+	.command('docs', 'Browse Liferay JavaScript Toolkit documentation')
 	.command(
 		'new <name>',
 		'Create a new project with the given name',
--- a/projects/js-toolkit/packages/liferay-cli/src/dependencies.json
+++ b/projects/js-toolkit/packages/liferay-cli/src/dependencies.json
@@ -43,6 +43,12 @@
 		"devDependencies": {}
 	},
 	"target-liferay": {
-		"platforms": ["portal-7.4-ga1", "portal-agnostic"]
+		"platforms": {
+			"portal-7.4": "Liferay Portal CE 7.4",
+			"portal-7.3": "Liferay Portal CE 7.3",
+			"portal-7.2": "Liferay Portal CE 7.2",
+			"portal-7.1": "Liferay Portal CE 7.1",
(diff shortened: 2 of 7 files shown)


bryceosterhaus:
I think the way you have it is fine for now. I think it would be rare for someone to use the portal-agnostic one. If we need to provide further docs, we could link back to our README or something to give more info.

izaera:
> we could link back to our README or something to give more info.

I like that. I will add some way to show the documentation in liferay-cli :+1: (probably printing a link to begin with, but we may open a browser in the future, for example. Or even provide text documentation similar to the one provided by `man`...

izaera:
I've added https://github.com/liferay/liferay-frontend-projects/pull/756/commits/77702f42fb0edaa5abf6cd8b99b1e87601e5409d that prints the link to the docs, like this:

imagen

I think it's OK as a first step towards something more powerful. I'm thinking in something like this, that I saw yesterday, from Rust: 

imagen

Rust has its own package manager named `cargo`, which has a command `doc` that opens the relevant documentation for any project. We wouldn't be able to do it because there's no official way to generate documentation for npm packages, but we can be inspired by it to create a good documentation tool for our JS Toolkit, I think...

izaera:
I'll defer merging this until I get some opinions on the `liferay docs` command (if it is OK to keep it or we prefer not to ship it yet). 

Since I need to prepare more PRs to make the switch from yeoman generator to liferay-cli happen, there's no need to rush.

bryceosterhaus:
@izaera I think thats a great starting point, just having a reference is helpful. Instead of a separate command of `docs` entirely, can we just add that info to `--help`? That way it might be a bit more findable, I don't know if I have ever seen a `docs` command before.

izaera:
:thinking: I don't think we can use `--help` because that's used by `yargs` to describe the usage of the tool (commands and switches).

Nevertheless, the command is very discoverable, IMO, because if you run `liferay` alone (or `liferay --help`) you get this:

imagen

So, there you have the `liferay docs`...

Also, the good thing about having a real command is that we can then implements things like `liferay docs search bundling`, `liferay docs search adapting`, or `liferay docs open online`, etc. to access different parts of the documentation (something similar to `man` and `apropos` in linux systems...).

bryceosterhaus:
Ah I see, I forgot it was built into yargs. That sounds good to me then!

izaera:
Merging this to reduce the PR queue, since it is approved.

PR #763, opened 2021-11-22
feat(liferay-cli): implement `liferay update-project` to update old yeoman-based projects

izaera:
This is part 1 of [IFI-2817](https://issues.liferay.com/browse/IFI-2817). We may also need to upgrade adapted projects, too. However, another option would be to tell people to remove the adaptation and readapt again. I will have a look at it, then decide what to do with it.


bryceosterhaus (approved):
LGTM!

PR #765, opened 2021-11-23
feat(liferay-cli): upgrade adapted projects

izaera:
This is part 2 of [IFI-2817](https://issues.liferay.com/browse/IFI-2817). It upgrades old adapted projects that were created with the yeoman generator.

I've tested it manually for the three supported projects types (Angular, React and Vue) and it works :+1:


bryceosterhaus (approved):
LGTM!

> I've tested it manually for the three supported projects types (Angular, React and Vue) and it works 👍

I haven't tested myself but I have no other option but to trust you 🙏

Issue #740, opened 2021-11-08
Next steps for JS Toolkit and target platforms


izaera:
This issue gathers the following steps we need to follow to replace the old Liferay JS Toolkit Yeoman generator by our new 
`@liferay/cli` tool so that we can progress in our migration from the [old JS Toolkit repo](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit) to the [new one](https://github.com/liferay/liferay-frontend-projects/tree/master/projects/js-toolkit), then evolve the `liferay-npm-bundler` from an AMD architecture to -hopefully- a [browser modules one](https://github.com/liferay-frontend/liferay-portal/pull/1533).

## Current status

✅ We have already merged the necessary [target platforms](https://github.com/liferay/liferay-frontend-projects/tree/master/target-platforms/packages) to cover the previous functionality provided by [liferay-npm-build-support](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit/packages/liferay-npm-build-support).

✅ We have also merged a functional `@liferay/cli` that can completely replace the [old Yeoman generator](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit/packages/generator-liferay-js).

✅ Our [documentation](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit/docs#readme) still points to the old artifacts.

✅ We still need to release some artifacts to the public npm registry.

## Following steps

Sorted by the order we should follow when making them happen:

1. :heavy_check_mark: Publish, at least, [portal-agnostic](https://github.com/liferay/liferay-frontend-projects/tree/master/target-platforms/packages/portal-agnostic) and all `portal-adapt-*` platforms so that old projects can continue working.
2. :heavy_check_mark: Compile a list of desired target platforms based on the current releases of Liferay Portal available.
3. :heavy_check_mark: Generate and publish the target platforms listed in step 2.
4. :heavy_check_mark: Update `@liferay/cli` with the published platforms. (#756)
5. :heavy_check_mark: Update our docs (copying them, as needed, to the v3 repo). (#757)
6. :heavy_check_mark: : Investigate how to migrate old projects, based on `liferay-npm-build-support` to `target-platforms`
    1. regular projects: :heavy_check_mark: https://github.com/liferay/liferay-frontend-projects/pull/763
[...]

PR #766, opened 2021-11-23
feat(theme-tasks): update theme dependencies to match Liferay 7.4.13

bryceosterhaus:
I pulled these version numbers from this release of Liferay portal, https://github.com/liferay/liferay-portal/commit/e64d3f8d5f0b61fe497d4728f1f2c878d8e589c1

### DIFF
--- a/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/devDependencies.js
+++ b/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/devDependencies.js
@@ -76,9 +76,9 @@ module.exports = {
 			default: {
 				'compass-mixins': strict('0.12.10'),
 				'gulp': gulpVersion,
-				'liferay-frontend-css-common': strict('6.0.0'),
-				'liferay-frontend-theme-styled': strict('6.0.2'),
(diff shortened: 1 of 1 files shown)


bryceosterhaus:
Would it be worth automating something here based on liferay-portal releases?

izaera:
> Would it be worth automating something here based on liferay-portal releases?

Not sure if it can be done... Where can we get that information from? I already grab information from GH in my target platforms script (reading yarn.lock for example) and it's quite easy, so maybe we can do something similar for this...

bryceosterhaus:
I was thinking we could possibly do a daily/weekly action to check the [tags on liferay-portal](https://github.com/liferay/liferay-portal/tags) and then if there is a new one, we start the publishing process or at minimum we could spawn a new issue on the repo.

izaera:
I like the idea. We can use it to publish target platforms too.

Can create a GH action that executes daily to check for new tags? That way we can see the results in our mailbox. And also, in the future, we may grow it to make the publication automatic, if we find the way. Like we discussed [here](https://github.com/liferay/liferay-frontend-projects/issues/740#issuecomment-970891049).

PR #769, opened 2021-11-24
test(js-toolkit): fix QA scripts

izaera:


izaera:
Merging this as it is trivial

PR #771, opened 2021-11-25
feat(generator-liferay-js): encourage users to use @liferay/cli

izaera:
Just show a message to users as per point 9 of #740


izaera:
@dsanz said he likes this in a private slack conversation, and @bryceosterhaus gave the OK too, so merging it :+1:

PR #775, opened 2021-12-01
chore: add workflow for checking for new Liferay releases

bryceosterhaus:
This workflow runs daily and creates a new issue for us in the case that a new liferay-portal version has been released


bryceosterhaus:
going to merge so that I can verify it works. Please still leave feedback, I think I may have missed some areas to check that we can include in the issue created

izaera:
this is fantastic

PR #789, opened 2021-12-13
fix: add check for liferay github owner so that this doesn't run on forked repos

bryceosterhaus:
I think this check should prohibit the workflow from running on forked repos


izaera:
:dancers:

PR #790, opened 2021-12-13
feat(npm-scripts): add path to global type definition file in liferay-portal

bryceosterhaus:
This adds support to our tsconfig so that they auto-generate and reference the global.d.ts file which we will be adding in https://issues.liferay.com/browse/LPS-131257


izaera:
:+1: LGTM

PR #792, opened 2021-12-15
chore(github-to-jira-synchronization): Better error handling and retry mechanism

victorg1991:
This is the second PR of the github-to-jira-integration tool:

This PR includes a bunch of enhancements after some testing of the tool, namely:

* Extensive logging to a file. This will allow easier debugging when it is running in the server
* Better events handling. We make sure that the issue or comment exists before doing any operation. This will allow to include issues/comments that are not in jira yet, due to the impasse between exalate expiring and this tool being ready
* Webhooks event are now being handled sequentially. It seems that when creating a github issue with some labels generate two webhooks, one for the issue and one for the label added. If we don't run those sequentially, we may end up with two issues being created. (This didn't happen in my machine but it did happen in the server :()
* Add a retry mechanism in case jira is down. When a webhook fail with a server error, we save the webhook in a file and retry later. For now it retries every hour (it seemed a reasonable time, since jira it is not usually down and when it is down it is not too much time) 

There is already a version deployed and listening for hooks from yesterday :)

Let me know what do you think

### DIFF
--- a/support/packages/github-to-jira-synchronization-tool/src/config/config.js
+++ b/support/packages/github-to-jira-synchronization-tool/src/config/config.js
@@ -63,7 +63,7 @@ function getLabelMapping(label) {
 
 	const labelMappings = mappingConfig.labelMappings;
 
-	return labelMappings[label?.replace('type:', '')];
+	return labelMappings[label?.replace('type:', '')?.trim()];
 }
 
 function getDefaultLabelMapping() {
--- a/support/packages/github-to-jira-synchronization-tool/src/config/config.js
+++ b/support/packages/github-to-jira-synchronization-tool/src/config/config.js
@@ -34,7 +34,8 @@ function loadConfig() {
 		config = JSON.parse(fs.readFileSync(CONFIGURATION_PATH, 'utf-8'));
 
 		return config;
-	} catch (error) {
+	}
+	catch (error) {
 		if (config) {
 			log('new config file is wrong, reusing past config');
 
--- a/support/packages/github-to-jira-synchronization-tool/src/config/config.js
+++ b/support/packages/github-to-jira-synchronization-tool/src/config/config.js
@@ -37,7 +37,20 @@ function getLabelMapping(label) {
 
 	const labelMappings = mappingConfig.labelMappings;
 
-	return labelMappings[label?.replace('type:', '')] || labelMappings.default;
+	return labelMappings[label?.replace('type:', '')];
 }
 
-module.exports = {getLabelMapping, getUserMapping, loadConfig};
+function getDefaultLabelMapping() {
+	const mappingConfig = loadConfig();
+
+	const labelMappings = mappingConfig.labelMappings;
+
+	return labelMappings.default;
+}
+
+module.exports = {
+	getDefaultLabelMapping,
+	getLabelMapping,
+	getUserMapping,
+	loadConfig,
+};
--- a/support/packages/github-to-jira-synchronization-tool/src/config/config.js
+++ b/support/packages/github-to-jira-synchronization-tool/src/config/config.js
@@ -37,7 +37,7 @@ function getLabelMapping(label) {
 
 	const labelMappings = mappingConfig.labelMappings;
 
-	return labelMappings[label] || labelMappings.default;
+	return labelMappings[label?.replace('type:', '')] || labelMappings.default;
 }
 
 module.exports = {getLabelMapping, getUserMapping, loadConfig};
--- a/support/packages/github-to-jira-synchronization-tool/src/config/config.js
+++ b/support/packages/github-to-jira-synchronization-tool/src/config/config.js
@@ -3,21 +3,46 @@
  * SPDX-License-Identifier: MIT
  */
 
+const fs = require('fs');
+const path = require('path');
+
+const log = require('../utils/log');
+
 let config = null;
+let lastModifiedTime = 0;
+
+const CONFIGURATION_PATH = path.resolve(__dirname, '../../mapping-config.json');
+
+function fileHasChanged() {
+	const stat = fs.statSync(CONFIGURATION_PATH);
+
+	if (stat.mtime > lastModifiedTime) {
+		lastModifiedTime = stat.mtime;
+
+		return true;
+	}
+
+	return false;
+}
 
 function loadConfig() {
-	if (config) {
+	if (!fileHasChanged() && config) {
 		return config;
 	}
 
 	try {
-		config = require('../../mapping-config.json');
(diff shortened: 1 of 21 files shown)


support/packages/github-to-jira-synchronization-tool/src/utils/runSequentially.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+let promise = Promise.resolve();
+
+module.exports = function runSequentially(fn) {
+	promise = promise.then(async () => {
```
victorg1991, line 9: This may look a bit rudimentary but it does fit my needs here

victorg1991: We may optimise this a bit and only process related webhooks sequentially but since the amount of webhooks is not too big I think it is not worth the effort

support/packages/github-to-jira-synchronization-tool/src/config/config.js
```diff
@@ -3,21 +3,45 @@
...
 
+const fs = require('fs');
+const path = require('path');
+
 let config = null;
+let lastModifiedTime = 0;
+
+const CONFIGURATION_PATH = path.resolve(__dirname, '../../mapping-config.json');
+
+function fileHasChanged() {
```
izaera, line 14: how can this change? do you modify it manually in the deployment environment without redeploying? :thinking:

support/packages/github-to-jira-synchronization-tool/src/config/config.js
```diff
@@ -3,21 +3,45 @@
...
 
 	try {
-		config = require('../../mapping-config.json');
+		config = JSON.parse(fs.readFileSync(CONFIGURATION_PATH, 'utf-8'));
 
 		return config;
 	}
 	catch (error) {
+		if (config) {
+			console.error('new config file is wrong, reusing past config');
```
izaera, line 38: rock solid :+1:

support/packages/github-to-jira-synchronization-tool/src/jira/JiraRetryHandler.js
```diff
@@ -0,0 +1,106 @@
...
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+const isJiraUp = require('../utils/isJiraUp');
+
+module.exports = class JiraRetryHandler {
```
izaera, line 11: We need a RetryHandler for JIRA :scream:? ... :roll_eyes:

support/packages/github-to-jira-synchronization-tool/src/config/config.js
victorg1991: Yep, that was my idea. This can happen if Bruno has to add some user mapping while the app is running. This way we can avoid redeploying the app for that minor change

support/packages/github-to-jira-synchronization-tool/src/utils/runSequentially.js
izaera: I had something similar in the bundler for file ops and after some tests found out that it was better to do things sequentially than in parallel :sweat_smile: 

In general, sequential leads to more throughput and less interactivity. But given that this is a daemon/service, it may even make sense, who knows without testing :shrug: 

TL;DR: if it works, leave it like this. There's time to optimize it.

support/packages/github-to-jira-synchronization-tool/src/config/config.js
victorg1991: 😂  I didn't want the tool to fail when hot-changing the configuration. Once the tool starts it checks the config and fails if the config is correct, but if we at least one valid configuration is better to stick to that instead of aborting I think 🤔  what do you think?

support/packages/github-to-jira-synchronization-tool/src/jira/JiraRetryHandler.js
victorg1991: Yeah, in case jira is down for some time we don't want to lose those events :'( historically it has been down just a couple of times, so it doesn't seems like a big deal

izaera:
The code looks OK: pretty organized and well thought :+1: 

One doubt I have is how is all this setup: where is it installed?, who has permissions=, etc. so that someone without any knowledge (like me) could recreate it if it was needed. I've seen [the README](https://github.com/liferay/liferay-frontend-projects/blob/master/support/github-to-jira-synchronization/README.md) but I though this was going to run instead of Exalate, not inside Exalate. Am I right?

victorg1991:
This is are replacement of exalete, this is installed in one of the liferay spain servers, I do have plans to add documentation about this, credentials, how to deploy it, how to configure it, etc :) 

Also I'll have to send a pr removing the past configuration https://github.com/liferay/liferay-frontend-projects/blob/master/support/github-to-jira-synchronization/README.md that is no longer relevant

support/packages/github-to-jira-synchronization-tool/src/jira/JiraRetryHandler.js
izaera: No, no, I like it. I found it a very good idea :+1: 

As long as the user gets immediate feedback if he changed the config and the load went wrong it's good. Otherwise (if the user won't be clearly notified) it may be a cause of trouble if the user looks at the config and think it's right but the application does a different thing.

support/packages/github-to-jira-synchronization-tool/src/config/config.js
izaera: No, no, I like it. I found it a very good idea +1

As long as the user gets immediate feedback if he changed the config and the load went wrong it's good. Otherwise (if the user won't be clearly notified) it may be a cause of trouble if the user looks at the config and think it's right but the application does a different thing.

support/packages/github-to-jira-synchronization-tool/src/jira/JiraRetryHandler.js
izaera: Yes, we all know JIRA :facepalm:

support/packages/github-to-jira-synchronization-tool/src/config/config.js
victorg1991: Actually I realized yesterday that with console.error they won't get immedita feedback, going to use log here to put this message in the log file

PR #793, opened 2021-12-17
fix: don't break deployment for shared bundle projects

izaera:
See https://issues.liferay.com/browse/IFI-2927

I'm going to merge this as soon as it becomes green to be able to release a new version.


izaera:
I said it :shrug: :ninja:

PR #807, opened 2022-01-10
feat(eslint-plugin): add new rule against A.Url

kresimir-coko:
Pretty straightforward, this PR adds a new rule that throws an error when `new A.Url()` is added to the code.

I didn't go with an autofix solution because we don't have any preexisting mentions, and because the returned object has different methods on it compared to the vanilla URL object.

### DIFF
--- a/projects/eslint-plugin/rules/aui/lib/rules/no-url.js
+++ b/projects/eslint-plugin/rules/aui/lib/rules/no-url.js
@@ -8,11 +8,29 @@ const message = 'use the vanilla URL class instead of A.Url';
 module.exports = {
 	create(context) {
 		return {
-			NewExpression(node) {
+			CallExpression(node) {
+				const {callee} = node;
+
 				if (
-					node.callee.object &&
-					node.callee.object.name === 'A' &&
-					node.callee.property.name === 'Url'
+					callee &&
+					callee.property &&
+					callee.object.name === 'A' &&
+					callee.property.name === 'Url'
+				) {
+					context.report({
+						message,
+						node,
+					});
+				}
+			},
+			NewExpression(node) {
+				const {callee} = node;
+
+				if (
+					callee &&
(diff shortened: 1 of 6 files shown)


bryceosterhaus:
> we don't have any preexisting mentions

Hmm, since we don't have any existing cases of this in liferay-portal I think we are probably better off just not adding this rule for now. If for some reason we see it popping up, we can add it, but it seems like people already don't use A.Url.

That seem okay?

kresimir-coko:
> That seem okay?

Yea, sure, although [Brian himself did comment](https://github.com/brianchandotcom/liferay-portal/pull/107047#issuecomment-918144372) that it should be added so 🤷

bryceosterhaus:
@kresimir-coko I know we don't have any in portal, but have you run it in liferay-portal yet? I just want to make sure we don't have any false positives or errors thrown

projects/eslint-plugin/rules/aui/lib/rules/no-url.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+const message = 'use the vanilla URL class instead of A.Url';
+
+module.exports = {
+	create(context) {
+		return {
+			NewExpression(node) {
+				if (
+					node.callee.object &&
```
bryceosterhaus, line 13: I think we would also need to make sure `node.callee` exists. Unless it always exists in NewExpression nodes?

Also, I think we would also need to check for just `A.Url`, without the `new`. I can imagine someone might accidentally forget the `new` and so we should warn that as well.

kresimir-coko:
Hey @bryceosterhaus I've improved the rule by addressing your notes. I've ran the rule inside `liferay-portal` and it produced no errors.

bryceosterhaus (approved):
LGTM!

PR #815, opened 2022-01-12
feat(eslint-plugin): Add new rule that preferes length checks

kresimir-coko:
There are 583 errors in `liferay-portal` when I run this rule. We probably want it to autofix because of the number of errors, I assume? @bryceosterhaus 

It can probably be improved, just code-wise, so setting it as a draft for now.

### DIFF
--- a/projects/eslint-plugin/rules/general/lib/rules/prefer-length-check.js
+++ b/projects/eslint-plugin/rules/general/lib/rules/prefer-length-check.js
@@ -3,41 +3,58 @@
  * SPDX-License-Identifier: MIT
  */
 
-const message = 'prefer using .length instead of .length === 0 or .length > 0';
+const message = 'prefer using .length';
+const STRICT_EQUALITY = '===';
+const MORE_THAN = '>';
+const LESS_THAN = '<';
+const LENGTH = 'length';
 
 module.exports = {
 	create(context) {
 		return {
 			BinaryExpression(node) {
 				const {left, operator, right} = node;
-				const STRICT_EQUALITY = '===';
-				const MORE_THAN = '>';
-				const LENGTH = 'length';
 
 				if (
 					operator &&
-					left.property &&
-					left.property.name === LENGTH &&
-					right.value === 0
+					(operator === STRICT_EQUALITY ||
+						operator === MORE_THAN ||
+						operator === LESS_THAN)
 				) {
-					if (
-						operator === STRICT_EQUALITY ||
-						operator === MORE_THAN
-					) {
-						const source = context.getSourceCode();
-						let replacement;
+					const source = context.getSourceCode();
+					let replacement;
 
-						if (operator === STRICT_EQUALITY) {
-							replacement = `!${source.getText(left)}`;
-						}
-						else {
-							replacement = `!!${source.getText(left)}`;
-						}
+					if (
+						right.property &&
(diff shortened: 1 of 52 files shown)


projects/eslint-plugin/rules/general/lib/rules/prefer-length-check.js
```diff
@@ -0,0 +1,62 @@
...
+
+module.exports = {
+	create(context) {
+		return {
+			BinaryExpression(node) {
+				if (node.operator) {
+					if (
+						node.operator === '===' ||
+						node.operator === '<' ||
+						node.operator === '>'
```
bryceosterhaus, line 16: I think we would also need to check what the operator is comparing against. We only would want to check when explicitly checking against `0`


```js
// valid
if (foo.length > 5) {
}
```

bryceosterhaus (changes requested):
@kresimir-coko yeah we will probably need an autofix so that we don't have to manually update every case

projects/eslint-plugin/rules/general/lib/rules/prefer-length-check.js
```diff
@@ -0,0 +1,61 @@
...
+const message = 'prefer using .length instead of .length === 0 or .length > 0';
+
+module.exports = {
+	create(context) {
+		return {
+			BinaryExpression(node) {
+				const {left, operator, right} = node;
+				const STRICT_EQUALITY = '===';
+				const MORE_THAN = '>';
+				const LENGTH = 'length';
```
bryceosterhaus, line 15: These can be moved up and out of this expression, that we only create these variables once.

If we leave them here, we essentially are creating these variables _everytime_ eslint crawls the AST and sees a BinaryExpression

projects/eslint-plugin/rules/general/tests/lib/rules/prefer-length-check.js
```diff
@@ -0,0 +1,54 @@
...
+};
+
+const ruleTester = new MultiTester(parserOptions);
+
+const message = 'prefer using .length instead of .length === 0 or .length > 0';
+
+ruleTester.run('prefer-length-check', rule, {
+	invalid: [
+		{
+			code: `if (foo.length === 0) {}`,
```
bryceosterhaus, line 23: We additionally should test against `if (0 === foo.length) {}`

projects/eslint-plugin/rules/general/tests/lib/rules/prefer-length-check.js
```diff
@@ -0,0 +1,54 @@
...
+			errors: [
+				{
+					message,
+					type: 'BinaryExpression',
+				},
+			],
+			output: `if (!foo.length) {}`,
+		},
+		{
+			code: `if (foo.length > 0) {}`,
```
bryceosterhaus, line 33: Same as above, we should also check for if the arguments are flipped `0 < foo.length`

bryceosterhaus (changes requested):

projects/eslint-plugin/rules/general/lib/rules/prefer-length-check.js
kresimir-coko: This one was a test for you, and you passed

bryceosterhaus:
LGTM!

PR #821, opened 2022-01-18
feat: make adaptations work with create-react-app 5.0.0

### LINKED ISSUE #809: Cannot adapt a React App with react-scripts@5.0.0
> 👀 Please check the [troubleshooting guide](https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/docs/How-to-troubleshoot-your-setups.md) before reporting anything. It contains important information on how to fix or diagnose errors.

-   [ ] :thinking: Question
-   [x] :bug: Bug report

### Version (mark with `x`)

-   [x] :two: v2.x
-   [ ] :three: v3.x

### Description

Cannot adapt a new React app.

A React app generated with `react-scripts@5.0.0` cannot load once deployed.

I verified that `react-sctipts@5.0.0`'s build process generates the following js files:

- `main.<hash>.js`
- `787.<hash>.chunk.js`

while the generated `index.js` file inside `build.liferay` folder requires the following:

- `"runtime-main"`
-  `"2.chunk"`
- `"main.chunk"`

The problem can be solved by changing `react-scipts` to 4.x version in `package.json`'s dependencies.

You can verify this behavior by simply generating a new app with `npx create-react-app` and making sure you have `react-scripts@5.0.0` as a dependency.

I verified this problem with:
- `generator-liferay-js@2.28.2`
- Node `v14.15.0`

izaera:
This fixes #809 in both JS Toolkit v2 and v3.

I've tested both versions with the QA scripts and everything is working.

Once we merge this we need to release new versions of affected modules.


bryceosterhaus (commented):
Screen Shot 2022-01-19 at 3 12 51 PM

Are all these changes necessary for the yarn.lock? Seems like quite a bit of churn here.

izaera:
The changes to `yarn.lock` are due to the introduction of `@liferay/portal-adapt-base` as a dependency of the bundler presets.

I did that in commit cb2d5e3f67907d403ef2d8a668a966a8e1e97c02 and deprecated the old loaders in 296dffa494487b95143318525e96e31925031262 to avoid using duplicated code in toolkit v2 and v3.

But seeing how the dependencies increase it may be better to live with that duplication for some time... :thinking: 

What do you think @bryceosterhaus ? In any case we already have duplication in the build processes, so removing a bit is not going to fix everything and, at the same time, it has a huge impact in the dependency tree.

I'll wait for your opinion before merging this.

bryceosterhaus:
@izaera I'm not really sold either way... On one hand, this is a step in the right direction by removing the duplication but I also fear that with the amount of transitive dependencies brought in, we my inadvertently mess something up.

However, we can't really live in fear of messing things up and probably should just step forward and address any issues that pop up. Also, the dependency changes wouldn't look as drastic if we were actually removing the old solution rather than just deprecating it. So, I think I've actually convinced myself this is okay and likely expected. We will likely have dependency jumps when we update to our latest tooling while also keeping the old solution around during its deprecation period.

And at the end of the day, if this causes massive failures, we can always revert. So, I'd say go for it.

izaera:
Before merging and release I think I'll add some more code to check the react-scripts version of the project and warn if it is not the expected one. That way instead of failing, we signal the user on what to expect.

izaera:
Finally :tada: .

Going to merge and release.

PR #831, opened 2022-01-22
fix(dev-server): watcher directory lookup

ktor:
Before fix. Unexpected behavior happens in liferay workspace: directory loopup iterates up until `/` because directory containing `yarn.lock` is not called `modules`. Standard situation in Liferay workspace where `yarn.lock` is places in workspace root directory. `WD` is in computed correctly:
```
Liferay Dev Server is ready at http:127.0.0.1:3000
Watching changes at /modules/**/*.{css,js,jsx,scss,ts,tsx}
```
After the fix:
```
Liferay Dev Server is ready at http:127.0.0.1:3000
Watching changes at /home/ktor/development/projects/lukreo/liferay/modules/**/*.{css,js,jsx,scss,ts,tsx}
```

### DIFF
--- a/projects/js-toolkit/packages/dev-server/src/server.ts
+++ b/projects/js-toolkit/packages/dev-server/src/server.ts
@@ -55,6 +55,9 @@ function findPortalRoot(directory: string): string {
 			if (base === 'modules') {
 				return join(directory, '..');
 			}
+			else {
+				return directory;
+			}
 		}
 
 		if (dirname(directory) === directory) {


bryceosterhaus:
Hey @ktor thanks for the PR. Technically the `@liferay/dev-server` project is aimed at development in the `liferay-portal` repository specifically. However, this change seems pretty simple enough. Have you been able to confirm this works for liferay workspaces and does it work consistently? If it seems to work I think we can merge and release. This project is fairly experimental and we haven't done a ton of work on improving it quite yet, so I would suspect plenty of issues with it.

ktor:
Hi @bryceosterhaus,

I understand the part about `liferay-portal` development. 

My use case is within the workspace and I though the project is going to support that. Before workspace upgrade, with version `7.3` I've used `@clavis/lfr-js-portlet-utils` package with success. With `7.4` it stopped working so I was wandering if I should fix it or use sth else and found `@liferay/dev-server`.  

Watcher for `@liferay/dev-server` works with workspace after the fix but.. there is one more problem.

Server is not able to map proxy requests to local files. Now I wonder if it's a good place to make it work for workspace. I presume yes if you're willing to support that, otherwise it probably needs forking.

bryceosterhaus:
@ktor I think providing support for liferay workspaces makes sense to me, at least in the same way we support liferay-portal. Let me look into it a bit deeper this week and test it out. Thanks again for opening this issue

bryceosterhaus:
follow up at https://github.com/liferay/liferay-frontend-projects/pull/933/files

PR #833, opened 2022-01-25
fix(js-themes-toolkit): mark completion of upgrade:css

izaera:
See https://issues.liferay.com/browse/LPP-43860


izaera:
I'm merging this since it is a fix for an LPP and the code is trivial

PR #835, opened 2022-01-27
docs(eslint-plugin): Update eslint docs

### LINKED ISSUE #832: Verify and document rules and external libraries we consume in eslint README
We need to verify and update our README in our [eslint plugin](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/eslint-plugin/README.md). I don't think we currently have all rules listed, for example [no-anonymous-exports](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/eslint-plugin/rules/general/lib/rules/no-anonymous-exports.js) is not listed in our README.

kresimir-coko:
Fixes #832


bryceosterhaus:
LGTM!

@diegonvs I think the yarn.lock is out of sync on upstream and this change fixes that as well.

PR #840, opened 2022-02-02
fix(js-themes-toolkit-v10-x): Move theme-toolkit 10.x to maintenance and lock sass

bryceosterhaus:
Hey @izaera, I realized we forgot to create a 10.x maintainance directory when we updated to 11.x in theme-tasks. Can you take a look here and verified I did everything necessary? This all stemmed from an issue we had with dart-sass breaking change in a minor version update of theirs.


bryceosterhaus:
I'm also open to other ideas on how we best support these older versions, branching or something else?

izaera:
:scream: We hoped not to have to release any version after 10.0 :disappointed: 

In any case... We have the maintenance folder because it's easier to handle than branches, but we can switch to branches if we like. I don't think there will be much difference in repo size, and having the old versions available in the source tree is handy, because you can compare with the newer ones, so...

Maybe it makes sense to hide very old stable versions that we haven't touched in -say- one year behind a branch, so that we can prune the maintenance directory every once in a while. But for now, I wouldn't worry too much, I think.

izaera:
One thing you need to update, if you release a version 11 is this table :point_right: https://github.com/liferay/liferay-frontend-projects/tree/master/projects/js-themes-toolkit#readme

bryceosterhaus:
> We hoped not to have to release any version after 10.0 

whoops, I didn't realize that. We had to make a breaking change when we went from node 10 -> 16 and because we use dart sass by default. If we kept it at v10, we would break old branches relying on node 10 and libsass.

> having the old versions available in the source tree is handy, because you can compare with the newer ones, so
This has definitely been helpful for me, but I think we can possibly remove some old versions.

> Maybe it makes sense to hide very old stable versions that we haven't touched in -say- one year behind a branch, so that we can prune the maintenance directory every once in a while. But for now, I wouldn't worry too much, I think.

I like this idea, I will think through which ones we can remove. If only we had some statistics on how many people are relying on 8 & 9.

izaera:
> If only we had some statistics on how many people are relying on 8 & 9.

I think we may do without statistics: we can simply pay attention to how often we update those projects looking at git's history. In the end, that's the important thing, not how much people uses it. Because, see the AMD loader: everybody uses it, but we've touched it 2-3 at most since I began to work on this stuff, so it could be a candidate to almost-freezing :shrug:

PR #842, opened 2022-02-04
feat: add JS Toolkit generator and build for Remote Apps

izaera:
This PR adds a generator and a build for custom element Remote Apps. It needs to be tested on a Liferay Portal containing the changes in https://github.com/liferay-frontend/liferay-portal/pull/1869

----

One caveat: I'm not yet sure if the commits modifying the target platforms should be merged on top of what we have now, or if we need to release more versions before merging. This is because the last version we have released is 7.4-GA4 (I think) but we are already at GA-8 or so, so versions <=8 won't have react and react-dom exported.

So I'm sending this for review, but before releasing versions for all this, we may need to analyze that...


projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Project.ts
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	/**
+	 * Resolves a module in the context of the project (as opposed to the
+	 * context of the calling package which would just use a normal
+	 * `require.resolve()` call).
+	 * @param moduleName
+	 */
+	resolve(moduleName: string): string {
+		return resolve.sync(moduleName, {
```
bryceosterhaus, line 81: Do you know if it matters if the module has a `main` file or not? I know if you use `require.resolve()`, it will error saying it can't find the module if the `main` field doesn't exist. One work around for this I have seen is resolving directly too the package.json of the module and taking the directory from there.

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Start.ts
```diff
@@ -0,0 +1,15 @@
...
+import LiferayJson from '../../schema/LiferayJson';
+import Project from './Project';
+
+export default class Start {
+	readonly port: number;
+
+	constructor(_project: Project, liferayJson: LiferayJson) {
+		this.port = liferayJson.start?.port || 8081;
+	}
+}
```
bryceosterhaus, line 15: I'm curious why you went with a class for everything? Seems like this could have just been a plain object

projects/js-toolkit/packages/liferay-cli/src/dependencies.json
```diff
@@ -35,6 +35,13 @@
 		},
 		"devDependencies": {}
 	},
+	"facet-remote-app-react": {
+		"dependencies": {
+			"react": "^17.0.0",
+			"react-dom": "^17.0.0"
```
bryceosterhaus, line 41: Do we need to align these versions with the exact versions in liferay-portal?

projects/js-toolkit/packages/portal-base/src/build/bundler2.ts
```diff
@@ -0,0 +1,133 @@
...
+	runCompiler(project);
+	runBundler();
+}
+
+function copyAssets(project: Project): void {
+	const assetFiles = findFiles(project.srcDir, (dirent) => {
+		const lowerCaseName = dirent.name.toLowerCase();
+
+		return (
+			!lowerCaseName.endsWith('.js') && !lowerCaseName.endsWith('.scss')
```
bryceosterhaus, line 34: Do we need to also copy .css files and possibly images?

bryceosterhaus (commented):
Looks pretty good, I didn't go super in-depth and test but from an initial look I don't see any major red flags.

One part I am not sure I fully understand is the webpack stuff for custom elements. Could you elaborate on that a bit or point me to any previous discussions we had on it?

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Deploy.ts
```diff
@@ -0,0 +1,51 @@
...
+		const file = this._project.dir.join('.liferay.json');
+
+		const liferayJson = JSON.parse(
+			fs.readFileSync(file.asNative, 'utf8')
+		) as LiferayJson;
+
+		liferayJson.deploy = liferayJson.deploy || {};
+		liferayJson.deploy.path = this.dir.asPosix;
+
+		fs.writeFileSync(
```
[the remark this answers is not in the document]
izaera: Mmm, not really, because it should point to `portal/bundles/osgi/deploy` or similar folders. So, they should exist right before the project is deployed.

:thinking: In fact, if it doesn't exist, it is probably better to fail the deployment so that the user realizes that the project is being deployed to a wrong place...

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Dist.ts
```diff
@@ -0,0 +1,34 @@
...
+			case 'bundler2':
+			default: {
+				/* eslint-disable-next-line @typescript-eslint/no-var-requires */
+				const bundler2Project = require('liferay-npm-build-tools-common/lib/project');
+
+				this.dir = new FilePath(
+					bundler2Project.jar.outputDir.asNative
+				).resolve();
+				this.file = this.dir.join(bundler2Project.jar.outputFilename);
+				break;
```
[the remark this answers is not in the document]
izaera: That was to account for the case where the build type is not configured (for backwards compatibility). However, because we have the `Project` class that simulates configuration defaults when they are not present, I think I can refactor this and make it behave as you say, which is much cleaner :+1:

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Project.ts
izaera: I think require.resolve works in the absence of a `main` entry if there's and `index.js` file. That's the only automagic case I think...

However, this method has been copied from the bundler3 and bundler2 projects and should be agnostic to that. This is because it is a low level primitive that we need in many occasions: because we are implementing dev tools, we sometimes need to require things inside the tool itself, but many other times, we need to import things relative to the project's directory. That's what these `require` and `resolve` function in `Project` are for.

So that then you can say things like:

```javascript
   const plugin = project.require('a-babel-plugin'); // we get the plugin from user's directory
   const babel = require('babel');                           // we get babel from us, because we are providing it

   babel.use(plugin);
```

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Start.ts
izaera: Mainly because I copied it from bundler2 and bundler3, and it's how it was structured there. In those two tools, the configuration options grew big enough to make it worth separating it into functional scopes (just to make the code easier to navigate).

So, even if it can be one single class now, I'm almost sure it will grow. If it doesn't grow, this will be a YAGNI pattern but still, I don't think it hurts separating things by scopes (IMO), so... :shrug:

projects/js-toolkit/packages/liferay-cli/src/dependencies.json
izaera: I had the same doubt :sweat_smile: , but then realized that we don't. This is because these are only used when you don't want to import the ones from Portal, so we can default to the latest ones in that case, maybe.

When you don't use the `platform-agnostic` target platform, you get `react` from the target platforms itself, transitively, so we don't need to put anything in the project's `package.json` file. :tada:

projects/js-toolkit/packages/portal-base/src/build/bundler2.ts
izaera: Yes, that's what we do: copy everything but `.js` and `.scss` files (those are processed with babel and sass so those tools take care of generating them in the build dir).

Notice that the conditions are negated: `lowerCaseName.end != '.js' and lowerCaseName.end != '.scss'`

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Start.ts
izaera: Also, you can write things like `project.deploy.dir` and `project.build.dir` which look beatiful :heart_eyes: :joy:

izaera:
> One part I am not sure I fully understand is the webpack stuff for custom elements. Could you elaborate on that a bit or point me to any previous discussions we had on it?

There have been so many discussions... :sweat_smile: 

OK so, as it has been implemented in the end, I thinks it's pretty simple. Let's see...

Basicallly, we use webpack to do everything. That creates a self-contained .js file with everything you need to run the custom element and, because the custom elements are registered by the user (with the `customElements.define()` API) we don't need to do anything, like we did in bundler2 to wrap things into `Liferay.Loader.define()` calls, so we are safe.

The only con of this approach is what happens if you want to share, for example, `react` between several custom elements or use the one we use in portal. For that case we use the [externals configuration option of webpack](https://webpack.js.org/configuration/externals/). 

If you declare `react` as an external of [type module](https://webpack.js.org/configuration/externals/#externalstypemodule) what webpack does is:

1. Don't bundle `react` inside the self-contained .js file
2. Put an `import react from 'react';` on top of the self-contained bundle

And that's all. The import on top makes the browser wire the `react` variable to our (portal) exported `react` (which is the same the AMD Loader is exporting) and that makes all code use the same `react`. That's pretty much everything :shrug: .

Of course, we need the import map to make the import generated by webpack work because the browser doesn't know what `'react'` in `import x from 'react'` is. But we could even implement it without import maps, if we wanted, by simply rewriting the `'react'` to some well known URL of the portal (for instance: `/o/exported-pretty-things/react.js`, or anything we decide). We have plenty of space for manoeuvres in the future, should we need to change things...

bryceosterhaus:
> There have been so many discussions... 

haha yeah sorry about re-hashing this again, I feel like I had all this info in parts but never put it fully together in my head. Your description was great though and helpful as a big-picture view. Thanks for re-explaining yet again.

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Project.ts
bryceosterhaus: > works in the absence of a main entry if there's and index.js file.

Oh that makes sense. I had previously seen a case where resolve was failing but it lacked both a main and an index.js

projects/js-toolkit/packages/portal-base/src/build/bundler2.ts
bryceosterhaus: 🤦  my apologies. I can't read good

izaera: NP, negations in logical conditions are seldom a good thing... :roll_eyes:

izaera:
Going to merge this as we can iterate over if something needs change.

PR #853, opened 2022-02-15
feat(@liferay/npm-scripts): Give Liferay Workspaces ability to use our Prettier setup with the VSC extension

### LINKED ISSUE #51: Visual Studio code settings for prettier not working in Liferay workspace
for liferay-npm-scripts there is some description about how to get liferay's prettier settings working in vs code: 

https://www.npmjs.com/package/liferay-npm-scripts?activeTab=readme#visual-studio-code

this however assumes you're working in portal and not in liferay workspace
by making these edit hacks i was able to get it to work for one of my modules in workspace:

```        //const root = getPortalRoot(filepath);
        const root = "/home/allenz/liferay/lfris-www/liferay/modules/apps/resources/resources-web";
        if (root) {
//              const modules = path.join(root, 'modules/node_modules');
                const modules = path.join(root, 'node_modules');
```

could that be built into the prettier wrapper script? that would be awesome!

kresimir-coko:
Fixes #51 

This one was a long journey on my part, getting the Workspace working and then Prettier started causing me issues, etc.

To get this done, I utilized the [debug](https://github.com/kresimir-coko/liferay-frontend-projects/blob/2bcdc8175c4b8c4f8862dfdc8c2253264e2feaab/projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js#L20-L35) function in order to get some logging happening which was helpful but took a little bit of figuring out. After I made sure this works in the local Liferay Workspace, I also tested it in this repository and in `liferay-portal` to make sure it properly works since I did split up the conditioning.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
+++ b/projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
@@ -992,6 +992,33 @@ module.exports = {
 	version: '2.1.2',
 };
 
+function getWorkspaceRoot(filepath) {
+	let current = filepath;
+
+	while (true) {
+		const parent = path.dirname(current);
+
+		const candidate = path.join(parent, 'settings.gradle');
+
+		if (fs.existsSync(candidate)) {
+			current = candidate;
+
+			break;
+		}
+		else if (parent === current) {
+
+			// Can't go any higher.
+
+			return null;
+		}
+		else {
+			current = parent;
+		}
+	}
+
+	return path.dirname(current);
+}
+
 /**
  * Returns the root directory if `filepath` is inside a liferay-portal
  * checkout, otherwise returns null.
@@ -1080,10 +1107,10 @@ function prepare(filepath) {
 
 	let file = 'prettier';
 
-	let root = getPortalRoot(filepath);
+	const workspaceRoot = getWorkspaceRoot(filepath);
 
-	if (root) {
-		const modules = path.join(root, 'modules/node_modules');
+	if (workspaceRoot) {
+		const modules = path.join(workspaceRoot, 'node_modules');
 
 		let scripts = path.join(modules, '@liferay/npm-scripts');
 
@@ -1106,12 +1133,40 @@ function prepare(filepath) {
 			file = fallback;
 		}
 	}
-	else {
-		root = getFrontendProjectsRoot(filepath);
 
-		if (root) {
+	const portalRoot = getPortalRoot(filepath);
+
+	if (portalRoot) {
+		const modules = path.join(portalRoot, 'modules/node_modules');
+
+		let scripts = path.join(modules, '@liferay/npm-scripts');
+
+		let wrapper = path.join(scripts, 'src/scripts/prettier.js');
+
+		if (!fs.existsSync(wrapper)) {
+			scripts = path.join(modules, 'liferay-npm-scripts');
+
+			wrapper = path.join(scripts, 'src/scripts/prettier.js');
+		}
+
+		const fallback = path.join(modules, 'prettier/bin-prettier.js');
+
+		if (fs.existsSync(wrapper)) {
+			dir = scripts;
+			file = wrapper;
+		}
+		else if (fs.existsSync(fallback)) {
+			dir = modules;
+			file = fallback;
+		}
+	}
+
+	const frontendProjectsRoot = getFrontendProjectsRoot(filepath);
+
+	if (frontendProjectsRoot) {
+		if (frontendProjectsRoot) {
 			const scripts = path.join(
-				root,
+				frontendProjectsRoot,
 				'projects/npm-tools/packages/npm-scripts'
 			);
 
--- a/projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
+++ b/projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
@@ -1108,37 +1108,22 @@ function prepare(filepath) {
 	let file = 'prettier';
 
 	const workspaceRoot = getWorkspaceRoot(filepath);
-
-	if (workspaceRoot) {
-		const modules = path.join(workspaceRoot, 'node_modules');
-
-		let scripts = path.join(modules, '@liferay/npm-scripts');
-
-		let wrapper = path.join(scripts, 'src/scripts/prettier.js');
-
-		if (!fs.existsSync(wrapper)) {
-			scripts = path.join(modules, 'liferay-npm-scripts');
-
-			wrapper = path.join(scripts, 'src/scripts/prettier.js');
-		}
-
-		const fallback = path.join(modules, 'prettier/bin-prettier.js');
-
-		if (fs.existsSync(wrapper)) {
-			dir = scripts;
-			file = wrapper;
-		}
-		else if (fs.existsSync(fallback)) {
-			dir = modules;
-			file = fallback;
-		}
-	}
-
 	const portalRoot = getPortalRoot(filepath);
 
-	if (portalRoot) {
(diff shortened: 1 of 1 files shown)


projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
```diff
@@ -1106,12 +1133,40 @@ function prepare(filepath) {
...
+		}
+		else if (fs.existsSync(fallback)) {
+			dir = modules;
+			file = fallback;
+		}
+	}
+
+	const frontendProjectsRoot = getFrontendProjectsRoot(filepath);
+
+	if (frontendProjectsRoot) {
```
izaera, line 1166: Is this if duplicated? Or is GH UI playing with me? :thinking:

izaera:
If it works... LGTM :-)

projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
kresimir-coko: It is duplicated, gonna fix

projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
```diff
@@ -1106,31 +1133,57 @@ function prepare(filepath) {
...
+
+		if (fs.existsSync(wrapper)) {
+			dir = scripts;
+			file = wrapper;
+		}
+		else if (fs.existsSync(fallback)) {
+			dir = modules;
+			file = fallback;
+		}
+	}
```
bryceosterhaus, line 1162: This logic looks to be the exact same as the logic above, one for portalRoot and one for workspaceRoot. Can we only keep only block of this logic and instead do

```js
const portalOrWorkspaceRoot = getPortalRoot(filepath) || getWorkspaceRoot(filepath);

if (portalOrWorkspaceRoot) {
//...
```

bryceosterhaus (changes requested):
I haven't tested this yet. Could you give a brief summary of how to test this?

Also,  I made a note about reducing some repeated logic.

kresimir-coko:
@bryceosterhaus you can test it by creating a workspace, adding the liferay-npm-scripts package, and saving the file (if that triggers format for you).

I can't remember all the details now, but you can use the `debug` function inside the workspace relevant parts to confirm that prettier reaches the code. It requires closing and reopening VSC each time you do it.

projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
```diff
@@ -992,6 +992,33 @@ module.exports = {
...
 
+function getWorkspaceRoot(filepath) {
+	let current = filepath;
+
+	while (true) {
+		const parent = path.dirname(current);
+
+		const candidate = path.join(parent, 'settings.gradle');
+
+		if (fs.existsSync(candidate)) {
```
bryceosterhaus, line 1003: We will need to do more than just check if settings.gradle exists because this file exists at `liferay-portal/modules/settings.gradle` as well. I think you also need to read the content of the file and check for the workspace plugin

```suggestion
		if (fs.existsSync(candidate) && fs.readFileSync(candidate).includes('apply plugin: "com.liferay.workspace"')) {
```

projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
```diff
@@ -1080,11 +1107,23 @@ function prepare(filepath) {
...
 	let file = 'prettier';
 
-	let root = getPortalRoot(filepath);
+	const workspaceRoot = getWorkspaceRoot(filepath);
+	const portalRoot = getPortalRoot(filepath);
 
-	if (root) {
-		const modules = path.join(root, 'modules/node_modules');
+	const portalOrWorkspaceRoot =
+		getPortalRoot(filepath) || getWorkspaceRoot(filepath);
```
bryceosterhaus, line 1114: We are adding some unnecessary work her, we should just crawl the directories once

```js
	const workspaceRoot = getWorkspaceRoot(filepath);
	const portalRoot = getPortalRoot(filepath);

    const portalOrWorkspaceRoot = portalRoot || workspaceRoot;
```

Even better, we may consider creating a generic `findRoot` function that only has to crawl the file structure once and then returns what _type_ of root it is, whether it be a workspace or liferay-portal. This would allow us to avoid spending extra resources on crawling once for workspace and once for portal

projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
```diff
@@ -1106,31 +1145,30 @@ function prepare(filepath) {
...
+	else {
+		const extension = require('vscode').extensions.getExtension(
+			'esbenp.prettier-vscode'
+		);
+
+		if (extension) {
+			dir = path.join(extension.extensionPath);
+			file = path.join(dir, 'node_modules/prettier');
 		}
 	}
```
bryceosterhaus, line 1173: I think we are always ending up in this `else` branch and thus always taking the extensions version of prettier

bryceosterhaus: Yeah I think this is the case why you don't see it working properly, I just tested locally and I am always ending up in this else block and not using our custom prettier rules

bryceosterhaus (changes requested):

projects/npm-tools/packages/npm-scripts/contrib/prettier/prettier.js
kresimir-coko: This is a good suggestion. Here's what I ended up with, it's faulty and untested:
```js
function findRoot(filepath){
	let current = filepath;

	while (true) {
	const parent = path.dirname(current);

	const portalRootCandidate = path.join(parent, 'portal-web');
	const frontendProjectsRootCandidate = path.join(parent, '.liferay-frontend-projects');
	const workspaceRootCandidate = path.join(parent, 'settings.gradle');

	const candidates = [portalRootCandidate, frontendProjectsRootCandidate, workspaceRootCandidate];

	const candidate = candidates.find(candidate => {
		if (candidate === portalRootCandidate) {
			return fs.existsSync(
				path.join(current, 'docroot/WEB-INF/liferay-web.xml')
			) ? true : false;
		}

		if (candidate === workspaceRootCandidate) {
			return fs.existsSync(candidate).includes(
				'apply plugin: "com.liferay.workspace"'
			) ? true : false;
		}

		return fs.existsSync(candidate);
	})

		if (candidate) {
			current = candidate;

			break;
		}
		else if (parent === current) {

			// Can't go any higher.

			return null;
		}
		else {
			current = parent;
		}
	}
}
```

I wonder how will the `prepare` function be able to tell what root it is, like we are currently doing:
```
const workspaceRoot = getWorkspaceRoot(filepath);

if (workspaceRoot) {
...
}
```

There are definitely improvements to the `findRoot` function that would make it possible, but I just don't see it right now. It's gonna require a bit more time than what I have today, so I won't push anything since it won't do anything good.

kresimir-coko:
It was actually me deduplicating the `if` in the 1st commit after @izaera commented that bugged it to always use the `else` haha 😄 

I've pushed the changes that tackle the easy things, we can perhaps leave the `findRoot` optimization for a different task, but I'd be glad to be the one to handle that, it sounds fun (and to learn more about the node filesystem thingies).

I've tested this in my workspace and on saving it properly adds single quotes around the import statements.

bryceosterhaus:
Looks good! I added [one commit](https://github.com/liferay/liferay-frontend-projects/pull/853/commits/40cc45e07a416ca523c8001ed8314906b59a76cd) just moving that variable near where it is called.

PR #857, opened 2022-02-23
chore(alloy-font-awesome): update liferay-font-awesome version

kresimir-coko:


bryceosterhaus:
Hey @kresimir-coko, just a heads up, version bumps are usually done in the release process and I noticed we hadn't actually released this version yet. I went ahead and added the release, but typically we would only update the version in the release process.

PR #860, opened 2022-02-23
fix(local-npm): fixes for local-npm tool

izaera:


izaera:
I'm going to fix this since it's internal code.

Issue #858, opened 2022-02-23
Render <script> nodes with type="module" attribute


izaera:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [ ] :bug: Bug report
-   [X] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Description

This is so that we can use ESM syntax (import and export) to interoperate ES modules with AMD loader.

izaera:
This is done

PR #862, opened 2022-02-23
feat(amd-loader): load scripts as browser modules (i.e.: type='module')

izaera:
see #858


izaera:
I'm going to merge this since I need to release a new version of the loader to integrate it in liferay-portal

PR #865, opened 2022-02-24
fix(local-npm): prevent local-npm from screwing NodePlugin

izaera:
see commit for more info


izaera:
The first commit was in my local copy but wasn't pushed to upstream because the release scripts of amd-loader are not automating that task.

PR #866, opened 2022-02-25
fix(local-npm): fix override of @liferay/npm-scripts once more

izaera:
Still trying to know what would be the correct sequence to make yarn
reinstall an existing version.

Note that this is forcing yarn to install a newer version with a different
checksum (to make the developer life easier) so it's natural that it
refuses to do it unless forced. That's why I've needed to change this so
many times.

Let's hope this is the last time...


izaera:
Merging this to continue other work.

PR #868, opened 2022-02-25
feat(npm-scripts): add a webpack based build leveraging ESM

### LINKED ISSUE #859: Add a webpack based build leveraging ESM

-   [ ] :thinking: Question
-   [ ] :bug: Bug report

### Description
The idea is to get rid of the bundler v2 to build portal modules so that we may be able, some day, to make the whole portal work using ESM instead of AMD loader.

We need to keep AMD <-> ESM bridges until we remove all traces of AMD so that portal doesn't break.

In the future we may need to keep the ESM -> AMD bridges to avoid breaking customer's projects.

izaera:
Closes #859 

This implements the first viable webpack+ESM based build for liferay-portal. I need to merge it to test it in the CI environment.

It appears to work locally, but I want to run a thorough test to see what happens.


izaera:
:ninja:-merger strikes again! :scream:

projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdExportsBridges.js
```diff
@@ -0,0 +1,114 @@
...
+		bridgeSource += `"${namespacedScopedPkgName}@${pkgJson.version}/index"`;
+		bridgeSource += `, ['module'], function (module) {`;
+
+		bridgeSource += '\n';
+
+		bridgeSource += `  module.exports = esModule;`;
+
+		bridgeSource += '\n';
+
+		bridgeSource += `});`;
```
bryceosterhaus, line 92: Could we use a larger template literal here to make it a bit more readable?

projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdExportsBridges.js
```diff
@@ -0,0 +1,114 @@
...
+			bridgeSource,
+			'utf8'
+		);
+	});
+}
+
+function getPackageTargetDir(pkgJson) {
+	const {name, version} = pkgJson;
+
+	let targetFolder = name.replace('/', '%2F');
```
bryceosterhaus, line 105: Rather than calling `replace` can we use `encodeURIComponent`? I think it might be a bit more explicit that way. Or do we _only_ want to encode `/`?

bryceosterhaus (commented):
Looks pretty good to me from a cursory glance. I am eager to test it out locally though and see it in action.

I really like how verbose you made the output and providing helpful feedback to the user.

projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdExportsBridges.js
izaera: Nope. This is copied from JS Toolkit v2 and this is meant to only replace `/` by `%2F` so that we keep a flat structure inside the JAR (instead of having a nested folder for scoped packages).

projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdExportsBridges.js
izaera: Yes, we could.

izaera: And I will, in the next PR :joy:

bryceosterhaus: 👍  no rush

PR #870, opened 2022-03-01
feat(github-to-jira-sync-tool): Add the /status page that displays a message if the tool is running

### LINKED ISSUE #867: Create a status page to easily check if the GH-Jira sync is down
Can you think of an easy way to know if the app is down? Because right now we can only know it if the sync doesn't work or connecting to the server and checking the logs or the running processes. Not saying that we need to get a notification, but maybe a very simple status web that says "I'm up"?

Asked Victor about this possibility and he says it makes sense. Right now the server is accesible without the VPN
https://github-jira-synchronization.liferay.org.es/
Maybe it would be possible to create
https://github-jira-synchronization.liferay.org.es/system_status
And this could be a very simple page that says "up and running" or "error, not found"

kresimir-coko:
Fixes #867 

This PR utilizes [Octokit's middleware onUnhandledRequest](https://github.com/octokit/app.js/#createnodemiddlewareapp-options) to add a check for the `/status` page, displaying a simple message if it finds the url.

### DIFF
--- a/support/packages/github-to-jira-synchronization-tool/src/Server.js
+++ b/support/packages/github-to-jira-synchronization-tool/src/Server.js
@@ -10,12 +10,30 @@ class Server {
 	constructor(secret, port) {
 		this.port = port;
 		this.webhooks = new Webhooks({secret});
+		this.responseHead = {
+			'Content-Type': 'application/json',
+		};
+
+		this.handleStatusRequest = (req, res) => {
+			if (req.url === '/status' && req.method === 'GET') {
+				res.writeHead(200, this.responseHead);
+				res.write('The server is running!');
+				res.end();
+			}
+			else {
+				res.writeHead(404, this.responseHead);
+				res.end(JSON.stringify({message: 'Route not found'}));
+			}
+		};
+
+		this.middleware = createNodeMiddleware(this.webhooks, {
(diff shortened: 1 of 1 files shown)


bryceosterhaus:
Hey @kresimir-coko, just so I understand the full context. What would happen in the two cases of 1) if the server is active, 2) the server crashes and isn't running.

And is the only goal here to provide a endpoint to see if the server is currently online?

brunofernandezg:
Thanks for this @kresimir-coko !!

> And is the only goal here to provide an endpoint to see if the server is currently online?

@bryceosterhaus yes, for now. I requested this because the more time the app is down the critical. If it crashes and we re-start it in a few hours we won't loose any data. The only way to know if it's down right now is if the sync doesn't work or accessing the server, so we wanted to have an easier way of checking it for the person that replaces me during my paternity leave (and if it's down will ask the squad to re-start it).

There would be lots of things we could do to improve this tool including investigating and fixing the crash, add auto restart if it fails, improve logging... Feel free to suggest, both of you, ideas, improvements or whatever you think it's best. But for know we only needed something quick to easily check if it was down (even if it was not very elegant or beautiful :p) and didn't want to interfere much in the rest of the tasks.

PR #873, opened 2022-03-01
fix(npm-scripts): update theme-tasks dependency

bryceosterhaus:


kresimir-coko (approved):
Simple enough, LNHTM 👍

PR #875, opened 2022-03-03
feat(npm-scripts): support bridged exports

izaera:
See commits for a full explanation on what this PR does.

This is to be tested against https://github.com/liferay-frontend/liferay-portal/pull/1955 as per https://github.com/liferay-frontend/liferay-portal/pull/1955#issuecomment-1057983985.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/utils/runWebpackAsBundler.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/runWebpackAsBundler.js
@@ -3,10 +3,14 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
+/* eslint-disable @liferay/no-dynamic-require */
+
 const path = require('path');
+const resolve = require('resolve');
 const TerserPlugin = require('terser-webpack-plugin');
 
 const convertImportsToExternals = require('./convertImportsToExternals');
+const createTempFile = require('./createTempFile');
 const flattenPkgName = require('./flattenPkgName');
 const runWebpack = require('./runWebpack');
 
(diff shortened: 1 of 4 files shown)


izaera:
Maybe `bridge` is not the best name for the configuration parameter, given how many bridges we already have :thinking: 

I think I will change it to something different...

izaera:
> I think I will change it to something different...

The problem is that I can't come up with a meaningful name... Things that come up:

- normalize
- esmify
- reExport
- mangleExports
- normalizeExports
- ....

Suggestions?

izaera:
I think I'll go with `rewriteExports` to merge and release, then we can change it in a following PR...

izaera:
Going to merge this to release a new version

projects/npm-tools/packages/npm-scripts/src/utils/runWebpackAsBundler.js
```diff
@@ -3,82 +3,151 @@
...
+
+function getWebpackConfigs(config) {
+	const {exports, imports} = config;
+
+	return exports.reduce((webpackConfigs, exportsItem) => {
+		const {flatPkgName, importPath, pkgName} = getEntryImportDescriptor(
+			exportsItem
+		);
+
+		const externals = convertImportsToExternals(imports);
```
bryceosterhaus, line 81: Should this be hoisted outside of the reduce and inside of getWebpackConfigs? It doesn't depend on anything in this closure and the `delete` below made me think we want to mutate an object repeatedly and not just once.

bryceosterhaus: I could easily be wrong here, but it just stuck out to me

izaera: > the delete below made me think we want to mutate an object repeatedly and not just once

Nope, it's the opposite. We want a fresh new object in each iteration. Then we delete just one property from that object.

But since I made the same reasoning as you and then had to undo the change locally to make it work, I'll rewrite it hoisting the `convertImportsToExternals`, then explicitly duplicating the object before `delete` :+1: .

PR #883, opened 2022-03-16
feat(npm-scripts): webpack based build

izaera:
This is a draft PR to be tested against a [liferay-portal PR](https://github.com/liferay-frontend/liferay-portal/pull/2013) to demonstrate the webapck based build.


projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -134,15 +149,75 @@ module.exports = async function (...args) {
...
+
+			fs.writeFileSync(
+				path.join(BUILD_CONFIG.output, 'manifest.json'),
+				JSON.stringify(manifest, null, '\t'),
+				'utf8'
 			);
 		}
 		else if (BUILD_CONFIG.exports) {
+
+			// TODO: remove this
```
bryceosterhaus, line 219: nit: Maybe we can add a bit more of a description here so that we don't get confused later.

projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdIndexBridge.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {object} buildConfig
+ */
+function createEsm2AmdIndexBridge(projectDir, buildConfig, manifest) {
+	const pkgJson = require(path.join(projectDir, 'package.json'));
+
+	const {output} = buildConfig;
+
+	const webContextPath = getBndWebContextPath() || `/${pkgJson.name}`;
+
+	const rootDir = pkgJson.name.startsWith('@') ? '../../../..' : '../../..';
```
bryceosterhaus, line 26: `'../../../..' : '../../..'` seems like a sort of "magic" destination. What do these two values reference?

bryceosterhaus (commented):
Looks pretty good, I had a few thoughts but nothing too critical. Overall it seems to work pretty well as I tested it locally though.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: Well, my plan is to do a full rewrite of the commits and a refactor once everything is working, because everything is a bit suboptimal as it is now... :sweat_smile:

projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdIndexBridge.js
izaera: I will add a comment there because it's real magic. Those are the relative positions of each bridge related to the real ES module.

It is so strange because AMD modules are server under things like `/o/js/resolved-module/my-bundle$my-package@1.2.3/path/to/module.js` whereas ESM are under `/o/my-context-path/__liferay__/export/my-package.js`.

Also, depending on whether a package is npm-scoped or not, it needs to add/remove a `..`.

projects/npm-tools/packages/npm-scripts/src/scripts/build.js
izaera: I've enhanced the comment in a new push force.

izaera:
Merging this as it was already reviewed, nothing has substantially changed, and it's green.

PR #893, opened 2022-03-24
fix(portal-base): remove logic that checks for typescript

bryceosterhaus:
Removing these bits as they didn't actually do anything for us.

We added it here, https://github.com/liferay/liferay-frontend-projects/commit/85b566505d4d0da05490857841b5f5c9082094a6, but I'm not sure we had a use case for adding.


izaera:
LGTM

bryceosterhaus:
Checked and verified qa scripts all work as well. merging

PR #894, opened 2022-03-24
feat(npm-scripts): support .scss files

izaera:
We need this to be able to migrate frontend-taglib-clay, for example. It includes source like:

```javascript
import from './perico.scss';
```

Which was previously handled by the bundler. Now we support it with a webpack loader.


izaera:
Shouldn't this trigger the CI? :thinking:

bryceosterhaus:
I think github actions took a quick nap

izaera:
So, it's green. Should I merge?

bryceosterhaus:
It looks good to me, feel free to merge

Issue #895, opened 2022-03-24
v3.x not able to support 0 or >1 js chunks for Create React App adaptations


kgkars:
### Issue type

-   [ ] :thinking: Question
-   [x] :bug: Bug report
-   [ ] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version

-   [ ] :two: v2.x
-   [x] :three: v3.x

### Description

**Desired behavior:**
If you bootstrap a React application with the current [Create React App](https://create-react-app.dev/) version - v5.x - then the `liferay build` command should be able to support the Webpack configurations native to CRA. With migration from CRA and React Scripts to version 5 (along with bumping Webpack from v4 to v5 under the hood), the basic configurations have changed slightly, and it is now more likely to create builds with 0 to n .chunk.js files in the build process.  (You can see others have run into this noted in [CRA Issue #11278](https://github.com/facebook/create-react-app/discussions/11278#discussioncomment-1808511) and [CRA issue #11963](https://github.com/facebook/create-react-app/issues/11963), and there have been [attempts](https://github.com/facebook/create-react-app/pull/11763) to [fix](https://github.com/facebook/create-react-app/pull/11966) update, but no successful pull  requests yet.

**Current behavior:**
In the current **Liferay JS Tookit v3** build process for projects adapted from **Create React App** bases, the build script assumes there will always be 1 .chunk.js file inside the /static/js/ directory.
- If there are no .chunk.js files, the build script will fail without finishing.
- If there are more than one .chunk.js files, then the build script will finish, but will only include the first .chunk.js file in the .jar, causing 404 errors when attempting to load the React Application.

**Repro instructions (if applicable):**
I cannot include my code or package.json, as it is related to an enterprise project, however you can recreate parts of the issues by running the following steps:
1. Bootstrap a CRA app:
```
npx create-react-app my-reproduction
```
2. Inside the application, create at least two lazy loaded components, following the example set out in the [code splitting guide](https://reactjs.org/docs/code-splitting.html).
3. Convert the application to a Liferay module.
```
liferay adapt
```
4. When you run `liferay build` or `npm run build:liferay`, compare the number of files noted in the `build.liferay/work/static/js/` directory to those listed in the `build.liferay/generated/index.js` file and you will notice it is only listing the first .chunk.js file and the main.js file for imports (lines 13-14).

**Other information (environment, versions etc):**
- react v17.0.2
- typescript v4.6.2
- react-scripts v5.0.0
[...]

PR #897, opened 2022-03-25
fix: allow varied number of CRA chunks (#895)

kgkars:
Related to [issue #895](https://github.com/liferay/liferay-frontend-projects/issues/895) which I opened yesterday.

- Given the updated/current **Create React App** Webpack configurations, there may be 0 to n *.chunk.js* files created for a build; however, current Liferay bundle process assumes there will always be 1, and only 1, *.chunk.js* file to include.
- Added code to first check if there are no chunks, then, if chunks exist, turn them into an Array and dynamically import them all into the `index.js` file to be included in the .jar.


izaera:
Going to merge and check :+1: 

Thanks again for the contribution @kgkars

izaera:
Apparently the latest `@vue/cli-service` from the `4.x` version series builds things in a new dir :tada: :

image

So I'll need to add some logic on top of the merged PR to make it safe to run.

izaera:
@kgkars Can you confirm which version you used to test the PR?

My plan is to add an if before listing the chunks so that if they are in `dist` instead of `build` I look for them in that folder. I hope that'll be enough :crossed_fingers: .

izaera:
Ah, sorry. I thought your fix was for Vue because Vue is failing now, but now I've realized yours is for React. 

Never mind about my request, then :+1:

PR #905, opened 2022-04-05
fix(npm-scripts): update path to node_modules in prettier script

bryceosterhaus:
Looks like we had created a bug in a previous fix when we did some source formatting.

cc @kresimir-coko can you verify this works as expected for both workspaces and liferay-portal?


kresimir-coko:
Sure @bryceosterhaus I'll try to squeeze it in today

kresimir-coko:
Hey @bryceosterhaus I've tested this in both liferay-portal and my preexisting workspace and it works fine. Not sure what the bug was, but everything is fine on my end with your changes.

What I did was copy the contents of the prettier.js file to my local one that the extension points to, I also copied over `@liferay/npm-scripts` to `node_module` just in case and changed single quotation marks into doubles and hit save, upon which it formatted them back to single quotes.

PR #909, opened 2022-04-07
feat: bring ESM support to AMD loader

izaera:
This adds a new module flag named `useESM` that, when interpreted by the AMD loader, lets JS scripts be loaded with `<script type="module">`.

It is needed to fix [this error](https://github.com/liferay-frontend/liferay-portal/pull/2080#issuecomment-1090195656).


bryceosterhaus (approved):
LGTM!

PR #915, opened 2022-04-13
fix(npm-scripts): honor NODE_ENV setting when invoking webpack

izaera:


izaera:
Oops, lint failed. We should find a way to make our selection processes better... :roll_eyes:

bryceosterhaus:
LGTM

>  We should find a way to make our selection processes better

Not entirely sure what you mean by "selection processes" here

izaera:
> Not entirely sure what you mean by "selection processes" here

The process of selecting who enters (in this case me) the Company and who doesn't... :smiling_face_with_tear:

PR #916, opened 2022-04-13
feat(npm-scripts): add customBridges support

izaera:
Custom bridges are (hopefully) temporary bridges from ESM to AMD that are
needed by legacy pre-AMD code to work.

So, once we remove the pre-AMD code and the AMD code, I guess we can remove
these...


bryceosterhaus (approved):
LGTM

PR #921, opened 2022-04-25
fix(npm-scripts): create output dir for custom bridges

izaera:


izaera:
Going to merge this so that I can release a new version

PR #928, opened 2022-05-11
fix(npm-scripts): Updates getPathContext to DXP's default value

diegonvs:
When running Liferay.ThemeDisplay.getPathContext() inside a DXP instance where we don't change pathContext, it returns an empty string instead of a slash :)

Context: https://github.com/liferay-frontend/liferay-portal/pull/2190#issuecomment-1124230621
Fixes https://github.com/liferay/liferay-frontend-projects/issues/929


bryceosterhaus:
LGTM!

PR #931, opened 2022-05-13
LPS-153647 Remove rules that start with *

georgel-pop-lr:
LPS-153647 Remove rules that start with *, these rules are fixes for IE7 that we don't support anymore, and it should be removed as we have clients that complain about their accessibility tests failing cause of it

### DIFF
--- a/third-party/projects/alloy-ui/src/aui-button-item-deprecated/assets/aui-button-item-deprecated-core.css
+++ b/third-party/projects/alloy-ui/src/aui-button-item-deprecated/assets/aui-button-item-deprecated-core.css
@@ -6,22 +6,15 @@
     white-space: nowrap;
     width: auto;
     overflow: visible;
-    *padding: 2px 5px 2px 4px;
-    *width: 1;
 }
 .buttonitem-icon {
     display: inline-block;
     margin-top: -3px;
     vertical-align: middle;
-    /* IE6/7 */
-    *text-indent: 0;
-    *margin-top: 1px;
 }
 .buttonitem-label {
     line-height: 1em;
     display: inline-block;
-    /* IE6/7 */
-    *line-height: 1.4em;
 }
 .buttonitem-label {
     padding: 0 5px;
@@ -35,15 +28,9 @@
     box-sizing: content-box;
     -moz-box-sizing: content-box;
     -webkit-box-sizing: content-box;
-    /* IE6/7 */
-    *height: 24px;
-    *width: 24px;
 }
 .buttonitem-icon-only .buttonitem-icon {
     margin-top: 0;
-    /* IE6/7 */
-    *margin-top: -1px;
-    *margin-left: -1px;
 }
 .gecko .buttonitem-icon-only .buttonitem-icon {
     margin-top: -3px;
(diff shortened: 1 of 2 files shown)


bryceosterhaus:
@georgel-pop-lr are customers able to just ignore those these sort of rules? I am just slightly concerned about making changes to a deprecated package so that a customer can have their tests pass.

cc @dsanz, not sure if you know of any support-type rules we would have for this, I don't know if we have certain criteria of when we will update deprecated packages.

georgel-pop-lr:
> @georgel-pop-lr are customers able to just ignore those these sort of rules? I am just slightly concerned about making changes to a deprecated package so that a customer can have their tests pass.
> 
> cc @dsanz, not sure if you know of any support-type rules we would have for this, I don't know if we have certain criteria of when we will update deprecated packages.

Hi @bryceosterhaus,

I'm afraid the customer would want this fixed, as we already did in some other libraries and portal. The problem is that the tool they use for accessibility would not be able to process this CSS, so the tests would not be reliable.

Regards.

dsanz:
Hi @bryceosterhaus @georgel-pop-lr 

Although it looks like this looks to affect to IE7 only, I'd like to make sure that IE11 users, which we still support in maintenance branches, are not affected.

In that case, it should be safe to remove these, particularly if they're causing problems to current a11y checkers. This stance is supported by the fact that we should not consider any IE version below 11 ([browsers policy](https://help.liferay.com/hc/es/articles/360018305911)). 

> I don't know if we have certain criteria of when we will update deprecated packages.
Afaik there's no fixed rule for this. I'm supporting the idea of updating just because, outside of IE7 universe, we're serving wrong CSS

georgel-pop-lr:
> Hi @bryceosterhaus @georgel-pop-lr
> 
> Although it looks like this looks to affect to IE7 only, I'd like to make sure that IE11 users, which we still support in maintenance branches, are not affected.
> 
> In that case, it should be safe to remove these, particularly if they're causing problems to current a11y checkers. This stance is supported by the fact that we should not consider any IE version below 11 ([browsers policy](https://help.liferay.com/hc/es/articles/360018305911)).
> 
> > I don't know if we have certain criteria of when we will update deprecated packages.
> 
> Afaik there's no fixed rule for this. I'm supporting the idea of updating just because, outside of IE7 universe, keeping this means we're serving wrong CSS

Thanks, @dsanz, for checking this.

@bryceosterhaus please let me know your thoughts on this and if we can merge it.

Like this, I can give an answer to the client.

Regards.

bryceosterhaus:
@georgel-pop-lr I will look at adding and this and releasing today. Just gonna re-run the windows CI and see if that still fails

bryceosterhaus:
Hmm, this looks to continue to fail on windows.

georgel-pop-lr:
> Hmm, this looks to continue to fail on windows.

Hmm, I remember that I saw windows failing even on this PR https://github.com/liferay/liferay-frontend-projects/pull/937 opened yesterday:
image

Are we sure if not a generic issue that affects all the pulls and not just this one?

Regards.

bryceosterhaus:
Yep, you're right, looks like thats on all PRs. I will investigate

bryceosterhaus:
Released at https://repository.liferay.com/nexus/content/repositories/liferay-public-releases/com/liferay/webjars/com.liferay.webjars.alloy-ui/3.1.0-deprecated.93/

georgel-pop-lr:
> Released at https://repository.liferay.com/nexus/content/repositories/liferay-public-releases/com/liferay/webjars/com.liferay.webjars.alloy-ui/3.1.0-deprecated.93/

Great, thanks @bryceosterhaus.

Would you release this version into master?

bryceosterhaus:
@georgel-pop-lr https://github.com/liferay-frontend/liferay-portal/pull/2254

PR #941, opened 2022-05-25
LPS-153647 Remove IE6 IE7 IE8 hack classes no longer needed

georgel-pop-lr:
I finally cleaned all the IE rules, I guess we should not keep them anymore as we don't support this browsers

I think we should first merge https://github.com/liferay/yui3/pull/26 and then merge this as well and release a new version including the yui3 changes


bryceosterhaus:
@georgel-pop-lr will do! I am right in the middle of something but I will release both yui and aui today

PR #943, opened 2022-05-25
fix(npm-scripts): adds missing FeatureFlags mock

diegonvs:
I [needed ](https://github.com/liferay-frontend/liferay-portal/pull/2256/commits/1ba6e0a40e2ca1666fcc2892ed6fab2f1e7f66f8)to place a mock for `Liferay.FeatureFlags,` and apparently, other PTs are doing the same.

See https://cs.github.com/liferay/liferay-portal?q=%20%22FeatureFlags%22%20language%3AJavaScript.


projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
```diff
@@ -134,6 +134,11 @@ const ThemeDisplay = {
  */
 const Util = {
 
+	/**
+	 * https://github.com/liferay/liferay-portal/blob/efc53b8a82e3afff06aacc26118fbe7231acffff/portal-web/docroot/html/common/themes/top_js.jspf#L157
+	 */
+	FeatureFlags: {},
```
bryceosterhaus, line 140: I think this would be at the root level right and not Util, right? Since it is `Liferay.FeatureFlags` and not `Liferay.Util.FeatureFlags`

bryceosterhaus (changes requested):

diegonvs:
Updated! :)

PR #948, opened 2022-06-09
feat(npm-scripts): add more default values for our mocks in DXP

kresimir-coko:
This PR aims to move pretty much all of the local Jest mocks (found inside whichever file is declared as the host of Jest configs) into a global place inside of npm-scripts.

I forget how exactly we do versions for npm-scripts (or any project here), but we can bump the version of this and then update `package.json` to the new version inside the counterpart PR in `liferay-portal`. 

This PR doesn't do anything, and it will probably cause issues, if it makes it into `liferay-portal` without its counterpart.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
+++ b/projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
@@ -5,7 +5,33 @@
 
 /* eslint-env jest */
 
+const languageMap = {
+	'days-abbreviation': 'd',
+	'decimal-delimiter': '.',
+	'hours-abbreviation': 'h',
+	'minutes-abbreviation': 'min',
+	'mmm-dd': 'MMM DD',
+	'mmm-dd-hh-mm': 'MMM DD, HH:mm',
+	'mmm-dd-hh-mm-a': 'MMM DD, hh:mm A',
+	'mmm-dd-lt': 'MMM DD, LT',
+	'mmm-dd-yyyy': 'MMM DD, YYYY',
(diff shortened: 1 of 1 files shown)


projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
```diff
@@ -36,6 +67,18 @@ const events = {
 	once: jest.fn(),
 };
 
+const namespace = jest.fn((name) => {
+	global.Liferay[name] = {};
+
+	return global.Liferay[name];
+});
```
bryceosterhaus, line 74: Where does this come from? I'm not super familiar with this one

bryceosterhaus (commented):
Looks good! just had one question

bryceosterhaus:
Also, can you start preparing the counterpart PR for dxp? Easiest way to test is by using yarn link for npm-scripts

kresimir-coko:
> Also, can you start preparing the counterpart PR for dxp? Easiest way to test is by using yarn link for npm-scripts

I have it almost ready, gonna just double check again before opening a PR, that's what I was doing the last week or so 😄 
I've been testing it by copy-pasting the `Liferay.js` file and that seems to have done the trick.

projects/npm-tools/packages/npm-scripts/src/jest/mocks/Liferay.js
kresimir-coko: It's used in here https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/portal-search/portal-search-web/test/setup.js#L64-L70

PR #949, opened 2022-06-10
feat(eslint-plugin): handful of additions

### LINKED ISSUE #13: Lint for misuse of useIsMounted hook
As seen [here](https://github.com/wincent/liferay-portal/pull/169#discussion_r389640310), it is pretty easy to write:

```js
if (isMounted) {
    // ...
}
```

instead of:

```js
if (isMounted()) {
    // ...
}
```

so we should make a lint to catch the former whenever the return value of `useIsMounted()` from `frontend-js-react-web` is used in a conditional context in this way.

### LINKED ISSUE #26: Enforce use of const/let instead of var
As per this comment: https://github.com/wincent/liferay-portal/pull/319#discussion_r436915980

### LINKED ISSUE #719: Warn when a module has no main entry in package.json

-   [ ] :thinking: Question
-   [ ] :bug: Bug report

### Description
Currently, when a portal module is lacking the `main` entry in its `package.json` file, it makes tests fails for other modules that need to import it.

This is because [getJestModuleNameMapper](https://github.com/liferay/liferay-frontend-projects/blob/f35a9b3a6c17498a2d56b9892f99e6f742885011/projects/npm-tools/packages/npm-scripts/src/utils/getJestModuleNameMapper.js) is ignoring modules without that `main` entry.

We cannot fix the `getJestModuleNameMapper` because there's nothing to fix. Also, we cannot show a warning message because we would have lots of false positives.

However, the situation is quite unfortunate, because it's very difficult to diagnose the condition and/or realize that you are being hit by it when it happens.

So I propose creating a linting rule that applies heuristics to warn when a `main` entry is missing and it appears it should be informed.

**Desired behavior:**
When a module has no `main` entry but it has an `index.js`, `index.es.js`, `index.ts`, `js/index.js`, `js/index.es.js` or `js/index.ts` file in `src/main/resources/META-INF/resources" issue a warning.

We may want to add more "standard" main files if we see any other fit.

And I'm not sure if we should provide a way to opt out... I find it very unlikely that someone has an `index.js` file but doesn't want it as `main`, but if we find that, me may need to provide some flag to disable the rule, I guess...

See https://liferay.slack.com/archives/C3JBR21HA/p1635337119258800

### LINKED ISSUE #918: feat(eslint-plugin): Add a rule that adds squiggly brackets around imports from `frontend-js-web`
After confirming with @izaera we discovered that it would be useful to have an eslint rule to make sure that all imports from `frontend-js-web` are wrapped inside squiggly brackets `{}`. 

The image below doesn't have any warnings and after running source formatter it doesn't get fixed, we should add that in, so that devs don't make a potentially very confusing mistake. 

image

P.S. `gradlew deploy -a` doesn't produce an error either, so we might be preventing some huge issues by adding this rule.

bryceosterhaus:
Thought I would tackle a few of these while I was cleaning up old eslint issues.

fixes #719
fixes #13
fixes #918
fixes #26

I additionally added a preflight check for npm-scripts to check package.json's for a main entry point.


projects/eslint-plugin/rules/portal/tests/lib/rules/no-default-export-from-frontend-js-web.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+const MultiTester = require('../../../../../scripts/MultiTester');
+const rule = require('../../../lib/rules/no-default-export-from-frontend-js-web');
+
+const parserOptions = {
+	ecmaVersion: 6,
+	sourceType: 'module',
+};
+const ruleTester = new MultiTester({parserOptions});
```
kresimir-coko, line 13: Empty line between these 2 I'd think

kresimir-coko (approved):
Apart from the tiny SF thingy it looks great!

PR #952, opened 2022-06-14
chore: update CI checker for conventional commits, semantic pr bot is dead

bryceosterhaus:


izaera:
LGTM  :+1: 

If it works, I have nothing to say :sweat_smile: ...

bryceosterhaus:
It doesn't check against the actual commits, only the PR title, but I think I'll merge this for at least some support and then I can create something else later that checks commit messages

PR #954, opened 2022-06-16
fix(npm-scripts): add aspect-ratio as known property to stylelint

izaera:
I'm adding the `aspect-ratio` property to the list of known properties here, but I'm not sure if maybe upgrading stylelint would fix it too :thinking: 

Also, I haven't tested it, I wanted to first check if this would be the real solution.

yolo


bryceosterhaus:
Screen Shot 2022-06-17 at 2 00 36 PM

PR #960, opened 2022-06-21
fix(alloy-ui): LPS-153254 Render popover inside input parent node for…

georgel-pop-lr:
… better accessibility

### DIFF
--- a/third-party/projects/alloy-ui/src/aui-datepicker/js/aui-datepicker-popover.js
+++ b/third-party/projects/alloy-ui/src/aui-datepicker/js/aui-datepicker-popover.js
@@ -200,7 +200,6 @@ A.mix(DatePickerPopover.prototype, {
                 }
             ],
             position: 'bottom',
-            render: true,
             triggerShowEvent: 'click',
             triggerToggleEvent: null,
             visible: false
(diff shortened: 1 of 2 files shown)


georgel-pop-lr:
Hi @bryceosterhaus, 

The client is waiting for this fix, could we release a new version?

Regards.

bryceosterhaus:
@georgel-pop-lr yeah I can release a new version today.

Would it be helpful if we let you make the releases, that way you don't have to wait on us? I already assume you probably have a much better idea of how AUI is working, especially in the context of portal.

georgel-pop-lr:
> @georgel-pop-lr yeah I can release a new version today.
> 
> Would it be helpful if we let you make the releases, that way you don't have to wait on us? I already assume you probably have a much better idea of how AUI is working, especially in the context of portal.

Hi @bryceosterhaus,

Thanks for the confidence on giving me this permission, but anyhow, I would prefer if still one person does this releases in order to not mess up something.

Also, I prefer the team is checking this pulls in case I miss something in the code, always two eyes are better than one.

Let me know when the new version is released.

Regards.

bryceosterhaus:
@georgel-pop-lr here is the published version https://repository.liferay.com/nexus/content/repositories/liferay-public-releases/com/liferay/webjars/com.liferay.webjars.alloy-ui/3.1.0-deprecated.96/

Issue #516, opened 2021-04-28
Show documentation on what to expect about dependency version numbers


izaera:
### Issue type

-   :gift: Feature request

### Version

-   :nine: v9.x
-   :keycap_ten: v10.x or higher

### Description

Add a notice regarding this documentation https://github.com/liferay/clay/wiki/Liferay-Portal-@clayui-css-Versions to the end of the generation process so that users know that the projects being generated are simply templates but deps need to be maintained by the developer.

This issue was created after #513 and #514 were discussed in a [slack conversation](https://liferay.slack.com/archives/C3JBR21HA/p1619618333416500).

izaera:
This is the same as https://issues.liferay.com/browse/IFI-2293

Issue #513, opened 2021-04-27
npm package for  liferay-frontend-theme-* needs to be updated


fortunatomaldonado:
### Issue type (mark with `x`)

-   [ ] :thinking: Question
-   [x] :bug: Bug report
-   [ ] :gift: Feature request
-   [ ] :woman_shrugging: Other

### Version (mark with `x`)

-   [ ] :eight: v8.x
-   [x] :nine: v9.x
-   [x] :keycap_ten: v10.x or higher

### Description

**Desired behavior:**
When using 'yo liferay-theme' for 7.3.x, package.json should have the updated following theme package:
"liferay-frontend-theme-styled": "5.0.15",
"liferay-frontend-theme-unstyled": "5.0.20,

**Current behavior:**
When using 'yo liferay-theme' for 7.3.x, package.json has the following:
"liferay-frontend-theme-styled": "5.0.0",
"liferay-frontend-theme-unstyled": "5.0.0"

This results in issues in portal_normal.ftl files

**Repro instructions (if applicable):**

**Other information (environment, versions etc):**

jonmak08:
Adding https://issues.liferay.com/browse/IFI-2450 for reference.

izaera:
Mmm, we think this issue is incorrect per [this table](https://github.com/liferay/clay/wiki/Liferay-Portal-@clayui-css-Versions)

DXP 7.3 uses 5.x versions for those deps...

izaera:
The themes toolkit generator creates projects based on a template that must work for all 7.3 versions of DXP so we should restrict it to the lower version, then it's up to the developer to upgrade his/her project when he needs it.

However, I see that you were targeting 7.3, so I wonder how you came to the conclusion that you needed 6.x versions... :thinking: Can you elaborate more on that taking into account my comments? Thx.

fortunatomaldonado:
@izaera, I have updated the ticket to reflect the conversations and decision that "5.0.15" and "5.0.20" are acceptable for the default.
Thank for the info and https://github.com/liferay/liferay-frontend-projects/issues/516 to better understand what is required.
Let me know if there are any questions or comments about this.

izaera:
This should be fixed by now. Please reopen it if it is not the case.

PR #970, opened 2022-07-13
fix(theme-tasks): update liferay-frontend-theme-styled, liferay-front…

georgel-pop-lr:
…end-theme-unstyled and liferay-font-awesome version reference

### DIFF
--- a/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/devDependencies.js
+++ b/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/devDependencies.js
@@ -51,12 +51,12 @@ module.exports = {
 				'compass-mixins': strict('0.12.10'),
 				'gulp': gulpVersion,
 				'liferay-frontend-common-css': strict('1.0.4'),
-				'liferay-frontend-theme-styled': strict('4.0.21'),
-				'liferay-frontend-theme-unstyled': strict('4.0.17'),
+				'liferay-frontend-theme-styled': strict('4.0.42'),
+				'liferay-frontend-theme-unstyled': strict('4.0.46'),
 				'liferay-theme-tasks': themeTasksVersion,
 			},
 			optional: {
-				'liferay-font-awesome': strict('3.5.1'),
+				'liferay-font-awesome': strict('3.5.2'),
 			},
 		},
 		7.3: {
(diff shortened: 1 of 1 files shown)


izaera:
Didn't we agree time ago to not update these version numbers? :thinking: 

See this comment, for clay -> https://github.com/liferay/liferay-frontend-projects/issues/513#issuecomment-828502826

And this -> https://github.com/liferay/liferay-frontend-projects/issues/516 for the full context

@bryceosterhaus Any thoughts?

bryceosterhaus:
@izaera I also remember deciding not to update these versions.

@georgel-pop-lr what is the reason for updating them? Is there a specific LPS or something that is depending on this?


I really don't have many opinions with the theme stuff so at the end of the day, if it is absolutely necessary then I have no problem updating them.

georgel-pop-lr:
> @izaera I also remember deciding not to update these versions.
> 
> @georgel-pop-lr what is the reason for updating them? Is there a specific LPS or something that is depending on this?
> 
> I really don't have many opinions with the theme stuff so at the end of the day, if it is absolutely necessary then I have no problem updating them.

Hi @bryceosterhaus @izaera,

We have made lots of accessibility fixes in these versions, and I see many clients opening tickets and needing to ask them to upgrade the themes, so I guess it is better if our clients starts with a newer version, and maybe we have fewer tickets of this type.

Regards.

izaera:
> We have made lots of accessibility fixes in these versions, and I see many clients opening tickets and needing to ask them to upgrade the themes, so I guess it is better if our clients starts with a newer version, and maybe we have fewer tickets of this type.

Apparently, that's aligned with the Dave Truong's comment I linked above...

Whatever is better. If updating the numbers leads to less support load, then let's update them. In the end it's the user responsibility to put the correct number so we can provide the defaults that lead to less friction.

izaera (approved):

bryceosterhaus:
sounds good to me! thanks for the explanation @georgel-pop-lr

bryceosterhaus:
I'll release a new version first thing in the morning

bryceosterhaus:
Screen Shot 2022-07-14 at 3 47 24 PM


is github stalking me? 😨

bryceosterhaus:
released https://github.com/liferay/liferay-frontend-projects/releases/tag/liferay-theme-tasks%2Fv11.4.0

PR #990, opened 2022-08-19
feat: support live development and deployment for Remote App projects

izaera:
See https://issues.liferay.com/browse/LPS-160803


izaera:
I've fixed the QA scripts, but I'm not sure if the `yarn qa check` will work because it tries to compare the output ZIP files by hashing them but, as file times will change, it will probably fail in future runs. 

So I'll add a commit to check the internals of the ZIP file file by file.

izaera:
I've added support for diffing the internals of ZIP files in the QA scripts. This is ready for merge now.

izaera:
Converted to draft again because I want to add a `--only` switch to the `start` command so that the live development server can be launched without automatic deployment to liferay being executed.

projects/js-toolkit/packages/portal-base/src/build/bundler2.ts
```diff
@@ -106,7 +106,14 @@ function runBabel(project: Project): void {
 }
 
 function runCompiler(project: Project): void {
-	runBabel(project);
+	const dependencies = project.pkgJson.dependencies || {};
+
+	if (dependencies['@angular/core']) {
```
bryceosterhaus, line 111: 👍

projects/js-toolkit/packages/portal-base/src/build/customElement.ts
```diff
@@ -25,6 +34,42 @@ export default async function customElement(project: Project): Promise<void> {
...
 	await runWebpack(project, configuration);
+
+	await makeZip(project);
+}
+
+/**
+ * Add several files to a ZIP folder.
+ * @param srcDirPath source folder
+ * @param srcGlobs array of globs describing files to include (in
+ * 						globby, i.e. POSIX, format)
```
bryceosterhaus, line 45: whitespace looks a little odd here

projects/js-toolkit/packages/portal-base/src/start/customElement.ts
```diff
@@ -37,4 +54,108 @@ export default async function customElement(project: Project): Promise<void> {
...
+}
+
+async function checkForOldZipFile(zipFile: FilePath): Promise<void> {
+	if (!fs.existsSync(zipFile.asNative)) {
+		return;
+	}
+
+	const stats = fs.statSync(zipFile.asNative);
+
+	if (Date.now() - stats.mtimeMs > 1000 * 60 * 120) {
```
bryceosterhaus, line 76: Can we set `1000 * 60 * 120` to a variable with a explicit name?

projects/js-toolkit/packages/portal-base/src/start/customElement.ts
```diff
@@ -37,4 +54,108 @@ export default async function customElement(project: Project): Promise<void> {
...
+		const lines = createInterface({
+			input: process.stdin,
+		});
+
+		print(warn`
+There is an older than 2 hours version of this project in your Liferay server.`);
```
bryceosterhaus, line 82: Why the 2 hour limit? Was this just arbitrary?

projects/js-toolkit/packages/portal-base/src/util/promptForDeployPath.ts
```diff
@@ -0,0 +1,40 @@
...
+): Promise<FilePath> {
+	const lines = createInterface({
+		input: process.stdin,
+	});
+
+	print(question`Please enter your local Liferay installation directory`);
+
+	let deployDir: FilePath;
+
+	for await (const line of lines) {
```
bryceosterhaus, line 23: Whoa, haven't seen `for await` before, very cool

bryceosterhaus (approved):
LGTM! I had a few small comments but looks fine by me.

projects/js-toolkit/packages/portal-base/src/start/customElement.ts
izaera: Yes. It seems reasonable to me. My intention is to warn only in cases where you may have deployed the application some time ago, in a previous development session, but not every time you run `yarn start`. Two hours seems enough to not trigger false positives when you stop for lunch :sweat_smile: 

However, I'm now thinking that people won't be running `yarn start` a lot of times. Chances are that they just run it once per development session, so maybe we can warn always (without checking for the elapsed time)... :thinking:

projects/js-toolkit/packages/portal-base/src/util/promptForDeployPath.ts
izaera: you don't know about it until you need it :sweat_smile:

projects/js-toolkit/packages/portal-base/src/start/customElement.ts
bryceosterhaus: Sounds good, just wanted to make sure I fully understood!

projects/js-toolkit/packages/portal-base/src/start/customElement.ts
izaera: for example, something like:

```javascript
const ONE_WEEK = 1000 * 60 * 120;
```

:trollface:

projects/js-toolkit/packages/portal-base/src/start/customElement.ts
izaera: I'm going to remove the check for the two hours...

PR #996, opened 2022-08-29
fix(target-platforms): fix target platforms for DXP 7.3

### LINKED ISSUE #994: The create-platform script is not taking bundler excludes into account

-   [ ] :thinking: Question
-   [X] :bug: Bug report

### Version (mark with `x`)

-   [ ] :two: v2.x
-   [X] :three: v3.x

### Description

Example, this target platform -> https://unpkg.com/@liferay/dxp-7.3@2.0.0/config.json
Comes from https://github.com/liferay/liferay-portal/blob/7.3.1-ga2/modules/apps/frontend-js/frontend-js-web/package.json
But all packages are excluded from the bundler, so they should not be exported -> https://github.com/liferay/liferay-portal/blob/7.3.1-ga2/modules/apps/frontend-js/frontend-js-web/.npmbundlerrc#L3

See https://liferay.slack.com/archives/C03CAD9EF/p1661513286307359

izaera:
Fixes #994


target-platforms/packages/create-platform/lib/getBundlerImports.js
```diff
@@ -57,3 +119,29 @@ module.exports = async function getBundlerImports(tagOrDir, isEE) {
 
 	return imports;
 };
+
+function saveSummary(platformName, summary) {
+	console.log(platformName);
```
bryceosterhaus, line 124: This meant to be here?

target-platforms/packages/dxp-7.3/summary.json
```diff
@@ -0,0 +1,126 @@
...
+	"@clayui/shared": ["frontend-js-spa-web"],
+	"billboard.js": ["frontend-taglib-clay"],
+	"buffer": ["portal-workflow-metrics-web"],
+	"classnames": ["frontend-js-spa-web"],
+	"clay": ["frontend-js-spa-web"],
+	"clay-alert": ["frontend-js-spa-web"],
+	"clay-autocomplete": ["document-library-web"],
+	"clay-button": ["document-library-web", "frontend-js-spa-web"],
+	"clay-component": ["frontend-js-spa-web"],
+	"clay-data-provider": ["document-library-web"],
```
bryceosterhaus, line 17: I don't fully remember what this file is for, but it seems odd that document-library would be the primary reference to clay-data-provider

target-platforms/packages/create-platform/lib/getBundlerImports.js
izaera: nope :+1:

target-platforms/packages/dxp-7.3/summary.json
izaera: No, that's a new file to list what projects import each package, because it will help to diagnose what could be wrong in the future (saves us a complex search in the liferay-portal project to see what could be the issue).

izaera: So, in your case, `document-library-web` is bundler-importing `clay-data-provider`. That may make sense or not depending on how well if `liferay-portal` configured which, for version 7.3 or lower, is not quite good, because bundler-imports were distributed among all .npmbundlerrc files and that caused some of them to be incorrect or obsolete.

Sadly, we don't have a better way to infer the target platform, so we'll have to live with that.

PR #998, opened 2022-09-01
fix(js-toolkit-core): fix "No such file or directory .liferay.json" error

izaera:


izaera:
Closed because it was assigned to a wrong issue number

PR #1002, opened 2022-09-15
fix: hard-coded css/main.css file in create-react-app adaptations

### LINKED ISSUE #1000: Why is `css/main.css` hard-coded into the `create-react-app` portal?
## What is your question?

Hello all, I just have a quick question that is based on a small problem we are having with a React app (made with `create-react-app`) compiled to a Liferay portal using `@liferay/cli`.

I see that `/static/css/main.css` is being added to the application through `com.liferay.portlet.header-portlet-css` in `addCSSPortletHeader.ts` whether or not the file actually exists or there is an existing config in place already.

I guess my question is why this is happening? And is it necessary to hard-code this dependency in? We are using css-in-js for our application so don't have a global `main.css` and we're getting 404's on the network because there's nothing there and we can't remove this option.

izaera:
Fixes #1000

### DIFF
--- a/.github/workflows/projects.yml
+++ b/.github/workflows/projects.yml
@@ -22,7 +22,7 @@ on:
 
 env:
     CI: true
-    yarn-cache-name: yarn-cache-3
+    yarn-cache-name: yarn-cache-4
     yarn-cache-path: .yarn
 
 jobs:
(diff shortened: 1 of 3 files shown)


bryceosterhaus (approved):
LGTM!

PR #1009, opened 2022-09-26
feat(liferay-npm-bundler): add new configuration "strictGlobalDependencies" which looks at the global bundler config to determine what dependencies to bundle instead of just the module's package.json

bryceosterhaus:
This would allow us to specify all dependencies in our package.json files in DXP without necessarily having to bundle them.

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/dependencies.js
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/dependencies.js
@@ -56,6 +56,8 @@ export function addPackageDependencies(
 	dependencies = Object.keys(dependencies);
 	dependencies = dependencies.concat(extraDependencies);
 
+	dependencies = project.copy.filterDependencies(dependencies);
+
 	const dependencyDirs = dependencies
 		.filter(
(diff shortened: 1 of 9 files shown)


maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/dependencies.js
```diff
@@ -22,6 +22,8 @@ export function getRootPkg() {
 	return rootPkg;
 }
 
+let DEP_MAP;
```
izaera, line 25: why a global var? :thinking:

izaera (approved):

maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/dependencies.js
```diff
@@ -56,6 +56,9 @@ export function addPackageDependencies(
 	dependencies = Object.keys(dependencies);
 	dependencies = dependencies.concat(extraDependencies);
 
+	dependencies = project.copy.filterDependencies(dependencies);
+
```
izaera, line 60: oops, extra LF here

PR #1012, opened 2022-09-28
fix: MVC portlet source maps point to a wrong place

### LINKED ISSUE #1010: MVC portlet source maps point to a wrong place

-   [ ] :thinking: Question
-   [ ] :bug: Bug report

### Version (mark with `x`)

-   [X] :two: v2.x
-   [ ] :three: v3.x

### Description

**Desired behavior:**
We need MVC portlets generated by the Workspace to have the correct source file path in the `.map` files.

**Current behavior:**
Because we are running babel before the bundler in MVC portlets, babel gets `src/main/resources/META-INF/resources` as its input path (see https://github.com/lawrence-lee/react-debugging-example/blob/main/gradle-react-workspace/modules/sample-react/package.json#L16) and writes it to the sourcemap files.

This makes browsers place the mapped and unmapped sources in a different place in the debugger tree.

izaera:
Fixes #1010


bryceosterhaus (approved):
LGTM

PR #1015, opened 2022-09-30
fix: Handle scroll to into view after change height in aui-autosize-iframe.js

chileces:
Hello Team,
This PR is to solve an issue related to aui-autosize-iframe, please help to review this PR.
Issue: https://issues.liferay.com/browse/LPS-163952 The details page is loaded but you only see the bottom part.
PTR: https://issues.liferay.com/browse/PTR-3350
Thank you very much,
Chi

### DIFF
--- a/third-party/projects/alloy-ui/src/aui-autosize/js/aui-autosize-iframe.js
+++ b/third-party/projects/alloy-ui/src/aui-autosize/js/aui-autosize-iframe.js
@@ -78,6 +78,17 @@ var AutosizeIframe = A.Component.create({
             value: true
         },
 
+         /**
+         * Indicates if the scroll into view should be executed
+         *
+         * @attribute scrollIntoView
+         * @default false
+         * @type {Object}
+         */
+          scrollIntoView: {
+            value: false
+        },
+
         /**
          * The width of the iframe.
          *
@@ -126,6 +137,7 @@ var AutosizeIframe = A.Component.create({
             instance.after('heightChange', instance._afterHeightChange);
             instance.after('widthChange', instance._afterWidthChange);
             instance.after('monitorHeightChange', instance._afterMonitorHeightChange);
+            instance.after('scrollIntoViewChange', instance._afterScrollIntoViewChange);
         },
 
         /**
@@ -200,6 +212,19 @@ var AutosizeIframe = A.Component.create({
             var instance = this;
 
             instance._uiSetMonitorHeight(event.newVal);
+        },
+
+        /**
+         * Fires after `scrollIntoView` attribute changes.
+         *
+         * @method _afterScrollIntoViewChange
+         * @param event
+         * @protected
+         */
+         _afterScrollIntoViewChange: function(event) {
+            var instance = this;
+
+            instance._uiSetScrollIntoView(event.newVal);
         },
 
         /**
@@ -309,7 +334,23 @@ var AutosizeIframe = A.Component.create({
                 instance._iframeHeight = value;
 
                 instance.node.setStyle('height', value);
+
+                if (instance._attrs.scrollIntoView){
+                    instance.node.scrollIntoView(instance._attrs.scrollIntoView);
+                }
             }
+        },
+        /**
+         * Sets the iframe scroll into view attribute in the UI.
+         *
+         * @method _uiSetScrollIntoView
+         * @param value
+         * @protected
+         */
+         _uiSetScrollIntoView: function(value) {
+            var instance = this;
+            instance._attrs.scrollIntoView = value;
+            instance.node.scrollIntoView(value)
         },
 
         /**


third-party/projects/alloy-ui/src/aui-autosize/js/aui-autosize-iframe.js
```diff
@@ -309,6 +309,7 @@ var AutosizeIframe = A.Component.create({
                 instance._iframeHeight = value;
 
                 instance.node.setStyle('height', value);
+                instance.node.scrollIntoView(true);
```
dsanz, line 312: This change affects other places where the aui-autosize-iframe is being used, being the embedded layout one of them. Are we 100% sure we want to scrollintoview for all of them? Would not be this better handled via opt-in configuration? @markocikos your thoughts are welcome

chileces: hello @dsanz @markocikos , if i understood you correctly, the term "handled via opt-in" is that we should create a new attribute and let outside control the behavior of the iframe. is that correct?
i put a draft code in my repo for this matter, please help to correct me
https://github.com/chileces/liferay-frontend-projects/commit/193f70bd10547b070d34b33d79790e0e990b523e

markocikos: I like the idea of adding this to API, as a new attribute. Looking at https://github.com/chileces/liferay-frontend-projects/commit/193f70bd10547b070d34b33d79790e0e990b523e, I'd make two changes:
- `true` probably makes sense as the default, I would expect that to be the most common desired behavior. 
- In addition to boolean, I would allow object as an argument, as it is allowed in [the native scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView).

chileces: i pushed the changes according to your comments, please check if it is ok.

markocikos (commented):
The code LGTM, but I did not test for final confirmation this works. It is not reproducible in `master`, only 7.1. @chileces were you able to test changes? If all good, we can merge this.

chileces:
> 

hello @markocikos , dsanz,
I'm unable to build the Liferay frontend project locally and test the integration to the Liferay portal 7.1.x. I just have a test for this file in the developer tools (f12), although this seems to work on my side, it is not the right way. Could you please help to verify it on your side?

bryceosterhaus:
One note in regards to testing, we need to make sure this change is tested in all places where this code will be executed and not just a single instance. Have we identified maybe the top 3-5 places where this file runs and verified that it causes no regressions there? Using dev tools for those places should be sufficient.

chileces:
thank you for your feedback, i will test it

third-party/projects/alloy-ui/src/aui-autosize/js/aui-autosize-iframe.js
dsanz: @chileces yes, that's the idea. This way, DXP client code manages the flag instead of us producing a behavioral change across all usages of the component (note that our customers may be using this in their third party developments too).

dsanz:
These are the usages (master branch) that motivated my comment above:
```
modules/apps/iframe/iframe-web/src/main/resources/META-INF/resources/view.jsp:<aui:script use="aui-autosize-iframe">
modules/apps/layout/layout-type-controller/layout-type-controller-embedded/src/main/resources/META-INF/resources/layout/view/embedded_js.jspf:<aui:script use="aui-autosize-iframe">
modules/apps/license-manager/license-manager-web/src/main/resources/META-INF/resources/view.jsp:<aui:script use="aui-autosize-iframe">
modules/dxp/apps/portal-search-elasticsearch-monitoring/portal-search-elasticsearch-monitoring-web/src/main/resources/META-INF/resources/view.jsp:<aui:script use="aui-autosize-iframe">
```
Namely: the iframe widget, embedded layouts (included in LPS reproduction steps), license manager and elastic search monitoring.

The 2 first are probably the most important, together with the elastic search monitoring (this is a dxp app not a community app). 

Regarding where it'd make sense to make `scrollIntoView` configurable, there are 2 cases:

- Probably yes, make it configurable: iframe widget allows page admins to turn auto-resize on/off, an extra flag would make sense. Embedded layouts don't allow configuration of resizable behavior, so here decision is harder.
- No, don't make it configurable: license and search apps use the iframe because they need, it's not meant for them to allow any configuration for the iframe behavior.

As aui-autosize-iframe maintainers, we don't have to make any decision on the above. Apps using the plugin have to analyze what's best for them. No matter what default value we set for the flag, we must give app owners a chance to decide. If we change the default value of the flag to `true` we likely need to explicitly set it to false in the existing usages so that we don't break existing behavior while they make a decision.

markocikos:
> If we change the default value of the flag to true we likely need to explicitly set it to false in the existing usages so that we don't break existing behavior while they make a decision.

@dsanz Yeah, you are right. My line of thinking was that scrolling to top on iframe open would be by far the most common use case, but I may be wrong on this!

chileces:
dear @markocikos , @dsanz 
I want to escalate to you this issue and report my testing with the proposed PR. due to a lack of experience on how to build the frontend project before applying it to the LiferayPortal code base, i do a workaround like this.
1. Open jar file of the component, ie: com.liferay.frontend.js.aui.web.jar, and replace the aui-autosize-iframe.js in that file 
2. Remove osgi/state, tomcat.x.x/works/Catalina folders
3. restart bundles
Observe it does not break the current behavior of the existing module, no error in the console 
4.
from Liferay Portal, I set 
iframe.autosizeiframe.set('scrollIntoView', true); 
in to https://github.com/liferay/liferay-portal-ee/blob/3f7f033f735a245cca3d5ee3edd09f693c6d298a/portal-web/docroot/html/portal/layout/view/embedded_js.jspf#L22
Observe this will solve the issue with aui-autosize-iframe at Liferay 7.1.x.

By default, the attribute "scrollIntoView" value is false, when the setHeight is triggered, aui-autosize-iframe.js will check the scrollIntoView before scrolling. This will prevent the unexpected behavior of the existing modules. Only where it is set it will be executed.

Please review the changes and let me know what you think. With this fix, we can resolve an LPP issue for our customer.

best regards,
Chi

markocikos (commented):
@chileces Code does not affect DXP master, and I trust you it enables the fix on 7.1. But a quick question: why don't we use [vanilla JS scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) to fix 7.1? Do we need to extend `alloy-ui` library for this?

If we do need to move forward with this PR, please tidy up code a bit:
- Sort attributes, placing the new one before `width`
- Change the commit name to `feat: ...`. We are extending API, not fixing the issue directly.

chileces:
> @chileces Code does not affect DXP master, and I trust you it enables the fix on 7.1. But a quick question: why don't we use [vanilla JS scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) to fix 7.1? Do we need to extend `alloy-ui` library for this?
> 
> If we do need to move forward with this PR, please tidy up code a bit:
> 
> * Sort attributes, placing the new one before `width`
> * Change the commit name to `feat: ...`. We are extending API, not fixing the issue directly.

i've changed the pr as tidy as your suggestion. The issue happened inside an iframe, when its height has changed, it is not auto-scroll to the top, so the user has to scroll manually in order to see the whole iframe content. From liferay portal 7.1.x, i will handle this inside the portal code where we put the aui-autosize-iframe tag and control its behavior.

markocikos (approved):
LGTM!

PR #1016, opened 2022-09-30
feat(eslint-plugin): add ESLint rule to disable non-wrapped web storage API

tinycarol:
### References
- [LPS-158441](https://issues.liferay.com/browse/LPS-158441)

### What is the goal?
* Add an ESLint rule to prevent usages of `localStorage` and `sessionStorage`
   * In [LPS-155901](https://issues.liferay.com/browse/LPS-155901), we added a JS API to enforce user consent when managing web storage in DXP. This PR adds an ESLint rule to prevent using the global `localStorage` and `sessionStorage` objects, redirecting devs to the new wrapped ones.
   
### How to replicate
* Tested this in DXP following these steps:
   * Run `yarn link` in `liferay-frontend-projects/projects/eslint-plugin`
   * Run `yarn link “@liferay/eslint-plugin”` in `liferay-portal/modules`
   * Run `yarn checkFormat`


tinycarol:
Hi @bryceosterhaus! Could you take a look at this PR? Thanks ❤️

bryceosterhaus:
@tinycarol looks great! Thank you.

Do you need me to release a new version right away or when would you like a new version? I suspect we will need to send a PR to portal with the new version and necessary changes

tinycarol:
We actually need to wait for other product teams to update their usages of local/session storage before applying this rule (and the cookie one) in portal. I'll have to send PRs to each of them with partial changes, so I'll let you know when we can activate it.

bryceosterhaus:
@tinycarol sounds good

I think I can actually release this and I will send a PR to portal but leave the rule disabled in our [eslint config](https://github.com/liferay/liferay-portal/blob/master/modules/.eslintrc.js)

bryceosterhaus:
I will cc' you on the PR later

PR #1021, opened 2022-10-04
fix(js-toolkit): stop throwing an error and instead return empty json

bryceosterhaus:
Fixes issue where bundler parses all package.json, even invalid ones.

See internal convo for context: https://liferay.slack.com/archives/C3JBR21HA/p1664782682850919

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/__tests__/alias.test.ts
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/__tests__/alias.test.ts
@@ -75,10 +75,10 @@ describe('loadAliases', () => {
 		).not.toThrow();
 	});
 
-	it('throws when package.json is not valid JSON', () => {
-		expect(() =>
+	it('returns empty json when package.json is not valid JSON', () => {
+		expect(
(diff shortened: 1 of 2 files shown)


georgel-pop-lr (commented):
Would it to be good to show the error as a warning, maybe is helping the user?

izaera:
> Would it to be good to show the error as a warning, maybe is helping the user?
I would write it to the report file (eg: https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/index.ts#L70)

georgel-pop-lr:
> > Would it to be good to show the error as a warning, maybe is helping the user?
> 
> I would write it to the report file (eg: https://github.com/liferay/liferay-frontend-projects/blob/master/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/index.ts#L70)
> 
> Other than that, LNHTM

@bryceosterhaus would you do this change? 

Thanks.

bryceosterhaus:
@georgel-pop-lr a warning or log wouldn't be necessary because this method is specifically looking for the "alias" field in the package.json. If the package.json is malformed, we can expect that alias doesn't work either so returning an empty object should be fine.

PR #1025, opened 2022-10-07
feat(amd-loader): add moduleType configuration setting

izaera:
Since this is done I think it doesn't do any harm to merge it. It may be used in the future if we decide to ship the shims and otherwise it's a simple config setting.


bryceosterhaus (approved):
LGTM!

PR #1030, opened 2022-10-13
fix(npm-scripts): check config if tsc is disables for 'types' command

bryceosterhaus:
fixes inconsistencies between `build` and `types` commands


izaera:
LGTM

PR #1033, opened 2022-10-19
fix(npm-scripts): don't fail when there's no main entry in build config

izaera:
This is needed for https://issues.liferay.com/browse/LPS-165925


izaera:
:ninja: was here

PR #1034, opened 2022-10-21
feat(npm-scripts): create caching mechanism for build script

bryceosterhaus:
~Copies our css logic and only builds js if files have been modified more recently than the built files~

Re-did the whole thing. Here is how it works

First build:
- Stores built files at `./modules/.npmscripts/build/artifacts/${MODULE_NAME}/build/**`
- Hashes src and built files at `./modules/.npmscripts/build/artifacts/${MODULE_NAME}/buildinfo.json`

- Before running babel, bundler, webpack, etc. we check if any previous build exists
    - Checks `./modules/.npmscripts/build/artifacts/${MODULE_NAME}/buildinfo.json`
    - Checks `./modules/.npmscripts/build/artifacts/${MODULE_NAME}/build/**` ()

Subsequent builds:
Check...
- previous build info doesn't exist -> REBUILD
- can't read previous build info -> REBUILD
- number of source files change -> REBUILD
- src files content changes -> REBUILD
- src files content doesn't change
    - build file doesn't exist -> copy previous build from cache
    - previous build files exist
         - file is same as last build -> DO NOTHING
         - file has changed -> REBUILD

Use can also force a build with `yarn build --force` or by deleting the `./modules/.npmscripts` directory.

The only change needed in DXP is adding `./modules/.npmscripts` to .gitignore and also excluding it from `ant clean`.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/build.js
@@ -10,13 +10,16 @@ const path = require('path');
 
 let buildSass = require('../sass/build');
 let runTSC = require('../typescript/runTSC');
+const {isCacheValid, setCache} = require('../utils/buildArtifacts');
 const createAmd2EsmExportsBridges = require('../utils/createAmd2EsmExportsBridges');
 const createEsm2AmdCustomBridges = require('../utils/createEsm2AmdCustomBridges');
 const createEsm2AmdExportsBridges = require('../utils/createEsm2AmdExportsBridges');
 const createEsm2AmdIndexBridge = require('../utils/createEsm2AmdIndexBridge');
 const createTempFile = require('../utils/createTempFile');
+const expandGlobs = require('../utils/expandGlobs');
 const getMergedConfig = require('../utils/getMergedConfig');
 const instrument = require('../utils/instrument');
+const log = require('../utils/log');
 let minify = require('../utils/minify');
 let runBabel = require('../utils/runBabel');
 let runBridge = require('../utils/runBridge');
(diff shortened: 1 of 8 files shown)


projects/npm-tools/packages/npm-scripts/src/scripts/build.js
```diff
@@ -105,7 +108,26 @@ module.exports = async function (...args) {
 
 	const inputPathExists = fs.existsSync(BUILD_CONFIG.input);
 
-	if (!cssOnly) {
+	const files = expandGlobs(
+		[
+			path.join(BUILD_CONFIG.input, '**/*.js'),
```
izaera, line 113: Not sure if you need to add soy here, since the code below is building it too :thinking:

izaera:
I left a comment but other than that LGTM

projects/npm-tools/packages/npm-scripts/src/utils/buildArtifacts.js
```diff
@@ -0,0 +1,160 @@
...
+const root = findRoot();
+
+const npmScriptsCacheDir = path.join(root, '.npmscripts', 'build');
+const artifactCache = path.join(npmScriptsCacheDir, 'artifacts');
+
+const BUILD_INFO_FILE_NAME = 'buildinfo.json';
+
+const hashFile = (filePath) => {
+	const fileBuffer = fs.readFileSync(filePath);
+	const hashSum = crypto.createHash('sha256');
```
izaera, line 23: A SHA-1 would probably be enough for these purposes and may be faster. 

However, I'm OK with SHA-256 as it won't probably take too much longer.

izaera:
LGTM

izaera:
Let's hope it doesn't break anything :crossed_fingers: :sweat_smile:

bryceosterhaus:
re-sending because tests were being weird, https://github.com/liferay/liferay-frontend-projects/pull/1036

PR #1035, opened 2022-10-21
fix(npm-scripts): fix exports when explicit symbols are used for ESM

izaera:
If we export symbols explicitly by configuring this in npmscripts.config.js:

```
module.exports = {
	build: {
		exports: [
			{
				name: 'metal-component',
				symbols: [
					'Component',
					'ComponentDataManager',
					'ComponentRegistry',
					'ComponentRenderer',
					'addListenersFromObj',
					'getComponentFn',
				],
			},
		],
	},
};
```

We need to differentiate if the module exports things in ESM or plain CommonJS format. This is important because it makes the treatment of __esModule and default fields different.


izaera:
Going to :ninja: merge this because I need to test another PR using it

PR #1038, opened 2022-10-31
fix: don't mangle class/function names when minimizing

izaera:


bryceosterhaus (approved):
LGTM!

izaera:
I merged this before it was green and it broke the CI :-(

PR #1042, opened 2022-11-09
feat(npm-scripts): add support for internal modules in esm builds

bryceosterhaus:
I think I will release an alpha so that I can send a PR to dxp testing CI. So far I think this works though.

If you want to include internal modules for an npm module, you need to add `includeInternalPaths` which is an array of paths to include, this when generates a bundle for each of those files and makes them accessible via ESM and AMD(via bridge) in DXP


bryceosterhaus:
testing on dxp with an alpha version in [this PR](https://github.com/liferay-frontend/liferay-portal/pull/2833)

izaera:
Let's see what happens :crossed_fingers:

bryceosterhaus:
Gonna re-visit this at a later date in [this story](https://issues.liferay.com/browse/LPS-168439)

PR #1043, opened 2022-11-17
fix(liferay-npm-build-tools-common): check link target absoluteness for win platform correctly

ruvor:


bryceosterhaus:
@ruvor can you provide some context for this PR? What issues are you facing before this PR and how can we reproduce and check?

ruvor:
I tried to bundle my module in the yarn-operated monorepo. And while it was being done on Mac it worked. Then on Windows the bundler vainly tried to read a dir by the path like D:\work\my-repo\node_modules\D:\work\my-repo\common\npm-bundler-preset-my instead of just D:\work\my-repo\common\npm-bundler-preset-my. Turned out that the bundler getting my config preset wrongfully assumed the path relative and prepended it with additional segment.
The symlinks were placed inside node_modules by yarn which is used to run my monorepo.

izaera:
Maybe we can use https://nodejs.org/api/path.html#pathisabsolutepath instead of using a regexp?

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -516,7 +516,7 @@ export class Project {
 			const linkTarget = fs.readlinkSync(this._toolsDir.asNative);
 
 			this._toolsDir = new FilePath(
-				linkTarget.startsWith('/')
+				process.platform === 'win32' && /^[A-Z]:\\/.test(linkTarget) || linkTarget.startsWith('/')
```
izaera, line 519: How about using https://nodejs.org/api/path.html#pathisabsolutepath instead of using a regexp?

ruvor: Sounds great. And obviously it works. Updated the PR.

bryceosterhaus:
Looks great. Thanks for your help, @ruvor!

maintenance/projects/js-toolkit/packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: Perfect then :+1. Thx :-)

izaera:
I'll release a new version of the JS Toolkit once we merge https://github.com/liferay/liferay-frontend-projects/pull/1045

izaera:
I've released [v2.31.2](https://github.com/liferay/liferay-frontend-projects/releases/tag/liferay-js-toolkit%2Fv2.31.2) that includes this PR.

PR #1045, opened 2022-11-18
fix(liferay-npm-bundler): getBundleVersionAndClassifier was called twice for the bundle-version when generating the jar manifest 

dbagrey:
Fixed a bug with the re-call of bundle-version normalization.
For example, for the version 1.2.3-alpha-SNAPSHOT
Before fix: 1.2.3.alpha.SNAPSHOT
After fix: 1.2.3.alpha-SNAPSHOT
according to the [osgi-core](https://github.com/osgi/osgi/blob/r6-core-spec-final/org.osgi.framework/src/org/osgi/framework/Version.java#L182-L206) - the first option is incorrect

### DIFF
--- a/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
+++ b/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js
@@ -12,7 +12,6 @@ import path from 'path';
 
 import * as ddm from './ddm';
 import Manifest from './manifest';
-import * as osgi from './osgi';
 import * as xml from './xml';
 
 const pkgJson = project.pkgJson;
@@ -106,12 +105,12 @@ function addLocalizationFiles(zip) {
  * @param {JSZip} zip the ZIP file
  */
 function addManifest(zip) {
-	const bundleVersion = osgi.getBundleVersionAndClassifier(pkgJson.version);
-
 	const manifest = new Manifest();
 
 	manifest.bundleSymbolicName = pkgJson.name;
-	manifest.bundleVersion = bundleVersion;
+	manifest.bundleVersion = pkgJson.version;
+	const bundleVersion = manifest.bundleVersion;
+
 	if (pkgJson.description) {
 		manifest.bundleName = pkgJson.description;
 	}


izaera:
We need to check this more thoroughly, because the semver specs for npm and OSGi are not the same, so there are some issues that may arise from trying to convert from one version format to another.

In fact, we (Liferay) experienced such issues too some time ago with the patching tool we have to maintain released versions. I have to investigate what we did then.

What's inside the `package.json` of the project experiencing the issue, in any case? `1.2.3.alpha.SNAPSHOT` ?

izaera:
Some related data:

- npm semver follows this spec -> https://semver.org/ as explained in https://docs.npmjs.com/about-semantic-versioning
- OSGi semver follows this spec -> https://docs.osgi.org/whitepaper/semantic-versioning/040-semantic-versions.html

dbagrey:
In package.json:
``"version": "1.2.3-alpha-SNAPSHOT"``
The code responsible for converting the version works correctly. The only problem is that it is called twice:
1. [in index.js explicitly (line 109)](https://github.com/liferay/liferay-frontend-projects/blob/00d5e493e14f7dfe1a7e0ae9f78a11f29234feeb/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js#L109)
2. [When assigning a bundleVersion value to manifest](https://github.com/liferay/liferay-frontend-projects/blob/00d5e493e14f7dfe1a7e0ae9f78a11f29234feeb/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/index.js#L114)(line 114). [Line in manifest.ts which the repeated call occurs](https://github.com/liferay/liferay-frontend-projects/blob/00d5e493e14f7dfe1a7e0ae9f78a11f29234feeb/maintenance/projects/js-toolkit/packages/liferay-npm-bundler/src/jar/manifest.ts#L35) (line 35)

izaera:
> The code responsible for converting the version works correctly. The only problem is that it is called twice:

Oh,  I see...  So I already fixed the semver issue in the past... That's good :sweat_smile:

izaera:
So, I'm OK with this change, but I want to do some refactor on top of it to make thinks safer and clearer for the future.

Specifically,  I want to move the conversion from manifest.ts to index.js because Manifest is just a data holding class, so it shouldn't be doing non-idempotent conversions on its own. But for that, I also need to move the tests and refactor index.js a bit.

I'll put my refactor on top of your commit, resend this PR, and link the two. Stay tuned :wink: .

And thanks for the contribution :+1:

izaera:
Resent as https://github.com/liferay/liferay-frontend-projects/pull/1046

izaera:
I've released [v2.31.2](https://github.com/liferay/liferay-frontend-projects/releases/tag/liferay-js-toolkit%2Fv2.31.2) that includes this PR.

PR #1050, opened 2022-11-24
feat(amd-loader): update resolve path when a redirect is received from server

izaera:
This is part of https://issues.liferay.com/browse/LPS-169194

See commit for a thorough description of what is changed.


projects/amd-loader/bin/util.js
```diff
@@ -13,18 +13,40 @@ const copy = (input, output) =>
...
+	const options = {
 		shell: true,
 		stdio: 'inherit',
-	});
+	};
+
+	if (env) {
+		options['env'] = {
+			...process.env,
+			...env,
```
bryceosterhaus, line 28: nit: should env override process.env? I would think it might be the other way around. Not sure what is typically done for that

projects/amd-loader/src/loader/config.js
```diff
@@ -207,6 +207,22 @@ export default class Config {
...
 	}
 
+	/**
+	 * Update the resolve path (usually as a consequence of a redirect being
+	 * received from the server when a module resolution is attempted).
+	 * @param {string} resolvePath the updated resolve path
+	 * @return {void}
+	 */
+	updateResolvePath(resolvePath) {
+		const prefix = `${window.location.protocol}//${window.location.host}`;
```
bryceosterhaus, line 217: nit: I think you can also use `new URL` so that you don't have to manually concat. Although if this is for a fie path, not sure itd work as expected

bryceosterhaus (approved):

projects/amd-loader/bin/util.js
izaera: Yes, you are right: I'm breaking the usual expectations. 

The problem [here](https://github.com/liferay/liferay-frontend-projects/pull/1050/files#diff-03935a18504c29dc8a0c37cf7c460bdd348624e90ad919380a8640fb5d1f974fR8-R10) is that if someone sets `flavor` in the env, the build will build the same three times.

But, since this function (run) is generic, I'll make it work generically, then tweak the calling code :+1:

projects/amd-loader/src/loader/config.js
izaera: :thinking: I don't think so, because [URL's constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) would need to get ``${window.location.protocol}//${window.location.host}`` in any case...

Or am I missing anything?

izaera: For the reference: this is just an optimization to remove the host and port from the redirect URL because, when configured from scratch, the AMD loader is missing them too. But if we didn't remove it, it would work too, so we don't need to be exhaustive when `cleaning` the URL.

izaera: By exhaustive I mean that this replacement would fail if the URL had credentials in it, for example.

projects/amd-loader/bin/util.js
bryceosterhaus: If its too much of a headache, don't worry about it

projects/amd-loader/src/loader/config.js
bryceosterhaus: ah I gotcha, I think I was actually mistaken with URL, my mind was somewhere else...

projects/amd-loader/bin/util.js
izaera: I've fixed it :rocket:

projects/amd-loader/src/loader/config.js
izaera: Maybe you were thinking in Java's URL class... :monocle_face:

PR #1053, opened 2022-12-19
LRQA-78830 | Update Poshi Standalone support for liferay-frontend-projects repo

nkkwon:
Updating Poshi standalone to work for liferay-frontend-projects.
Need to check with versions of portal-master with poshiRunnerResources in build.gradle files to see if it works with current tests


izaera:
> see if it works with current tests

IINM, these tests were done for JS Toolkit v2, which is at https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit, but some parts of that project are being slowly deprecated in favor of @liferay/cli, which is at https://github.com/liferay/liferay-frontend-projects/tree/master/projects/js-toolkit, so we may want to create the tests there, instead of updating the old ones...

CC: @john-co

PR #1054, opened 2022-12-21
feat(npm-scripts): add support for multiple entry points

bryceosterhaus:
cc @izaera let me know what you think about this, it seems to work fine locally for me


izaera (approved):
LGTM. 
However, I would move the logic to test for Array to the place where the config is read, then assume that it is always an array in the other places, where you consume the data. That would save you a lot of code :+1:

bryceosterhaus:
@izaera good idea, I moved it to our `normalizeNpmscriptsConfig` util

izaera:
So, LGTM again :sweat_smile:

bryceosterhaus:
gonna test again tomorrow and release if it works as expected

PR #1058, opened 2023-01-06
feat: add spritemap sample test file for svg client extension functional test

john-co:
The spritemap file will be a dependency for functional test described in https://issues.liferay.com/browse/LPS-171424

Test Scenario file will be used in:
Given: Sample SVG file from external site
// spritemap.svg available on https://liferay.github.io/liferay-frontend-projects/
And Given:  A created Theme svg extensions using the external URL 
// from github.io
And Given: Navigate to edit the “Look and Feel" on home page
When:  Update with registered "Theme Spritemap Client Extension"
And When: Navigate to home page
Then: new SVG spritemap is applied on the page
// browser devtools > network tab or find an api command to check the SVG icon is available
// Or swap two known icons in the SVG file and assert those two icons have been swapped on the portal UI


bryceosterhaus:
Thanks @john-co !

PR #1060, opened 2023-01-11
fix(npm-scripts): copy package.json main entry point if it exists and reflect file name accurately in manifest.json

bryceosterhaus:


izaera:
LGTM if it works :+1:

PR #1062, opened 2023-01-12
fix(theme-tasks): update liferay-frontend-theme-styled, liferay-frontend-theme-unstyled

rolandpakai:
Hi Team.

could you please check my fix?

The fix upgrades the liferay-frontend-theme-styled, liferay-frontend-theme-unstyled to the @liferay_frontend["icon-options"] taglib instead of the depreceted @liferay_portlet["icon-options"] 

Thanks, Roland

https://issues.liferay.com/browse/LPS-172679

### DIFF
--- a/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/devDependencies.js
+++ b/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/devDependencies.js
@@ -77,8 +77,8 @@ module.exports = {
 				'compass-mixins': strict('0.12.10'),
 				'gulp': gulpVersion,
 				'liferay-frontend-css-common': strict('6.0.4'),
-				'liferay-frontend-theme-styled': strict('6.0.38'),
-				'liferay-frontend-theme-unstyled': strict('6.0.33'),
+				'liferay-frontend-theme-styled': strict('6.0.52'),
+				'liferay-frontend-theme-unstyled': strict('6.0.45'),
 				'liferay-theme-tasks': themeTasksVersion,
 			},
 			optional: {


izaera:
According to @dsanz (in an internal slack conversation):

> if we merge the pull, any new theme created with the generator won't work in older 7.4 releases as it'd use a non-existent tag

I'm not aware of the nuances of this, but looks like we cannot merge this PR without looking at it, so I'm setting this PR to draft to prevent it from being merged until we find out...

dsanz:
Thanks @izaera, yes we need to decide how to handle the case depending on the 7.4 version for which theme is being created

rolandpakai:
Hi @dsanz  @izaera 

do you have any update about this case ?

Thanks, Roland

dsanz:
Hi @rolandpakai 

in the end we need to merge this one. The conversation is more about whether if we shall do more to inform users about the consequences of using these dependencies in earlier 7.4 releases or not.

With the new dependencies, we start to use a new tag in the ftl tmplates. This tag does not exist in earlier versions, so the theme we create from the dependencies won't work.

@izaera, @bryceosterhaus probably, we can just have our tooling to explain this to the user when creating a new theme, providing the necessary info so that user can change the dependencies manually if needed. By default we shall use the ones in this PR, then just ask user to use older if not in 7.4U_xy_ how does that sound?

izaera:
> By default we shall use the ones in this PR, then just ask user to use older if not in 7.4U_xy_ how does that sound?

Yeah, that seems to be the global agreement everybody assumed in Slack when discussing about this:

1. Updating the themes toolkit to the latest versions whenever they are updated in NPM (would need to create an LPS to write a CI action for that)
2. Show a notice to the user with a link to the official documentation (which would be a responsibility of the PEDS team).

izaera:
I've unlocked the PR to merge it and release a new themes toolkit version when we have time (most probably on Monday).

izaera:
Resent as https://github.com/liferay/liferay-frontend-projects/pull/1074

izaera:
Resent as https://github.com/liferay/liferay-frontend-projects/pull/1074

rolandpakai:
Hi @izaera 

thank you for the updateThemeDeps.js script and automating the package upgrade.
Where can we check the new theme generator release which contains the  upgrades?

Thanks, Roland

izaera:
@rolandpakai I'm waiting for the PEDS teams (which are the real owners of these assets) to give their OK, then I'll release a new Themes Toolkit version (see https://github.com/liferay/liferay-frontend-projects/pull/1074#issuecomment-1419119992 and https://liferay.slack.com/archives/C049XG98GQL/p1675691519605839).

PR #1064, opened 2023-01-18
fix(npm-scripts): temporary work around for grabbing rules config at both the root and module level

bryceosterhaus:
This is temporary until we provide a full solution in https://issues.liferay.com/browse/LPS-172749


izaera:
:scream:

izaera:
I've relaunched the windows test, since it looks like it could be valid :shrug:

izaera:
It passed the test :+1:

bryceosterhaus:
The windows failure was due to a async timeout. I've seen it happen before, I think it may have to just do with the machine on github actions, so I figured it would be fine to merge. Plus I like to live dangerously

izaera:
YOLO

PR #1066, opened 2023-01-19
feat: use portlet category "Client Extensions" by default, for new projects

izaera:
See https://issues.liferay.com/browse/LPS-172978.

The goal of this PR is to keep using `category.remote-apps` for old projects (that's detected because `liferay.json` is empty for them) but use `category.client-extensions` from now on.


bryceosterhaus (approved):
LGTM

izaera:
Released artifacts to npm for the three modified packages.

dsanz:
Iinm, this one does not require deps upgrade in DXP so we basically are done with this

izaera:
> we basically are done with this
yes

PR #1069, opened 2023-01-26
feat: support internal path exports in npm-scripts

izaera:
https://issues.liferay.com/browse/LPS-172747

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdExportsBridges.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdExportsBridges.js
@@ -32,7 +32,9 @@ function createEsm2AmdExportsBridges(projectDir, buildConfig, manifest) {
 	const {exports, output} = buildConfig;
 
 	exports.forEach((exportItem) => {
-		let {modulePath, pkgName, scope} = splitModuleName(exportItem.name);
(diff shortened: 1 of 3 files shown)


projects/npm-tools/packages/npm-scripts/src/utils/createEsm2AmdExportsBridges.js
```diff
@@ -71,18 +91,34 @@ function createEsm2AmdExportsBridges(projectDir, buildConfig, manifest) {
...
+		// Also, for internal module paths, we must add more `../`s.
+		//
+
+		let rootDir = '../..';
+
+		if (exportItem.name.startsWith('@')) {
+			rootDir += '/..';
+		}
+
+		const hopsToAdd = modulePath.split('/').length - 1;
```
bryceosterhaus, line 103: is `hops` a technical thing? It totally makes sense to me, but wasn't sure if it was actually a thing

izaera: Not really in this context, but I couldn't come up with a better word 🤷.

In computer networking terminology [a hop is a jump from one router to another](https://www.lifewire.com/what-are-hops-hop-counts-2625905). It resembled this case (for me).

izaera:
Windows test failing once again :open_mouth: 

How come? :roll_eyes:

PR #1073, opened 2023-02-06
chore: script to check theme dependency versions against npm registry

izaera:
This script will be used to maintain Themes Toolkit's dependency versions up to date.

For now it can only be run locally, but we can use the --check option of the script to create a GH workflow to check for new versions daily and notify the team accordingly.

### DIFF
--- a/projects/js-themes-toolkit/scripts/updateThemeDeps.js
+++ b/projects/js-themes-toolkit/scripts/updateThemeDeps.js
@@ -1,0 +1,164 @@
+const fs = require('fs');
+const https = require('https');
+
+const DEV_DEPENDENCIES_PATH = require.resolve(
+	'../packages/liferay-theme-tasks/lib/devDependencies.js'
+);
+const SIGNIFICANT_DEPS = [
+	'liferay-frontend-css-common',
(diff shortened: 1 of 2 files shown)


projects/js-themes-toolkit/scripts/updateThemeDeps.js
```diff
@@ -0,0 +1,171 @@
...
+
+			currentVersions[dxpVersion][dep] = depVersion;
+			wantedVersions[dxpVersion][dep] = latestVersion;
+		}
+	}
+
+	console.log('');
+	console.log('');
+
+	if (argv[2] === '--check' || argv[2] === '-c') {
```
bryceosterhaus, line 65: Not a big deal, but you could also do `argv.includes('--check')` in case at a later date we add another arg that may be in a different order.

bryceosterhaus (commented):
Looks good to me

projects/js-themes-toolkit/scripts/updateThemeDeps.js
izaera: True :+1:

PR #1086, opened 2023-02-24
feat: support internal path in exports field

izaera:
This is needed for [LPS-176053](https://issues.liferay.com/browse/LPS-176053) implementation because we need to create a `@liferay/api` module that will export all our public APIs and we have decided (as a start) to modularize the API with internal paths, like this:

* `@liferay/api`: will contain general APIs like the ones we provided now in the `Liferay` object
* `@liferay/api/data-set`: will contains data-set APIs like the custom cell renderer we are investigating in [LPS-176053](https://issues.liferay.com/browse/LPS-176053)
* ...

If this doesn't break the current portal build (it shouldn't) I'll release a new version so that those wanting to do it can run the PoC locally.


izaera:
Resent as https://github.com/liferay/liferay-frontend-projects/pull/1087

PR #1087, opened 2023-02-24
feat: support internal path in exports field 

izaera:
This is needed for [LPS-176053](https://issues.liferay.com/browse/LPS-176053) implementation because we need to create a `@liferay/api` module that will export all our public APIs and we have decided (as a start) to modularize the API with internal paths, like this:

* `@liferay/api`: will contain general APIs like the ones we provided now in the `Liferay` object
* `@liferay/api/data-set`: will contains data-set APIs like the custom cell renderer we are investigating in [LPS-176053](https://issues.liferay.com/browse/LPS-176053)
* ...

If this doesn't break the current portal build (it shouldn't) I'll release a new version so that those wanting to do it can run the PoC locally.


izaera:
Since this is green and I need to release a new version to test, I'll merge it.

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
```diff
@@ -252,14 +253,47 @@ function normalizeNpmscriptsConfig(mergedConfig) {
...
+				// to force users to specify it in the configuration
+
+				if (!exportsItem.name.startsWith('__liferay__/')) {
+					exportsItem.name = `__liferay__/${exportsItem.name}`;
+				}
+				else {
+					throw new Error(
+						'Export names should not begin with __liferay__/ ' +
+							'since it is automatically added by the build.'
+					);
```
bryceosterhaus, line 293: Is it weird to throw an error here? Seems like if we know that the user is starting the export path with `__liferay__/` then we just let it go but not explicitly add it ourselves.

Granted, this could lead to inconsistencies in our modules where some have it and some don't and both work

projects/npm-tools/packages/npm-scripts/src/utils/getExternalExportsWebpackConfigs.js
```diff
@@ -0,0 +1,246 @@
...
+				experiments: {
+					outputModule: true,
+				},
+				externals,
+				externalsType: 'module',
+				module: {
+					rules: [
+						{
+							exclude: /node_modules/,
+							test: /\.js$/,
```
bryceosterhaus, line 50: Will we need to account for .ts(x) files or .jsx files? I see you are doing that in the other two webpack config files

projects/npm-tools/packages/npm-scripts/src/utils/runWebpackAsBundler.js
```diff
@@ -3,16 +3,12 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-/* eslint-disable @liferay/no-dynamic-require */
+/* XXeslint-disable @liferay/no-dynamic-require */
```
bryceosterhaus, line 6: XX ?

bryceosterhaus (commented):
Looks good to me, just had a few random questions while reading through

projects/npm-tools/packages/npm-scripts/src/utils/runWebpackAsBundler.js
izaera: Oops, that must be an error. I didn't type that on purpose...

projects/npm-tools/packages/npm-scripts/src/utils/getMergedConfig.js
izaera: Yeah, I prefer to fail so that we don't end up with a mix of different configurations. Since this would only happen when changing the configuration of an existing project or adding a new one, I feel it's helpful.

projects/npm-tools/packages/npm-scripts/src/utils/runWebpackAsBundler.js
izaera: https://github.com/liferay/liferay-frontend-projects/pull/1088

PR #1095, opened 2023-03-02
chore: remove leftover of past freeze

izaera:
We freeze old versions of Themes Toolkit when a new version of DXP appears because the lifecycles are different. However `bs3-bs4-compat` is an utility package that is not tied to any DXP version series in particular, so it should remain always just in master.


izaera:
Merging because @bryceosterhaus told me to do it 

not my fault

PR #1100, opened 2023-03-09
feat: add @liferay/js-api project

izaera:
Part of https://issues.liferay.com/browse/LPS-177316


.eslintrc.js
```diff
@@ -26,6 +26,8 @@ module.exports = {
 		'/projects/amd-loader/src/demo/modules',
 		'/projects/js-themes-toolkit/packages/liferay-theme-tasks/lib/r2',
 		'/projects/js-themes-toolkit/qa',
+		'/projects/js-toolkit/packages/js-api/**/*.d.ts',
+		'/projects/js-toolkit/packages/js-api/**/*.js',
```
bryceosterhaus, line 30: why ignore? Shouldn't the generated js and d.ts only be built when published?

projects/js-toolkit/packages/js-api/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "../../tsconfig.json",
+	"compilerOptions": {
+		"composite": true,
+		"lib": ["DOM", "ES2020"],
+		"outDir": ".",
+		"rootDir": ".",
```
bryceosterhaus, line 7: should this be `./src` to follow our other patterns?

projects/js-toolkit/packages/js-api/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "../../tsconfig.json",
+	"compilerOptions": {
+		"composite": true,
+		"lib": ["DOM", "ES2020"],
+		"outDir": ".",
```
bryceosterhaus, line 6: Our other patterns would have this be `./lib`

target-platforms/packages/dxp-7.4/package.json
```diff
@@ -147,16 +162,23 @@
...
+		"react-transition-group": "4.4.1",
 		"recharts": "1.8.5",
 		"senna": "2.7.9",
+		"text-mask-addons": "3.8.0",
+		"text-mask-core": "5.1.2",
+		"uuid": "9.0.0",
 		"xss-filters": "1.2.7"
 	},
-	"description": "Target platform for Liferay Portal 7.4.13-u4",
+	"description": "Target platform for Liferay Portal 7.4.13-u66",
```
bryceosterhaus, line 181: whoa, quite the diff here

target-platforms/packages/dxp-7.4/package.json
```diff
@@ -170,5 +192,5 @@
 		"preversion": "yarn ci",
 		"test": "cd ../.. && yarn test"
 	},
-	"version": "5.0.0"
+	"version": "67.0.0"
```
bryceosterhaus, line 195: Why is our version our uXX version plus 1?

bryceosterhaus (commented):
idea is good, file structure I think can fall under our typical standards of `src` and `lib`(or build or dist) and then src gets generated to that other folder which is .gitignored and really only used on publish

.eslintrc.js
izaera: Yeah, the problem is that locally you will usually have them generated and the linter complains before sending a PR if you run it, for example. So, to make our lives easier...

projects/js-toolkit/packages/js-api/tsconfig.json
izaera: No, it can't be because of how typescript works when this package is consumed :-(. The problem is that in typescript you can either:

1. Designate a single global .d.ts file as your package entry point, or 
2. Let it work by default (but then everything needs to be at the root of the package)

And the problem with 1 is that nobody knows how such file can be generated :shrug: . The official docs make a reference to a "bundled .d.ts file", but there's no explanation anywhere on how to build such file.

It could be that I haven't found it, but I can tell you I tried hard :-/

izaera: We could also create src and lib directories, then create a prepublish script to undo that, but it sounds risky bc we can break releases. Though I have to admit I don't like having everything hanging around when doing local development because even with .gitignore and .eslintignore properly configured, it's a bit of a hotchpotch...

target-platforms/packages/dxp-7.4/package.json
izaera: Yeah, that's because we don't release automatically yet and because, for now (without formal APIs) any version is valid for any customer.

target-platforms/packages/dxp-7.4/package.json
izaera: See here -> https://github.com/liferay/liferay-frontend-projects/tree/master/target-platforms/packages/create-platform#liferay-dxp-tags (and save it as a bookmark because I have to visit it each time I have to modify or release target platforms :joy: ).

Basically :point_down: 

> The 7.4 structure is similar to the one in CE but using ga# for the first release of a major version series, and -u# for successive updates.

So, 1.x.x is for the initial release and 2.x.x for the second (which is `u1`).

In CE there's not that problem because the first release is `ga1`.

projects/js-toolkit/packages/js-api/tsconfig.json
izaera: Found this https://github.com/TypeStrong/dts-bundle, but it's experimental...

izaera:
I'm merging this to be able to continue. We may probably iterate in the next days, but I need to release a snhpshot to npm to continue.

PR #1101, opened 2023-03-10
feat(js-toolkit): adds support for building theme spritemap client extensions

bryceosterhaus:
Config for this CET looks like

```json
{
	"build": {
		"type": "themeSpritemap",
		"options": {
			"extendClay": true
		}
	}
}
```

Icons are placed in the `./src` of your project.

package.json looks like
```json
{
  "name": "sample-theme-spritemap-3",
  "version": "1.0.0",
  "description": "Sample Theme Spritemap 3",
  "dependencies": {
    "@liferay/dxp-7.4": "*"
  },
  "scripts": {
    "build": "liferay build",
    "clean": "liferay clean",
    "deploy": "liferay deploy"
  }
}
```

### DIFF
--- a/projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Build.ts
+++ b/projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Build.ts
@@ -8,15 +8,17 @@ import fs from 'fs';
 import FilePath from '../../file/FilePath';
 import LiferayJson, {
 	BuildConfig,
+	BuildType,
 	Bundler2BuildConfig,
 	CustomElementBuildConfig,
 } from '../../schema/LiferayJson';
 import Project from './Project';
 import persist, {Location} from './persist';
 
-type BuildType = 'bundler2' | 'customElement';
-
-type BuildOptions = Bundler2BuildOptions | CustomElementBuildOptions;
+type BuildOptions =
+	| Bundler2BuildOptions
+	| CustomElementBuildOptions
+	| ThemeSpritemapBuildOptions;
 
 export interface Bundler2BuildOptions {
 	minify: boolean;
@@ -27,6 +29,11 @@ export interface CustomElementBuildOptions {
 	htmlElementName: string | null;
 	minify: boolean;
 	portletCategoryName: string;
+}
+
+export interface ThemeSpritemapBuildOptions {
+	enableSVG4Everybody: boolean;
+	extendClay: boolean;
 }
 
 type OptionValue = boolean | number | string;
@@ -49,6 +56,11 @@ export default class Build {
 					project,
 					config as CustomElementBuildConfig
 				);
+				break;
+			case 'themeSpritemap':
+				this.type = 'themeSpritemap';
+				this.dir = project.dir.join('build');
+				this.options = config as ThemeSpritemapBuildOptions;
 				break;
 
 			case 'bundler2': {
--- a/projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Build.ts
+++ b/projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Build.ts
@@ -8,12 +8,13 @@ import fs from 'fs';
 import FilePath from '../../file/FilePath';
 import LiferayJson, {
 	BuildConfig,
-	BuildType,
 	Bundler2BuildConfig,
 	CustomElementBuildConfig,
 } from '../../schema/LiferayJson';
 import Project from './Project';
 import persist, {Location} from './persist';
+
+type BuildType = 'bundler2' | 'customElement' | 'themeSpritemap';
 
 type BuildOptions =
 	| Bundler2BuildOptions
--- a/projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Project.ts
+++ b/projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Project.ts
@@ -8,6 +8,7 @@
 
 import merge from 'deepmerge';
 import fs from 'fs';
+import yaml from 'js-yaml';
 import os from 'os';
 import path from 'path';
 import resolve from 'resolve';
@@ -20,6 +21,7 @@ import Deploy from './Deploy';
 import Dist from './Dist';
 import Start from './Start';
 
+import type ClientExtensionYaml from '../../schema/ClientExtensionYaml';
 import type {Writable} from './Writable';
 
 export default class Project {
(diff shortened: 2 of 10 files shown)


projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Build.ts
```diff
@@ -8,15 +8,17 @@ import fs from 'fs';
 import FilePath from '../../file/FilePath';
 import LiferayJson, {
 	BuildConfig,
+	BuildType,
```
izaera, line 11: Mmm, this was duplicated on purpose. The reason being that `liferay.json`'s schema is a contract between us and our users and the build type is internal to `@liferay/cli` and may change in the future.

So, for example, we could have three different types of builds in `liferay.json` that map to the same internal build in `@liferay/cli`.

However, that may be achieved at code level too maybe (by delegating). So I'm OK with deduplicating it for now, as long as we agree on not changing `LiferayJson` in the future because that would break customer projects.

Note that it is also possible that we end up abandoning liferay.json if we finally converge with blade, so... :shrug:

projects/js-toolkit/packages/portal-base/src/build/themeSpritemap.ts
```diff
@@ -0,0 +1,172 @@
...
+
+import abort from '../util/abort';
+
+const {error, info, print} = format;
+
+const HEADER_REGEXP = /<!--(.*)-->/s;
+
+const SPRITEMAP_FILE_NAME = 'spritemap.svg';
+
+export default async function customElement(project: Project): Promise<void> {
```
izaera, line 28: Rename customElement to themeSpritemap ;-)

projects/js-toolkit/packages/portal-base/src/build/themeSpritemap.ts
```diff
@@ -0,0 +1,172 @@
...
+	const {enableSVG4Everybody = false, extendClay = false} = project.build
```
izaera, line 29: I would move defaults to the Project "framework". The idea is that any code using `project` can assume it doesn't need to do any defaults, normalization or even decision handling.

In the case of "decision" I'm referring to: if we can remove an `if` by returning a different value from the config, for example, then move that logic to the config Project framework.

Example: instead of adding an `if` that decides an output file name, add an output property in the config and decide that name in the Project side. That way the build code remains cleaner and simpler. Of course this cannot be done for everything, but I've found it's a good pattern in the past...

izaera: Also, when normalizing the config I try to canonicalize null, undefined, and that stuff to the most restrictive type possible. So, if I can use a boolean, I  normalize it so that it is not null or undefined, when reading the config.

projects/js-toolkit/packages/portal-base/src/build/themeSpritemap.ts
```diff
@@ -0,0 +1,172 @@
...
+		const clayPath = require.resolve('@clayui/css');
+
+		if (!clayPath) {
+			print(
+				error` @clay/css package not found. Add as a dependency to your client extension via npm.`
+			);
+		}
+
+		const claySpritemapPath = require.resolve(
+			'@clayui/css/lib/images/icons/icons.svg'
```
izaera, line 55: is this going to be there forever or could it change depending on clay's version?

If it can change we may need to move it to the config and provide a default by inspecting the project (or hard coded in first place).

projects/js-toolkit/packages/portal-base/src/build/themeSpritemap.ts
```diff
@@ -0,0 +1,172 @@
...
+			typeSettings: [
+				`enableSVG4Everybody=${enableSVG4Everybody}`,
+				`url=${SPRITEMAP_FILE_NAME}`,
+			],
+		},
+	};
+
+	const zip = new JSZip();
+
+	zip.file(
```
izaera, line 152: We may need to refactor all this stuff in the future bc we surely don't want to have the (almost) same zipping code in every CE type.

projects/js-toolkit/packages/portal-base/src/build/themeSpritemap.ts
bryceosterhaus: I think this will always be there, or at least it hasn't changed in the entire history of clay.

projects/js-toolkit/packages/portal-base/src/build/themeSpritemap.ts
bryceosterhaus: I removed the defaults, as they weren't even needed, but I'm not really sure what you mean by the examples...

> we can remove an `if`

Any "if"? or was there a specific one in mind?

bryceosterhaus:
@izaera updated

and yeah I still need to figure out the test failures

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Build.ts
```diff
@@ -50,6 +58,11 @@ export default class Build {
 					config as CustomElementBuildConfig
 				);
 				break;
+			case 'themeSpritemap':
```
izaera, line 61: I thought prettier enforced lines above `case` (as in `return`) :thinking: 

If, not, it's incredible how consistent I've managed to be :joy:

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Project.ts
```diff
@@ -8,6 +8,7 @@
 
 import merge from 'deepmerge';
 import fs from 'fs';
+import yaml from 'js-yaml';
```
izaera, line 11: YAFTDW (Yet Another Format to Deal With) :tada:

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Project.ts
```diff
@@ -61,7 +63,31 @@ export default class Project {
...
+				const yamlConfig = yaml.load(
+					fs.readFileSync(
+						this.dir.join('client-extension.yaml').asNative,
+						'utf8'
+					)
+				);
+
+				liferayJson = merge.all([
+					liferayJson,
+					this._normalizeClientExtensionYaml(yamlConfig),
```
izaera, line 82: Maybe we should warn about duplicates... I've seen tools doing it (for example `tsc` checkes all JSON files against their schemas) and sometimes saved me a lot of headaches.

No need to do it now if you don't want to, but we may once all this stuff is settled. Otherwise people can get very confused if they see the same thing in two files, for example. Or even worse, if they are only looking at one and forgetting about the other :grimacing: .

projects/js-toolkit/packages/portal-base/src/build/themeSpritemap.ts
izaera: :+1:

izaera:
LVGTM :+1:

projects/js-toolkit/packages/js-toolkit-core/src/project/liferayCli/Project.ts
izaera: Yes and no... Because I doubt we can do everything with just the yaml file :thinking: We may need more options I mean, and until we have a proper blade team in place I don't think it will be easy for us to put them in the yaml file. 

But yeah, it would probably be a good ending point: putting everything in the yaml file.

PR #1102, opened 2023-03-13
chore(js-toolkit): add qa scripts for theme spritemap cet

bryceosterhaus:


projects/js-toolkit/scripts/qa/index.js
```diff
@@ -124,6 +125,11 @@ async function main() {
 		generateAngularCli('angular-cli-portlet');
 		generateCreateReactApp('create-react-app-portlet');
 		generateVueCli('vue-cli-portlet');
+		generateThemeSpritemapClientExtension('theme-spritemap-cet');
```
izaera, line 128: :thinking: I'm thinking now that, if we are going to generate one project per new type to do QA we may as well move this code into `@liferay/cli`...

I still see the problem that if we move it there it will have to be "supported". But we could also just leave it as experimental. And, in any case, if it doesn't work for our users, it won't work for our QA, so....

What do you think @bryceosterhaus ?

bryceosterhaus: I was thinking similarly, especially since it is so simple.

What I will do is merge this and send a PR to portal first with the supported process for the workspace samples and then if I have more time later this week, I will work on that in this sprint, otherwise I will spawn a new ticket to support `liferay new`

izaera: :ok_hand:

PR #1104, opened 2023-03-16
feat(liferay-cli): add new 'theme spritemap' target type

bryceosterhaus:
note, I added "(Experimental)" to the name, this way we don't have to make any grand promises :trollface: 

image

I realized we still use "remote app" for naming stuff. Not sure if I should update that or just use "client extension" now and migrate those later.

### DIFF
--- a/.github/workflows/projects.yml
+++ b/.github/workflows/projects.yml
@@ -22,7 +22,7 @@ on:
 
 env:
     CI: true
-    yarn-cache-name: yarn-cache-4
+    yarn-cache-name: yarn-cache-1
     yarn-cache-path: .yarn
 
 jobs:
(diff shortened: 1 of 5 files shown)


izaera:
Woa, I never thought anyone could understand all the facets, target, etc, stuff without asking me and/or swearing :sweat_smile: 

Very good job :+1: 

> note, I added "(Experimental)" to the name, this way we don't have to make any grand promises :trollface:

We should add `experimental` everywhere. We should make a GH action to SF all static labels and add the suffix `(experimental)` to all of them :muscle: 

> I realized we still use "remote app" for naming stuff. Not sure if I should update that or just use "client extension" now and migrate those later.

Technically, "remote app" is one of "custom element client extension" or "iframe client extension" so you can also call any "remote app" a "client extension" => yes, we can stop using "remote apps" and replace it by "client extension" if that makes our lives easier.

izaera (approved):

bryceosterhaus:
> Woa, I never thought anyone could understand all the facets, target, etc, stuff without asking me and/or swearing 

it was actually pretty straight forward and easy to add, so well done on the initial implementation!

PR #1105, opened 2023-03-17
feat(@liferay/cli): add a way to specify what version of dxp that the target project is supported

bryceosterhaus:
Below is an example if we added `dxpSupport` key to multiple targets. Right now I am only adding experimental to the theme spritemap
Screen Shot 2023-03-17 at 1 11 15 PM


izaera:
Very good :ok_hand: 

we may refine it in the future to make it better, but for now it at least conveys to customer what we wanted

izaera:
And I think this will also help a lot with support tasks :heart_eyes:

PR #1108, opened 2023-03-21
fix(alloy-ui): AUI-3218 improves modal and popover accessibility movi…

georgel-pop-lr:
…ng the focus to the `role="dialog"`, this avoids escaping the Modal focus control along with `aria- modal="true"`


georgel-pop-lr:
Included in 3.1.0-deprecated.111

PR #1109, opened 2023-03-22
chore: add git ignore .idea files

georgel-pop-lr:


bryceosterhaus:
Adding the gulp script is a little odd at first glance... But I guess not necessarily wrong. 🤷

georgel-pop-lr:
> Adding the gulp script is a little odd at first glance... But I guess not necessarily wrong. shrug
Hi, 

I agree but is very useful, I always added it myself by hand but I guess if people see it there they would use it.

Regards.

izaera:
@georgel-pop-lr you can achieve the same functionality by running `npx gulp` in the project folder (without having to install the package globally)

izaera:
Also, I think that `./node_modules/bin/...` thing may fail in Windows

georgel-pop-lr:
> And in some circumstances, it will also fail if yarn hoists it.
> 
> TBH, I would send a follow-up to change that to either remove the script or use:
> 
> ```json
> "scripts": {
>      "gulp": "npx gulp"
> }
> ```
> 
> Or even better:
> 
> ```json
> "scripts": {
>      "gulp": "gulp"
> }
> ```
> 
> I think that works too and doesn't fail under yarn or windows scenarios.

Thanks for your suggestios: https://github.com/liferay/liferay-frontend-projects/pull/1112

PR #1115, opened 2023-03-28
feat: make Liferay global object typing available in target platforms

izaera:
And some more other miscellaneous tweakings

https://issues.liferay.com/browse/LPS-179879


izaera:
I'm going to extract some things from this PR that are not really related to the `Liferay` global object so that we can merge them and be able to generate a FDS cell renderer working example.

Then I'll move the `Liferay` stuff to `liferay-portal` so that we can provide:

1. The global `Liferay` ambient type
2. A real typescript `Liferay` interface for people to use (instead of the global object)

The problem we (@bryceosterhaus and me) see in this proposed pattern is that the switch from a global to an import and the fact that the imported type is a subset of the global object will possibly make adoption much more difficult/impossible because of how people are used to invoke the global and have all functions available.

So, we'll make sure that both the global and the new typescript type are in sync always (by moving this to portal and publishing them at the same time), then we'll promote the switch. Some useful resources for promoting the use of an import vs a global object that I found while searching the web:

- https://www.baeldung.com/cs/global-variables
- https://thevaluable.dev/global-variable-explained

In general, the web is full of articles explaining why using globals is such a bad idea.

ethib137:
@izaera , so is the plan to eventually remove access to the Liferay global object? Or, will this always stay around as a fallback when an import isn't a possibility such as in the JS of a page fragment?

izaera:
@ethib137 the global Liferay object will be available, we will just wrap it, most probably. Good point (the access from fragments) :+1:

izaera:
In any case, we may have support for imports in fragments in the future (if we wanted to) since we would only need to make them be `<script type="module">` and you would be able to do things like:

```
import React from 'react';
```

For example...

ethib137:
Yep, the ability to import modules in fragments would be great, and would solve that problem. We just need to make sure it is solved well before we remove the Liferay global object. 😄

izaera:
> before we remove the Liferay global object. 

It may not be possible to remove it completely, because its implementation is generated from the server side. Our plan is to try to make it an internal thing and provide a better API on top of it, but we cannot remove it completely.

PR #1116, opened 2023-03-29
feat: make @liferay/js-api available from portal-master target platform

izaera:


izaera:
We are not going to this yet. We will override the `@liferay/js-api` version to point to a pre-release and use a valid GA target plaform instead of `portal-master`. That saves us from having to pre-release `portal-master` versions only to be able to build projects for non-GA APIs.

PR #1122, opened 2023-04-03
fix(js-toolkit-core): if run from a workspace, traverse up tree to find client-extension.yaml

bryceosterhaus:
This is to add support for multi-CX containers


izaera:
LGTM

It may end up being useful to have the workspace folder in the `Project` object, too. We'll see...

PR #1124, opened 2023-04-03
fix(portal-base): ensure that there is whitespace surrounding the svg attribute

bryceosterhaus:
Without this change, there is possibility for breaking the svg markup depending on what is provided. We might want to look at an xml parser for doing this in the future, rather than just find/replace


bryceosterhaus:
nvm, need a better fix for this

PR #1130, opened 2023-04-14
feat(npm-scripts): ignore dependency check for modules that have their own yarn.lock

bryceosterhaus:
Before this change we do a greedy check against ALL package.json files. We had thought that every package.json would need to follow the same dependency rules, but now with the introduction of client extensions, these package.json files will work within their own "workspace" and is isolated from liferay-portal. So now we check if the package.json has a sibling yarn.lock file, and if it does then we don't worry about that project.

### DIFF
--- a/.github/workflows/projects.yml
+++ b/.github/workflows/projects.yml
@@ -22,7 +22,7 @@ on:
 
 env:
     CI: true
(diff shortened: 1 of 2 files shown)


bryceosterhaus:
one of these days I will actually look into the windows npm cache issue for our ci

PR #1133, opened 2023-04-20
refactor(npm-scripts): remove webpack federation code

izaera:
I've tested it locally in an `ant all` and it didn't break anything, apparently.

I'm not going to release a version because it is just a refactor.


bryceosterhaus (approved):
❤️ LGTM

PR #1138, opened 2023-05-09
fix: babel loader and buildrc

BalrokHS:
In the latest Liferay React generator, I faced the following issues, that I was able to reproduce on different OS and setups.

- Windows 10 (Node 16)
-  MacOS (Node 16) 

The issues are: 
- Not recognising the presets, although they are installed (not recognising .babelrc)
- Dev server not starting because of the missing dependency of babel-loader. 

I added the first version, that I found it was compatible. Feel free to up/downgrade the version of babel-loader.

### DIFF
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-react-portlet/dependencies.json
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/target-react-portlet/dependencies.json
@@ -3,7 +3,8 @@
 		"@babel/cli": "^7.7.5",
 		"@babel/core": "^7.7.5",
 		"@babel/preset-env": "^7.7.6",
-		"@babel/preset-react": "^7.7.4"
+		"@babel/preset-react": "^7.7.4",
+		"babel-loader": "^8.0.0-beta.0"
 	},
 	"dependencies": {
 		"react": "16.8.6",
--- a/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/utils/modifier/npmbuildrc.js
+++ b/maintenance/projects/js-toolkit/packages/generator-liferay-js/src/utils/modifier/npmbuildrc.js
@@ -81,7 +81,11 @@ export default class extends JsonModifier {
 
 			currentRules.push({
 				test,
-				use: loader,
+				"exclude": "/node_modules/",
+				"loader": loader,
+				"options": {
+					"configFile": "../.babelrc"
+				  }
 			});
 
 			prop.set(json, 'webpack.rules', currentRules);


izaera:
Hey @BalrokHS, first of all, thanks a lot for you contribution :hugs:, however, the `generator-liferay-js` was discontinued long ago in favor of `@liferay/cli` (see the [docs](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/js-toolkit/docs/README.md) and especially [@liferay/cli's manual](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/js-toolkit/docs/manuals/liferay-cli.md) for information on how to use it).

I'm not completely against merging this PR, but given that the tool is no longer maintained I wonder if it makes sense, because if it breaks anything, we will need to roll it back, as we (the owners of this project) are probably not going to fix it, which  may lead to a series of merges and rollbacks until you or someone else from the community stabilizes it :grimacing:.

So, my proposal is that you first check if you can use `@liferay/cli` and if it is absolutely impossible for you, we can check again whether we want to merge this, let the project be forked, or any possibility that makes sense. I would like to give you a complete reply about this PR now but I'm not in charge of the backlog of the project, so I cannot provide any authoritative answer.

In any case, let's first check if you can use `@liferay/cli` better... Thx.

izaera:
Strange. That should work as long as you use the correct target platform (see https://github.com/liferay/liferay-frontend-projects/blob/master/target-platforms/README.md)

izaera:
For example, the bundler import for classnames is defined in the `@liferay/portal-7.4` target platform here -> https://github.com/liferay/liferay-frontend-projects/blob/master/target-platforms/packages/portal-7.4/config.json#L78

Without having more information or a sample project it's difficult to tell what can be happening...

BalrokHS:
Im downgrading to 7.4.3.75, and I will come back with the results. 

Question. In a remote app can I use a react router with a HashRouter ?

BalrokHS:
@izaera After a long series of tries, I got it to work with Liferay/portal 7.4.3.55-ga55 on Docker. 

I imported the following dependencies: 
- @ClayUI
- Formik
- Installed React-Router-DOM @6 and it works beautifully with HashRouter

Also tried to build and succeeded. React router does not seem to have a problem. 

(Maybe I can post my results with a guide for the community. If you agree, please show me where 😆  ).

A recommendation is to fix the documentation of Liferay portlet creation, since the top search lead to the deprecated way and can be really confusing. 

Great work with the Liferay-cli. 

In the case of the PR, I think it resolves some issues that exist, so in case you decide to continue to maintain this, feel free to review the changes and accept them or even change them 😄

izaera:
@BalrokHS Great news! :tada: 

Very happy that you made it work.

Let me find out if there's some way for you to post a blog entry :+1: .

Regarding the docs, are you referring to the ones in this project?  The READMEs that I linked?

And for the PR, I'll ask too what we can do.

Thx.

izaera:
@pablo-agulla @dsanz what do you think about merging this PR? Should we try it, or should we reject it and recover this notice -> https://github.com/liferay/liferay-frontend-projects/commit/ec73c6589519415581a808c7b827b94d9fe47e29

We used to redirect people from generator-liferay-js to liferay/cli in the past, but we had to remove the message because of the Workspace (see https://github.com/liferay/liferay-frontend-projects/commit/066213973b23e5d532ac559e6c55f5e056beab80).

Maybe it's time to recover the message (possibly hiding it when inside a workspace) and freeze this project correctly (including the README, which is a bit ambiguous).

PR #1139, opened 2023-05-24
fix(npm-scripts): glob pattern isn't greedy enough if there are other patterns that want to look deeper in the directory

bryceosterhaus:
This issue originally appeared here, 
https://liferay.slack.com/archives/C03CAD9EF/p1684869953566979?thread_ts=1684869909.630399&cid=C03CAD9EF

Essentially the issue we are running into is that our `expandGlobs` function is returning too soon and not looking deep enough 
into a directory. For example, in portal we have the globs of `['/apps/*/*', '/apps/*/*/*']`. This means we want to check for 
directories in both the 2nd and 3rd level deep from "apps". An example of this is that we would want to match both 
`apps/frontend-js/frontend-js-web` and additionally match `apps/headless/headless-builder/headless-builder-web`.

This doesn't actually work as we expected though because expandGlobs gets a postive hit for `apps/headless/headless-builder` and 
then doesn't look any deeper. The change I am making here ensures that we continue to look deeper in the file structure in case 
another matcher returns a hit.This issue originally appeared here,
https://liferay.slack.com/archives/C03CAD9EF/p1684869953566979?thread_ts=1684869909.630399&cid=C03CAD9EF

Essentially the issue we are running into is that our `expandGlobs` function is returning too soon and not looking deep enough
into a directory. For example, in portal we have the globs of `['/apps/*/*', '/apps/*/*/*']`. This means we want to check for
directories in both the 2nd and 3rd level deep from "apps". An example of this is that we would want to match both
`apps/frontend-js/frontend-js-web` and additionally match `apps/headless/headless-builder/headless-builder-web`.

This doesn't actually work as we expected though because expandGlobs gets a postive hit for `apps/headless/headless-builder` and
then doesn't look any deeper. The change I am making here ensures that we continue to look deeper in the file structure in case
another matcher returns a hit.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/utils/getWorkspaces.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/getWorkspaces.js
@@ -10,8 +10,6 @@ const expandGlobs = require('./expandGlobs');
 const findRoot = require('./findRoot');
 const log = require('./log');
 
-const IGNORE_GLOBS = ['node_modules/**'];
-
 /**
  * Returns a list of workspaces.
  *
@@ -32,8 +30,8 @@ function getWorkspaces() {
 				fs.readFileSync('package.json', 'utf8')
 			);
 
-			const projects = expandGlobs(workspaces.packages, IGNORE_GLOBS, {
-				maxDepth: 3,
+			const projects = expandGlobs(workspaces.packages, [], {
+				maxDepth: 4,
 				type: 'directory',
 			});
 
--- a/projects/npm-tools/packages/npm-scripts/src/utils/getWorkspaces.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/getWorkspaces.js
@@ -30,10 +30,22 @@ function getWorkspaces() {
 				fs.readFileSync('package.json', 'utf8')
 			);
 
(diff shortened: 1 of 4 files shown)


projects/npm-tools/packages/npm-scripts/src/utils/getWorkspaces.js
```diff
@@ -10,8 +10,6 @@ const expandGlobs = require('./expandGlobs');
 const findRoot = require('./findRoot');
 const log = require('./log');
 
-const IGNORE_GLOBS = ['node_modules/**'];
```
izaera, line 13: Mmm, I think I've seen some `node_modules` directories inside projects. We shouldn't have them because we are supposed to dedupe deps and hoist all of them, but in theory it is possible that we may need them.

Also, I believe `bin` commands need these folders to exist.

izaera: Apparently you recover the globs in the second commit, so it's all OK :+1:

izaera:
LGTM

bryceosterhaus:
Created this for the future https://issues.liferay.com/browse/LPS-185590

PR #1140, opened 2023-05-31
feat(npm-scripts): support SVG exports

izaera:
They work similar to CSS but inline the content directly instead of retrieving it from the server. This is because SVG files don't undergo any process in the server, as CSS files do, so there's no need to make two trips to the server when we can make just one.

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/utils/getExternalExportsWebpackConfigs.js
+++ b/projects/npm-tools/packages/npm-scripts/src/utils/getExternalExportsWebpackConfigs.js
@@ -79,6 +79,27 @@ module.exports = function getExternalExportsWebpackConfigs(
 				},
 			};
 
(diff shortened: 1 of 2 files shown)


projects/npm-tools/packages/npm-scripts/src/utils/getExternalExportsWebpackConfigs.js
```diff
@@ -79,6 +79,27 @@ module.exports = function getExternalExportsWebpackConfigs(
 				},
 			};
 
+			if (process.env.NODE_ENV === 'development') {
```
izaera, line 82: I've moved this to the top so that I can disable source maps for SVG files, since they don't serve any purpose.

izaera:
Looks good to @bryceosterhaus (according to a private conversation I cannot reveal because it's private :shrug:).

So I'm going to merge since it's green.

PR #1141, opened 2023-06-08
feat(js-api): add types for FDSFilter client extension

bryceosterhaus:
If this looks good, I will publish and add to portal along with the new CX


projects/js-toolkit/packages/js-api/data-set/index.ts
```diff
@@ -10,3 +10,18 @@ export interface FDSCellRendererArgs {
...
+
+export interface FDSFilterData {
+	id: string;
+	odataFilterString: string;
+	selectedData: string;
+}
+
+export interface FDSFilterArgs {
+	filter: FDSFilterData;
+	setFilter: (val: Partial<FDSFilterData>) => void;
```
izaera, line 22: Didn't know about `Partial` :+1:

izaera (approved):

PR #1143, opened 2023-06-19
fix(npm-scripts): wipe the tsbuildinfo file to ensure consistent types

bryceosterhaus:
I think this should help resolve our issues of getting inconsistent .d.ts files with `/// <reference  ...`.

I can't totally figure out why we are getting odd behavior with those references in d.ts files, but I suspect it may be 
a bug in our TS version. I ran some tests with TS on the latest version and I wasn't able to reproduce this issue, 
although it brought other issues.

Now I'm not totally stoked with this because this may increase our build time a little bit, although not significantly 
because we have our own npm-scripts cacheing mechanism. So I think we stick with this for now and then investigate 
upgrading to TS 5 later.


bryceosterhaus:
improvements after this issue will be done in https://liferay.atlassian.net/browse/LPS-188133

PR #1146, opened 2023-06-22
feat(npm-scripts): improve strategy for generating types, only generate types for modules that have changes to ts files or if their dependencies have changes

bryceosterhaus:
Here the goal is to only generate types when ts files have been changed. This is particularly helpful when running `ant format-source-current-branch`

The strategy is to diff against the liferay working branch ([master](https://github.com/liferay/liferay-portal/blob/master/portal-impl/build.xml#L669)). If there are any typescript files changed then we re-build types for that module and it's dependents.

Here is the new output:

- No changes to any files
Screenshot 2023-06-22 at 11 39 17 AM

- new `--force` option when we want to re-build all types. Will primarily only be used by FI team
Screenshot 2023-06-22 at 11 40 38 AM

- Changes to a single module's ts files. This rebuilds the changed module and any dependent modules
Screenshot 2023-06-22 at 11 42 51 AM

- `ant format-source-current-branch` when only java files have changed
Screenshot 2023-06-22 at 11 44 45 AM

- `ant format-source-current-branch` when ts files have changed
Screenshot 2023-06-22 at 11 47 13 AM

### DIFF
--- a/projects/npm-tools/packages/npm-scripts/src/scripts/types.js
+++ b/projects/npm-tools/packages/npm-scripts/src/scripts/types.js
@@ -9,10 +9,40 @@ const getTypeScriptBuildOrder = require('../typescript/getTypeScriptBuildOrder')
 const getTypeScriptDependencyGraph = require('../typescript/getTypeScriptDependencyGraph');
 const runTSC = require('../typescript/runTSC');
 const findRoot = require('../utils/findRoot');
+const getLiferayWorkingBranch = require('../utils/getLiferayWorkingBranch');
 const getMergedConfig = require('../utils/getMergedConfig');
+const git = require('../utils/git');
 const log = require('../utils/log');
 
-async function types() {
+function tsDiffFromWorkingBranch() {
+	const upstream = getLiferayWorkingBranch();
+
+	const mergeBase = git('merge-base', 'HEAD', upstream);
+
+	const diff = git('diff', mergeBase, '--name-only');
+
+	const changedFiles = diff
+		.split('\n')
+		.filter(
+			(filePath) =>
+				filePath.endsWith('.d.ts') ||
+				filePath.endsWith('.ts') ||
+				filePath.endsWith('.tsx')
+		)
+		.join('\n');
+
+	if (!changedFiles.length) {
+		log(
+			`There are no modified typescript files on the current branch compared to working branch (${upstream})`
+		);
+	}
+
+	return changedFiles;
+}
+
+async function types(...args) {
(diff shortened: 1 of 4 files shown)


projects/npm-tools/packages/npm-scripts/src/scripts/types.js
```diff
@@ -9,10 +9,40 @@ const getTypeScriptBuildOrder = require('../typescript/getTypeScriptBuildOrder')
...
+const getLiferayWorkingBranch = require('../utils/getLiferayWorkingBranch');
 const getMergedConfig = require('../utils/getMergedConfig');
+const git = require('../utils/git');
 const log = require('../utils/log');
 
-async function types() {
+function tsDiffFromWorkingBranch() {
+	const upstream = getLiferayWorkingBranch();
+
+	const mergeBase = git('merge-base', 'HEAD', upstream);
```
izaera, line 20: Isn't this going to be `master` always? :thinking: 

Under which circumstance would it be something different? :thinking:

projects/npm-tools/packages/npm-scripts/src/scripts/types.js
```diff
@@ -9,10 +9,40 @@ const getTypeScriptBuildOrder = require('../typescript/getTypeScriptBuildOrder')
...
+
+	const diff = git('diff', mergeBase, '--name-only');
```
izaera, line 22: Will this return modified files in the working and stage areas?

projects/npm-tools/packages/npm-scripts/src/scripts/types.js
bryceosterhaus: It's almost always master, but it depends if someone sets `git.working.branch.name` in their portal.properties

projects/npm-tools/packages/npm-scripts/src/scripts/types.js
izaera: Yep :+1:

projects/npm-tools/packages/npm-scripts/src/scripts/types.js
izaera: Ah, true :+1:

izaera:
LGTM

izaera:
One more question (I missed the screenshots before approval). What happens with `ant format-source-local-changes`? Will it work correctly? :thinking:

bryceosterhaus:
Hmm. I haven't run that before. I will test and see

PR #1156, opened 2023-07-26
feat: ignore node_modules when minifying files and add new eslint rule for dependency check

bryceosterhaus:


projects/npm-tools/packages/npm-scripts/src/utils/minify.js
```diff
@@ -27,6 +27,7 @@ const MINIFY_GLOBS = [
 	path.posix.join(BUILD, '**', '*.js'),
 	'!*-min.js',
 	'!*.min.js',
+	'!node_modules',
```
bryceosterhaus, line 30: This was causing some crazy issues of build time in our build of commerce-js-web. Essentially we were minifying thousands of files and had hundreds of errors that were being swallowed, so I think it makes sense to ignore minifying and leave that to the library author. Also, I don't even think those files are loaded in the browser ever, most were devDependencies

bryceosterhaus:
gonna merge and try a release and add to a portal PR. Everything seems to work locally 🤞

PR #1157, opened 2023-08-03
feat(js-api): add FDS CX interfaces with `HTMLBuilder` naming convention

markocikos:
### References

- [LPS-189354 Standardize renderer APIs](https://issues.liferay.com/browse/LPS-189354)

### What is the goal of this PR?

In the PR, we are introducing the `HTMLBuilder` naming convention to public API:
- `*Renderer` is not quite right, as this is a property of a "Renderer" in DXP.
- `*Component` would not be right, as the function is not [used](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-components-web/src/main/resources/META-INF/resources/ClientExtension.tsx#L21) as a React component. It's used as a plain old function. 
- No suffix, like `FDSFilter`, might be confusing, as there are objects with the same name in DXP.
- We already introduced `htmlBuilder` as an argument of the [ClientExtension component](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-components-web/src/main/resources/META-INF/resources/ClientExtension.tsx#L10).

We are only adding API, and not renaming, so that the existing contracts don't break. After we migrate usages in DXP, we can delete old API here.

@bryceosterhaus @izaera What do you think?

### DIFF
--- a/projects/js-toolkit/packages/js-api/data-set/index.ts
+++ b/projects/js-toolkit/packages/js-api/data-set/index.ts
@@ -3,6 +3,14 @@
  * SPDX-License-Identifier: LGPL-3.0-or-later
  */
 
+export interface FDSTableCellHTMLElementBuilderArgs {
+	value: boolean | number | string | object | [];
+}
+
+export interface FDSTableCellHTMLElementBuilder {
+	(args: FDSTableCellHTMLElementBuilderArgs): HTMLElement;
+}
+
 export interface FDSCellRendererArgs {
 	value: boolean | number | string | object | [];
 }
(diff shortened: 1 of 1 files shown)


izaera:
Only thing that came to mind after writing the :ok_hand: ... 

It's not really an `HTMLBuilder` but an `HTMLElementBuilder`. I don't think it matters so much because both concepts are equivalent but if we plan to have builders that return strings (as opposed to `HTMLElement`s) in the future it can be worth reserving the name `HTMLBuilder` for stuff that returns strings and use `HTMLElementBuilder` for this. Maybe... 🤷

bryceosterhaus:
I'm pretty flexible either way, but if I had to lean one way I would say go with `HTMLElementBuilder` since we are explicitly expecting it to return a single HTMLElement

markocikos:
Good point! I'll push the change with rename.

markocikos:
@bryceosterhaus Do we need to manually publish `@liferay/js-api`? Any special process to do this? Looking at [versions](https://www.npmjs.com/package/@liferay/js-api?activeTab=versions), I guess it would be 0.4.0-pre.0.

bryceosterhaus:
Yep we would need to manually publish, I can do so later this afternoon or tomorrow

bryceosterhaus:
@markocikos  https://github.com/liferay/liferay-frontend-projects/releases/tag/js-api%2Fv0.4.0-pre.0

PR #1158, opened 2023-08-04
fix(amd-loader): don't emit the nonce if it is empty

izaera:


izaera:
I'm :ninja: merging this as soon as it is green because I need to release a new version

PR #1163, opened 2023-08-21
fix(liferay-theme-tasks): Fix sass version to 1.64.1

pat270:
https://liferay.atlassian.net/browse/LPP-50503

This fixes the Sass version to 1.64.1 to avoid the log function errors introduced in Sass@1.65.1. I think it would be better if we could review Sass minor and patch versions before passing it on to our users.

@ethib137 @dsanz

### DIFF
--- a/projects/js-themes-toolkit/packages/liferay-theme-tasks/package.json
+++ b/projects/js-themes-toolkit/packages/liferay-theme-tasks/package.json
@@ -36,7 +36,7 @@
 		"plugin-error": "^1.0.1",
 		"portfinder": "^1.0.20",
 		"resolve": "^1.15.1",
-		"sass": "^1.30.0",
+		"sass": "1.64.1",
 		"tar-fs": "^1.16.3",
 		"through2": "^2.0.0",
 		"tiny-lr": "^1.1.1",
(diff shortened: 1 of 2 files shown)


ethib137:
Fixing the version does provide the benefit of not unknowingly introducing new errors, but it also has it's own drawbacks. We miss out on automatic fixes, sometimes of security issues (granted, this is probably not an issue for a sass processor). How would we make sure to keep the version updated? Do we need to keep it updated? Are there any other instances where we have a fixed the version of something like this? Do we already have a process for consistently doing manual version updates?

pat270:
I don't have any ideas for making sure Sass gets updated other than periodic reminders. We could use the `~` symbol which will only update future patch versions (1.64.xx). Blindly accepting new versions from the cloud also come with vulnerabilities, packages get compromised (e.g., ua-parser-js).

ethib137:
Maybe to make sure we don't forget to make updates, we could add a github action that creates an issue once a month to test and update to the latest sass version. https://docs.github.com/en/actions/managing-issues-and-pull-requests/scheduling-issue-creation

ethib137:
Thanks @bryceosterhaus , that makes sense. Let me know if anything else is needed to get this merged.

bryceosterhaus:
@ethib137 do you also need a new version released ASAP?

ethib137:
Yes please @bryceosterhaus . We have clients asking for an official solution and having this released will provide that.

bryceosterhaus:
Released: https://github.com/liferay/liferay-frontend-projects/releases/tag/liferay-theme-tasks%2Fv11.5.3

I believe the theme generator should automatically pull in this latest version

PR #1172, opened 2023-09-27
feat(js-api)!: improve FDS filter CX interfaces

izaera:
See https://liferay.atlassian.net/browse/LPS-197299

This is a breaking change but we have not yet released any GA version of @liferay/js-api so there shouldn't be any problem.

Note that I've changed the names of the exported functions and types to meaningful (from the POV of the customer) ones. 

This is because I think we should not tie API names to portal internal stuff. In this case I've left the `HTMLBuilder` concept around because it makes sense, but I've renamed the names of the methods and they don't match the internal ones we use in portal any more.

The rationale for not aligning API and internal names is simple: internal names are subject to change and do change, in fact, as we evolve the product, whereas API names must remain stable forever unless we want to be continuously breaking customers code.

### DIFF
--- a/projects/js-toolkit/packages/js-api/data-set/index.ts
+++ b/projects/js-toolkit/packages/js-api/data-set/index.ts
@@ -13,14 +13,6 @@ export interface FDSTableCellHTMLElementBuilder {
 	(args: FDSTableCellHTMLElementBuilderArgs): HTMLElement;
 }
 
-export interface FDSCellRendererArgs {
-	value: boolean | number | string | object | [];
-}
-
-export interface FDSCellRenderer {
-	(args: FDSCellRendererArgs): HTMLElement;
-}
-
 // Frontend data set filter
 
 export interface FDSFilterData<T> {
--- a/projects/js-toolkit/packages/js-api/data-set/index.ts
+++ b/projects/js-toolkit/packages/js-api/data-set/index.ts
@@ -45,7 +45,7 @@ export interface FDSFilterDescriptionBuilder<T> {
 }
 
 export interface FDSFilter<T> {
-	buildFilterDescription: FDSFilterDescriptionBuilder<T>,
-	buildHTMLElement: FDSFilterHTMLElementBuilder<T>,
-	buildODataQuery: FDSFilterODataQueryBuilder<T>,
+	buildFilterDescription: FDSFilterDescriptionBuilder<T>;
+	buildHTMLElement: FDSFilterHTMLElementBuilder<T>;
+	buildODataQuery: FDSFilterODataQueryBuilder<T>;
 }
--- a/projects/js-toolkit/packages/js-api/data-set/index.ts
+++ b/projects/js-toolkit/packages/js-api/data-set/index.ts
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: LGPL-3.0-or-later
  */
 
+// Frontend data set cell renderer
+
 export interface FDSTableCellHTMLElementBuilderArgs {
 	value: boolean | number | string | object | [];
 }
@@ -19,26 +21,31 @@ export interface FDSCellRenderer {
 	(args: FDSCellRendererArgs): HTMLElement;
 }
 
-export interface FDSFilterData {
-	id: string;
-	odataFilterString: string;
-	selectedData: string;
+// Frontend data set filter
+
+export interface FDSFilterData<T> {
+	selectedData: T;
 }
 
-export interface FDSFilterArgs {
-	filter: FDSFilterData;
-	setFilter: (val: Partial<FDSFilterData>) => void;
+export interface FDSFilterHTMLElementBuilderArgs<T> {
+	filter: FDSFilterData<T>;
+	setFilter: (partialFilter: Partial<FDSFilterData<T>>) => void;
 }
 
-export interface FDSFilter {
-	(args: FDSFilterArgs): HTMLElement;
+export interface FDSFilterHTMLElementBuilder<T> {
+	(args: FDSFilterHTMLElementBuilderArgs<T>): HTMLElement;
 }
 
-export interface FDSFilterHTMLElementBuilderArgs {
-	filter: FDSFilterData;
-	setFilter: (val: Partial<FDSFilterData>) => void;
+export interface FDSFilterODataQueryBuilder<T> {
+	(selectedData: T): string;
 }
 
-export interface FDSFilterHTMLElementBuilder {
-	(args: FDSFilterHTMLElementBuilderArgs): HTMLElement;
+export interface FDSFilterDescriptionBuilder<T> {
+	(selectedData: T): string;
+}
+
+export interface FDSFilter<T> {
+	buildFilterDescription: FDSFilterDescriptionBuilder<T>,
+	buildHTMLElement: FDSFilterHTMLElementBuilder<T>,
+	buildODataQuery: FDSFilterODataQueryBuilder<T>,
 }
--- a/projects/js-toolkit/packages/js-api/data-set/index.ts
+++ b/projects/js-toolkit/packages/js-api/data-set/index.ts
@@ -37,7 +37,7 @@ export interface FDSFilterDescriptionBuilder<T> {
 }
 
 export interface FDSFilter<T> {
-	buildFilterDescription: FDSFilterDescriptionBuilder<T>;
-	buildHTMLElement: FDSFilterHTMLElementBuilder<T>;
-	buildODataQuery: FDSFilterODataQueryBuilder<T>;
+	descriptionBuilder: FDSFilterDescriptionBuilder<T>;
+	htmlElementBuilder: FDSFilterHTMLElementBuilder<T>;
+	oDataQueryBuilder: FDSFilterODataQueryBuilder<T>;
 }


izaera:
After we settle on API names, I'll send a PR for `liferay-portal` to be updated with the new contract.

projects/js-toolkit/packages/js-api/data-set/index.ts
```diff
@@ -19,26 +21,31 @@ export interface FDSCellRenderer {
 	(args: FDSCellRendererArgs): HTMLElement;
```
izaera, line 21: I was tempted to add a `buildHTMLElement` symbol to `FDSCellRenderer` for simmetry, but refrained from doing so because:

1. It would be a breaking change and, even though we are still in experiment phase, I prefer not to do it.
2. It makes the implementation of a cell renderer a bit more cumbersome, because instead of exporting a default function, the user would need to export an object containing a reference to the function, which seems a bit convoluted.
3. If in the future we need to export more things for a cell renderer (as is the case with the two extra methods in the FDS filter) we can always mimic the structure of `FDSFilter` and add a runtime check in liferay-portal so that if the CX exports a bare function we wrap it into an object with a `buildHTMLElement` property. That way we don't break anything while still evolving the API.

projects/js-toolkit/packages/js-api/data-set/index.ts
```diff
@@ -19,26 +21,31 @@ export interface FDSCellRenderer {
...
+export interface FDSFilterHTMLElementBuilderArgs<T> {
+	filter: FDSFilterData<T>;
+	setFilter: (partialFilter: Partial<FDSFilterData<T>>) => void;
 }
 
-export interface FDSFilterArgs {
-	filter: FDSFilterData;
-	setFilter: (val: Partial<FDSFilterData>) => void;
+export interface FDSFilterHTMLElementBuilder<T> {
+	(args: FDSFilterHTMLElementBuilderArgs<T>): HTMLElement;
```
markocikos, line 36: We have this [IHTMLElementBuilder interface locally in DXP](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/renderer/types.ts#L6-L8). It probably belongs here, something like:
```ts
export interface HTMLElementBuilder {
	(args: any): HTMLElement;
}

export interface FDSTableCellHTMLElementBuilder extends HTMLElementBuilder {
	(args: FDSTableCellHTMLElementBuilderArgs): HTMLElement;
}

export interface FDSFilterHTMLElementBuilder<T> extends HTMLElementBuilder {
	(args: FDSFilterHTMLElementBuilderArgs<T>): HTMLElement;
}
```

bryceosterhaus (approved):
LGTM! Names look reasonable to me


> I think we should not tie API names to portal internal stuff

👍 definitely agree


cc @markocikos, just gonna mention you for opinions as well

projects/js-toolkit/packages/js-api/data-set/index.ts
```diff
@@ -19,26 +21,31 @@ export interface FDSCellRenderer {
...
+export interface FDSFilterDescriptionBuilder<T> {
+	(selectedData: T): string;
 }
 
-export interface FDSFilterHTMLElementBuilder {
-	(args: FDSFilterHTMLElementBuilderArgs): HTMLElement;
+export interface FDSFilter<T> {
+	buildFilterDescription: FDSFilterDescriptionBuilder<T>;
+	buildHTMLElement: FDSFilterHTMLElementBuilder<T>;
+	buildODataQuery: FDSFilterODataQueryBuilder<T>;
```
markocikos, line 50: This is basically a Renderer. An object containing all necessary information to render something. We know that [the CX renderer](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/renderer/types.ts#L10-L16) should have an `htmlElementBuilder`. In case of filters, it requires two more properties. This, to me, implies something like this:
```ts
export interface IClientExtensionRenderer {
	externalReferenceCode?: string;
	htmlElementBuilder?: IHTMLElementBuilder;
	name?: string;
	type: 'clientExtension';
	url?: string;
}

export interface FDSFilterClientExtensionRenderer extends IClientExtensionRenderer {
	descriptionBuilder: FDSFilterDescriptionBuilder<T>;
    htmlElementBuilder: FDSFilterHTMLElementBuilder<T>;
    oDataQueryBuilder: FDSFilterODataQueryBuilder<T>;
}
```

markocikos (commented):
@izaera Please see comments inline. The two comments related of [types.ts in DXP](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/renderer/types.ts) makes me think that the whole content of that file should really be here. Maybe in a separate file, not in `data-set` dir, as they are not directly related to FDS. What do you think?

projects/js-toolkit/packages/js-api/data-set/index.ts
izaera: Done -> https://github.com/liferay/liferay-frontend-projects/pull/1172/commits/3e2f0863498737fbff31010546298351098c7e63

izaera:
> The two comments related of [types.ts in DXP](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/renderer/types.ts) makes me think that the whole content of that file should really be here. Maybe in a separate file, not in data-set dir, as they are not directly related to FDS. What do you think?

At first sight I would have bought your argument, but  then I reflected a bit more and now I'm not sure...

On the one hand, if this project was internal to portal (or to any other project) I would do what you say. But my intuition tells me that we should treat `@liferay/js-api` more as a spec than as code. In other words, think about the HTTP spec: it explains how things should work not how they should be implemented.

Modeling a `FDSCellRenderer` (or an `FDSFilter`) as an extension of an abstract type `Renderer` is a "meta-implementation" detail. I added "meta" to convey that it's not that we are implementing something, because these are only types and we are modelling, but we are imposing a restriction on the model itself: we are forcing how to "implement" the model.

Seem from the POV of the portal it may make a lot of sense to have a common `Renderer` type, but seen from the POV of a customer who's writing a cell renderer, why should he care?

I mean, having a clear spec of what is a `FDSCellRenderer` is more than enough, no matter if it could share fields with or look like an `FDSFilter` and no matter if the portal treats them internally as the same type object, which makes the sharing of the fields not a coincidence but a real thing.

Also, think what would happen if in the future we decide to change how `FDSCellRenderer`s are implemented but want to leave `FDSFilter` untouched. Or imagine that we stop treating these two types as subtypes of the same entity `Renderer`...

That said, there is an scenario where a supertype `Renderer` could make sense in `@liferay/js-api` and it would be the case of customers making complex CX packs and writing code to manage all of them. In that case it could be handle for a customer to have a `Renderer` type but even in this case we could argue that if they want it, they can create it in their project maybe, since that is possible in TypeScript (not in other languages, but we are lucky this time :sweat_smile: ).

So to sum up:

1. I would make `@liferay/js-api` minimal in the sense of not creating complex hierarchies of types and not creating any type that is not absolutely necessary.
2. I would treat each CX spec as isolated from the others, even if they share fields and/or are the very same thing in portal.

I'm 100% sure of :one: .

I'm open to discussing :two:  a bit more since I admit there's some risk of bloating `@liferay/js-api` with very similar types. I suppose that the problem here is placing the border line between what definitely lies in :one: and what doesn't...

izaera:
> I'm open to discussing 2️⃣ a bit more since I admit there's some risk of bloating @liferay/js-api with very similar types. I suppose that the problem here is placing the border line between what definitely lies in 1️⃣ and what doesn't...

The good news is that we don't need to create supertypes from the very beginning, we can introduce them in the future if we see they finally make sense and we are committed to them in the long term.

projects/js-toolkit/packages/js-api/data-set/index.ts
izaera: This example shows what I said below: why should the `IClientExtensionRenderer` type be in a spec like `@liferay/js-api` at all? It doesn't make sense...

```typescript
export interface IClientExtensionRenderer {
    externalReferenceCode?: string;
    htmlElementBuilder?: IHTMLElementBuilder;
    name?: string;
    type: 'clientExtension';
    url?: string;
}
```

The customer doesn't need to care about `externalReferenceCode` or `type`, for example. We want to hide that from him.

projects/js-toolkit/packages/js-api/data-set/index.ts
izaera: This falls in part :two: of my summary in the big comment below, most probably.

However, I would refrain for now from creating the `HTMLElementBuilder` type, even though it is very probable that we may add it in the future.

But until I see it used in more CX types and people have started implementing and using them, I prefer not to add it since it is not absolutely necessary :shrug:

projects/js-toolkit/packages/js-api/data-set/index.ts
markocikos: Yeah, you definitely have a good point there!

With that in mind, my only nit on this code would be names of properties, for consistency. Could we have `descriptionBuilder` instead of `buildFilterDescription`?

projects/js-toolkit/packages/js-api/data-set/index.ts
markocikos: Makes sense, at least until rule of three kicks in :+1:

markocikos:
A lot of good points there, @izaera! :sparkles:  I don't much to add, except that I completely agree.

projects/js-toolkit/packages/js-api/data-set/index.ts
izaera: Done -> https://github.com/liferay/liferay-frontend-projects/pull/1172/commits/f9e7f8192683329d997ed9d39a49c22dcb01b7bf

markocikos (approved):
@izaera LGTM! :sparkles:

PR #1176, opened 2023-10-02
fix(js-toolkit): add warnings for when creating, building, and deploying projects within a liferay workspace

bryceosterhaus:
Adds some console messages for when using liferay-cli in a liferay workspace. Also remove support for fdsCellRenderer and themeSpritemap


izaera:
LGTM

PR #1178, opened 2023-10-13
feat(npm-scripts)!: update tools for react 18

bryceosterhaus:
This PR is a work in progress for our upgrade to react 18. I will make pre-releases from this branch to test in DXP


bryceosterhaus:
This should be good for release. Once I am able to get CI to pass in the liferay-portal [PR](https://github.com/liferay-frontend/liferay-portal/pull/3717), I will make a proper release of this and add to that PR.

matuzalemsteles:
@bryceosterhaus  Oh, it's great to see that we are moving towards this, just a doubt if we had other problems, for example the entry of React 17.x brought changes to the global event system when preventing and for propagation 18.x introduced the new system competitor but still with incremental adoption, in the tests did you see anything different like the rendering behavior? side effects, for example in useEffect calls? Maybe this won't be a problem if you're not using StrictMode... Just curious if we can help with this on the Clay side, I remember having some Clay reports not working very well in React.js.

bryceosterhaus:
@matuzalemsteles still working on diagnosing CI errors in a bundle. So far I haven't seen any failures in the UI, but I need QA to do a full review of [this pr](https://github.com/liferay-frontend/liferay-portal/pull/3717), but CI has been a headache.

Since we aren't using StrictMode or any sort of server-side react code, I don't expect issues as that is where the bulk of updates have been for react.

PR #1185, opened 2023-12-18
feat(js-api): add types for Editor Config Contributor client extension

markocikos:
### References

- [LPS-187309 Apply editor config contributor client extensions to CKEditor](https://issues.liferay.com/browse/LPS-187309)
- Related to https://github.com/liferay-frontend/liferay-portal/pull/3902

### What is the goal of this PR?

Adding public types for Editor Config Contributor client extension.

Notes:
- I intentionally named this in a generic way, `EditorConfigTransformer` and not  `EditorConfigContributor`. We plan to add a different CET in the future, that will not use the Java `EditorConfigContributor` framework in DXP. That CET will still be able to use this API, as both will do the same config transformation in the end.
- I don't think we want to manage anything more specific than `[key: string]: any;` for config structure. This is based on the third-party APIs and is likely to change with time. Being so generic will also enable us to be consistent with APIs for CKEditor 5 or other editors. I added in commited docs link to offical docs for the `config` object.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -6,6 +6,6 @@ jobs:
     triage:
         runs-on: ubuntu-latest
         steps:
-            - uses: actions/labeler@main
+            - uses: actions/labeler@v4
               with:
                   repo-token: '${{ secrets.GITHUB_TOKEN }}'
--- a/projects/js-toolkit/packages/js-api/editor/index.ts
+++ b/projects/js-toolkit/packages/js-api/editor/index.ts
@@ -1,0 +1,19 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+/**
+ * Editor Transformer
+ *
+ * For CKEditor 4, see config API:
+ * https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html
+ */
+
+export interface EditorConfigTransformer<T> {
+	(args: T): T;
+}
+
+export interface EditorTransformer<T> {
+	editorConfigTransformer: EditorConfigTransformer<T>;
+}


projects/js-toolkit/packages/js-api/editor/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * For CKEditor 4, see API of CKEDITOR.config:
+ * https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html
+ */
+
+export interface EditorConfig {
+	[key: string]: any;
+}
+
+export interface EditorConfigTransformer {
```
izaera, line 17: Maybe we can specify a template parameter `T` here (like in https://github.com/liferay/liferay-frontend-projects/blob/6008120d95bc52f6d15cdaf1366c49e466147eb7/projects/js-toolkit/packages/js-api/data-set/index.ts#L39) so that users are able to type their editor configurations more deeply.

Not sure, however, if that would help or would be used. I would need to see a CX in action using this API to see if it would make sense...

projects/js-toolkit/packages/js-api/editor/index.ts
```diff
@@ -0,0 +1,19 @@
...
+	(args: EditorConfig): EditorConfig;
```
izaera, line 18: Also, I would consider exporting an object instead of a raw function (from the CX impl) unless you are 200% :sweat_smile:  sure users will never need to export metadata related to the function or even more functions.

Remember that I had to change that in `FDSFilter` when I added more functions to the CX as a consequence of evolving the CX features. So given that returning the function inside an object instead of returning it directly is not very cumbersome for the user and allows for much more future flexibility, I think we should consider it.

markocikos: Hmm the scope of customization that we always considered is the `config` object. This matches `EditorConfigContributor`s in DXP, through which we can only modify the `config` object as well. As long as that is the scope, I'd be 200% confident we will never need to export more. But! There is a lot more to [CKEditor API](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR.html) other than config object. You can modify and override its implementation, like we are [doing in plugins](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/_diffs/plugins/ballooneditor/plugin.js#L231). If we want to be open to support those kinds of customizations later on, it would make sense to set an object return value immediately, so it's compatible later on. In that case, this would better be named something like `EditorTransformer`.

projects/js-toolkit/packages/js-api/editor/index.ts
markocikos: I though about this and decided against it, but now I'm not that sure. The reason is that `config` [definition](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html) is massive, meaning full implementation of T would be massive. I don't think this is something we would want to maintain, and I can't imagine any customers would. We would only end up using `any` as T, so my reasoning was that `[key: string]: any;` is better than nothing. **But!** I just found that CKeditor 5 has [TS interface](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-core/src/editor/editorconfig.ts) that could be used for that purpose, so it does make sense to add it. Also, some customers may be crazy enough to implement this fot CKEditor 4 :man_shrugging:

markocikos:
@izaera I force pushed changes, please take a look. In addition, do you know what causes the 
`Pull Request Labeler / triage` check failure? The error is quite cryptic to me, `found unexpected type for label '10.x' (should be array of config options)`

izaera:
>do you know what causes the Pull Request Labeler / triage check failure?

Looks like an internal issue in GH actions. I'd rerun it again.

.github/workflows/labeler.yml
```diff
@@ -1,6 +1,11 @@
 name: 'Pull Request Labeler'
+
 on:
-    - pull_request_target
+    pull_request_target:
```
izaera, line 4: This is what is making the CI task fail, right?

izaera:
LGTM!

.github/workflows/labeler.yml
markocikos: No, it still failed after that :disappointed:   I'm trying to figure out what's going on, currently my best guess is that the latest version of https://github.com/actions/labeler requires different configuration.

markocikos: OK, I think I have this figured out:
- Check is failing because v5 introduced [new format for paths](https://github.com/marketplace/actions/labeler?version=v5.0.0). Our version is currently set to `main`, which worked fine until two weeks ago when v5 released.
- Check is still failing when I make the version fixed in PR, because it is still running actions/labeler@main

@izaera Any objections to merge the chore commit that sets version fixed to v4?

izaera: :shrug: I've never touched that part. If it works, push it :sweat_smile:

markocikos: I think there is no way for me to know, before merging. It'll be fine, I'm sure

markocikos:
@izaera Can you release new version of @liferay/js-api? I can try, but I'm not sure if I have permissions, or if there are and hidden quirks of this repo.

izaera:
let me check :+1:

izaera:
It's as explained here -> https://github.com/liferay/liferay-frontend-projects/blob/master/projects/js-toolkit/CONTRIBUTING.md

I'll try it now bc it's been long since I don't release a version of this project, but IIRC there's nothing special with this project. Only that we are releasing pre-versions until we decide to make it effective and start to maintain proper versioning of the contracts.

.github/workflows/labeler.yml
markocikos: Leaving a note for @bryceosterhaus here. Setting of version from `main` to `v4` fixes the labeler issue, but I'm not sure if it goes again some wider dependencies convention in this repo. If yes, then we can spend a bit of time updating configuration format to the latest version.

projects/js-toolkit/packages/js-api/editor/index.ts
dsanz: Definitely we shall not maintain the various `T` that could apply here, as we are not editor manufacturers. However, in case `T` existed out there as a type, then this could be leveraged by CX implementors to provide strong typing on their configurations. `any` is always at hand for those who don't want this.

PR #1187, opened 2024-01-09
feat(npm-scripts): detect root dir when package-lock.json present

izaera:
This is needed so that liferay-portal's Playwright tests can use npm-scripts without the `--experimental-standalone` flag.


bryceosterhaus (approved):
merge away

PR #1190, opened 2024-01-18
feat(npm-scripts): detect playwright folder as root

izaera:
This is needed so that playwright SF uses its own set of rules.


izaera:
Going to :ninja:-merge this as soon as it is green

PR #1195, opened 2024-03-04
fix(npm-scripts): ignore url options in css as we don't support loading files other than js and css

bryceosterhaus:
When adding an "exports" in a dxp module, we want to at times allow for setting css files. This is the case for `leaflet/dist/leaflet.css` which is used in DXP. However, this fails because it is trying to load an external .png file. By setting url to false, we disable webpack from processing url references.


bryceosterhaus:
gonna merge and release. this is pretty trivial as it only impacts .css files within `exports: []`

Current uses

Screenshot 2024-03-04 at 11 37 15 AM

izaera:
LGTM

PR #1196, opened 2024-03-04
fix: handle module aliases that start with "." correctly

izaera:
See https://liferay.atlassian.net/browse/LPD-19292


bryceosterhaus (commented):
LGTM!

PR #1204, opened 2024-04-18
feat(npm-scripts): provide a flag to disable sass

izaera:
I don't need this to be merged fast, but:

1. It makes sense since we already have disable flags for the rest of the toolling.
2. It will be handy when we start using the native sass compiler.


izaera:
The failure is due to the node version we use in Windows CI, not because of this PR, so going to merge...

PR #1247, opened 2025-05-08
feat(js-api): add itemData to FDSTableCellHTMLElementBuilderArgs

chestofwonders:
Issue: https://liferay.atlassian.net/browse/LPD-50122

<h1>Allow accessing all row Content in the Cell Renderer Client Extension</h1>

The goal of this PR is to provide support for `itemData` arg in `js-api` , which is needed for Frontend data set cell renderer to access all row items.

### DIFF
--- a/projects/js-toolkit/packages/js-api/data-set/index.ts
+++ b/projects/js-toolkit/packages/js-api/data-set/index.ts
@@ -6,6 +6,7 @@
 // Frontend data set cell renderer
 
 export interface FDSTableCellHTMLElementBuilderArgs {
+	itemData?: object;
 	value: boolean | number | string | object | [];
 }
 


bryceosterhaus:
Hey, couple of things:

- Not sure why my commit is in this PR, I think it should just be a single commit?
- Commit message should be `feat(js-api): add itemData to FDSTableCellHTMLElementBuilderArgs`

### After merging...
- [Release steps are here](https://github.com/liferay/liferay-frontend-projects/blob/master/projects/js-toolkit/CONTRIBUTING.md#releasing-new-versions)
- ^ not sure if you have permissions to release, as they aren't always set properly, if you don't just ping me and I can release

izaera:
Why has this passed all CI routines? They used to complain about commit messages... :thinking:

bryceosterhaus:
@izaera ci only validates the PR title, which is used to generate release notes

izaera:
> I've changed the title. 

You need to change the commit message too. See this guide on semantic commits -> https://github.com/liferay/liferay-frontend-projects/blob/ff76b3d915f8e8bf5a6c6476a6d361d337ea011a/guidelines/general/commit_messages.md?plain=1#L7

> No idea why there's an extra commit, my changes are just one commit 😕

I'm not sure, but I'd say you need to rebase on top of master. Apparently you rebased this PR on a commit that is the same as the last one in master but with a different hash (so maybe you took it from a PR or something like that... :shrug: ):

This is the hash of the commit in master:

image

And this is the hash in your PR:

image

> So, how is process to merge? It's simply a matter of force-push the branch into master?

We use merge commits (not rebase ones). It can be done from GitHub's UI (we can do it if you want). Only that we need to sort out the extra commit thingie and the commit message :grimacing: .

PR #1260, opened 2026-04-29

igor-franca:
## Summary

Two accessibility fixes in `aui-scheduler` flagged by axe DevTools on the Liferay calendar (Jira: [LPD-87992](https://liferay.atlassian.net/browse/LPD-87992) / [LPP-63881](https://liferay.atlassian.net/browse/LPP-63881)).

### `aui-scheduler-base-event` — `role="button"` on event template

The `EVENT_NODE_TEMPLATE` renders `aria-expanded="false"` and `aria-haspopup="dialog"` on a bare `<div>`, where ARIA disallows both attributes. The element is already behaviourally a button (`tabindex="0"`, click + keyboard handlers, opens a dialog popover). Adding `role="button"` legalises both ARIA attributes without changing visuals.

### `aui-scheduler-view-day` — populate and label header cells

Two related issues:

1. The headerView (the all-day events strip, rendered as a nested `SchedulerTableView`) never had its `<th class="scheduler-view-table-header-day">` cells populated. `syncDaysHeaderUI` only runs from `_uiSetDate`, and the headerView's date is never set during initial render. Result: seven empty table headers in the week and day views.
2. The day view's own day-of-week `<a>` elements had no robust accessible name. Downstream stylesheets sometimes shrink the weekday `<span>` with `font-size: 0` to abbreviate it for narrow viewports, which strips the accessible name on platforms that drop zero-sized text from the accessibility tree.

Calls `instance.headerView.syncDaysHeaderUI()` after `render()` and inside `_uiSetDate` so the all-day strip's day-name `<th>` cells stay populated, and sets `aria-label` on each header anchor with the localized full date so the accessible name survives any visual abbreviation.

### DIFF
--- a/third-party/projects/alloy-ui/src/aui-scheduler/js/aui-scheduler-base-event.js
+++ b/third-party/projects/alloy-ui/src/aui-scheduler/js/aui-scheduler-base-event.js
@@ -357,8 +357,8 @@ var SchedulerEvent = A.Component.create({
         'colorSaturationFactor', 'titleDateFormat', 'visible', 'disabled'],
 
     prototype: {
-        EVENT_NODE_TEMPLATE: '<div aria-expanded="false" aria-haspopup="dialog" class="' + CSS_SCHEDULER_EVENT + '" tabindex="0">' + '<div class="' +
-            CSS_SCHEDULER_EVENT_TITLE +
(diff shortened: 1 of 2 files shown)


dsanz:
see https://github.com/liferay/liferay-frontend-projects/pull/1262

PR #1262, opened 2026-05-12

dsanz:
Followup of https://github.com/liferay/liferay-frontend-projects/pull/1260 where:

- Minor changes to reuse an existing function are made
- Updated yui3 version

### DIFF
--- a/third-party/projects/alloy-ui/src/aui-scheduler/js/aui-scheduler-view-day.js
+++ b/third-party/projects/alloy-ui/src/aui-scheduler/js/aui-scheduler-view-day.js
@@ -890,9 +890,9 @@ var SchedulerDayView = A.Component.create({
          */
         syncDaysHeaderUI: function() {
             var instance = this;
-            var locale = instance.get('scheduler').get('locale');
             var viewDate = instance.get('scheduler').get('viewDate');
-            var formatter = instance.get('headerDateFormatter');
+            var headerFormatter = instance.get('headerDateFormatter');
+            var navigationFormatter = instance.get('navigationDateFormatter');
             var todayDate = instance.get('scheduler').get('todayDate');
 
             instance.colHeaderDaysNode.all('a').each(
@@ -903,13 +903,9 @@ var SchedulerDayView = A.Component.create({
                         CSS_SCHEDULER_TODAY_HD, !DateMath.isDayOverlap(columnDate, todayDate));
 
                     columnNode.setAttribute(
-                        'aria-label',
-                        A.DataType.Date.format(columnDate, {
-                            format: '%A, %B %d, %Y',
-                            locale: locale
-                        }));
+                        'aria-label', navigationFormatter.call(instance, columnDate));
 
-                    columnNode.html(formatter.call(instance, columnDate));
+                    columnNode.html(headerFormatter.call(instance, columnDate));
                 }
             );
         },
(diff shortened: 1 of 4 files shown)


third-party/projects/alloy-ui/src/aui-scheduler/js/aui-scheduler-view-day.js
```diff
@@ -903,13 +903,9 @@ var SchedulerDayView = A.Component.create({
                         CSS_SCHEDULER_TODAY_HD, !DateMath.isDayOverlap(columnDate, todayDate));
 
                     columnNode.setAttribute(
-                        'aria-label',
-                        A.DataType.Date.format(columnDate, {
-                            format: '%A, %B %d, %Y',
-                            locale: locale
-                        }));
+                        'aria-label', navigationFormatter.call(instance, columnDate));
```
igor-franca, line 906: I deployed the new webjar locally and walked through the calendar in day, week, and month views. Day view looks great with this change. I think there's an unintended side effect in week view though, and I wanted to write out what's happening in detail in case I'm missing something.

What you changed and why I think it makes sense in day view

Your commit replaces the hardcoded `'%A, %B %d, %Y'` with a call to the existing navigationDateFormatter attribute. That's a clean reuse Liferay.SchedulerDayView's navigationDateFormatter happens to return exactly `%A, %B %d, %Y`, so for day view your version produces the same string ("Wednesday, May 13, 2026") with one less hardcoded format. Nicer code.

What I'm seeing in week view

The issue is that `navigationDateFormatter` has a different contract per view. In calendar-web, each view overrides it to format the navigation bar label at the top of the calendar:

| View  | What the nav bar shows  | What navigationDateFormatter(date) returns       |
|-------|-------------------------|--------------------------------------------------|
| Day   | Wednesday, May 13, 2026 | the date itself, formatted as a full single date |
| Week  | May 10 — 16, 2026       | a range string of two dates joined by &mdash;    |
| Month | May 2026                | a month-year string                              |

Concretely, Liferay.SchedulerWeekView (in `modules/apps/calendar/calendar-web/.../scheduler.js`) defines it as:

```
navigationDateFormatter: {
    validator: isFunction,
    value(date) {
        const instance = this;
        const scheduler = instance.get('scheduler');
        const locale = scheduler.get('locale');
        const startDate = instance._firstDayOfWeek(date);
        const endDate = DateMath.add(startDate, DateMath.DAY, instance.get('days') - 1);
        // ...
        return [startDateLabel, '&mdash;', endDateLabel].join(' ');
    },
}
```

So when `syncDaysHeaderUI` loops over the seven day-header anchors and calls `navigationFormatter(columnDate)` per anchor, every call returns the same week-range string regardless of which columnDate is passed in. The rendered DOM ends up looking like:

```
<a aria-label="May 10 &mdash; 16, 2026">Sun 10</a>
<a aria-label="May 10 &mdash; 16, 2026">Mon 11</a>
<a aria-label="May 10 &mdash; 16, 2026">Tue 12</a>
...all seven identical...

```
Day view doesn't show either symptom because there's only one column anchor and the day view's `navigationDateFormatter` doesn't include any HTML entities, so both bugs are masked.

igor-franca (commented):
Hey Daniel 👋 thanks for taking this on and bumping yui3 too, really appreciate it.

PR #1264, opened 2026-05-13

dsanz:
refinement over https://github.com/liferay/liferay-frontend-projects/pull/1262 where we no longer reuse a function that does not work in all cases, instead, define a dedicated one.


igor-franca:
Hey @dsanz, the changes are aligned with our goal.

I ran tests on the calendar and everything seems to be working fine.

image

PR #1265, opened 2026-05-21
feat(js-api): Provides FDS atom utilities and state interface through js-api

antonio-ortega:
_This pull request is linked to [LPD-86378 - Technical Task | Create a Custom Element Client Extension for the Data Set Search Bar](https://liferay.atlassian.net/browse/LPD-86378). Including these changes and releasing a new js-api version is required in order to be able to merge [the pull request](https://github.com/liferay-frontend/liferay-portal/pull/5467) for that LPD, but both must be reviewed together._

# What’s New
 - Expose a new opinionated API offering some helpers to work with FDS Atom state. Callers could use `subscribeSearch` and `subscribeFilters` to interact with the atom. They just need to pass which `fdsName` they want to subscribe to, and once the subscription promise is resolved, they'll get an API (get, set, dispose) as a result to work with it.
 - Once the FDS Atom State grows (page, user views, order, ....) we can increase this API accordingly.
 - Since we want callers to only subscribe to that slice of the atom they will use, we hide `getFDSAtom`, because it is needed internally, but it should not be needed externally. We don't want them to manipulate the whole atom, just their slice.

# Dev Highlights
 - All the way until now this `js-api` have been used to provide some useful types for developers. Now, we had a bit of a different case because we want to offer some functionality rather than just some types, we are providing some API so external developers could get a slice of an atom just by the FDS Name and manipulate it.
 - In order to slice the atom, we are using selectors, so every API is offering to subscribe only to the specific selector.
 - `getFDSAtom` internal utility is in charge of polling for the atom and hiding the internal suffix `_fdsState` which does not need to be known by external callers.


antonio-ortega:
It follows here https://github.com/liferay/liferay-frontend-projects/pull/1267

PR #1266, opened 2026-06-05
feat(js-api): Add types for word-count ckeditor plugin

chestofwonders:
Issue: https://liferay.atlassian.net/browse/LPD-89734

<h1>Move @liferay/js-api out of js-toolkit and add editor/WordCount types</h1>

  `@liferay/js-api `was living inside `projects/js-toolkit/packages/`, which constrained it to` js-toolkit`'s tsconfig (moduleResolution:node, CommonJS) and liferay-workspace-scripts CI tooling. This made it impossible to consume CKEditor5 v46,
  which is ESM-only and requires `moduleResolution:bundler` and TypeScript 5.

  This PR:
  - Moves `@liferay/js-api` to `projects/js-api/` as an independent package, free of `js-toolkit`'s constraints
  - Configures it with its own tsconfig.json (TS5, `moduleResolution:bundler`) and `.eslintrc.js` (LGPL-3.0-or-later notice)
  - Adds `editor/index.ts` exporting `WordCount` from `@ckeditor/ckeditor5-word-count` and the `EditorTransformer/EditorConfigTransformer` interfaces for use in Client Extensions


projects/js-toolkit/packages/js-api/editor/index.ts
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: LGPL-3.0-or-later
  */
 
+/// <reference path="./ckeditor5-plugins" />
```
izaera: The lint errors seem correct: why not use `import` here? :thinking:

projects/js-toolkit/packages/js-api/editor/ckeditor5-plugins.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+declare module '@ckeditor/ckeditor5-word-count/dist/index.js' {
```
dsanz: at some point, we shall end up adding any ootb plugin type here, so that CX implementors can include them easily regardless the plugin they want to enable.

This is something we need to think: which types should be here? All? The ones which we don't include in the toolbars at this moment? Just the ones referenced in the sample workspace?

So, this is not blocking the PR, just raising awareness

dsanz (commented):
@chestofwonders pls check the linting errors

dsanz:
continues in https://github.com/liferay/liferay-frontend-projects/pull/1269

PR #1267, opened 2026-06-05
feat(js-api): Provides FDS state interface through js-api

antonio-ortega:
_This pull request is linked to [LPD-86378 - Technical Task | Create a Custom Element Client Extension for the Data Set Search Bar](https://liferay.atlassian.net/browse/LPD-86378). Including these changes and releasing a new js-api version is required in order to be able to merge [the pull request](https://github.com/liferay-frontend/liferay-portal/pull/5535) for that LPD, but both must be reviewed together._

# What’s New
 - Extends `liferay/js-api` package for interacting with Frontend Data Set state from components external to Liferay Portal. It adds a new `FDSConnection` interface which provides an opinionated way to read, write and subscribe to FDS State (for now, only to search slice, but it could be increased to other slices in the future, i.e, filters).
 - The key goal is to make easier for developers to work with FDS State management. Thanks to this new extension of `liferay/js-api` they can connect to FDS State without dealing with atoms, selectors or subscriptions.

# Dev Highlights
 - All the way until now this `js-api` have been used to provide some useful types for developers. Now, we are exposing a typed client surface in `data-set/client.ts` so a CX can redirects its import-map specifier to this module at build time with a tsconfig "paths" entry, so it imports FDSConnection and its companion types from a single specifier while the value is still resolved from the import map at runtime.


projects/js-toolkit/packages/js-api/data-set/FDSConnection.ts
```diff
@@ -4,13 +4,19 @@
...
 import {FDSState} from './index';
+import Atom = Liferay.State.Atom;
+
+const DEFAULT_TIMEOUT = 10000;
 
 export interface FDSStateChangeCallback {
 	search: (query: string) => void;
 }
 
+export interface FDSConnectionOptions {
```
dsanz: shall we reexport this interface in `intex.ts` so that consumers can import it?

projects/js-toolkit/packages/js-api/liferay.d.ts
```diff
@@ -4,6 +4,15 @@
  */
 
 declare namespace Liferay {
+	interface EventHandler {
+		detach: () => number;
+	}
+
+	function on(
```
dsanz: This signals we are fine exporting a `Liferay.on` function to CX implementors. Indeed, this has been the case for years. We must remember this when we get rid of AUI as the event registration API will need to remain. Of course, implementation will be different

dsanz (commented):
LGTM ;) please check inline comments, as we might need a small extra commit. 

Also, let's wait for other people's review.

That's why this is just commenting for the moment

juanjofgliferay:
You nail it. Having a simple public interface like this is for devs and managing the subscription/unsubscription under the hood is a really good idea.

projects/js-toolkit/packages/js-api/data-set/_internal.ts
```diff
@@ -0,0 +1,72 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+/**
+ * Package-private helpers shared by the opinionated subscription
+ * entry points (`subscribeSearch`, `subscribeFilters`). These are not
```
juanjofgliferay: This comment seems to be outdated as there is no `subscribeSearch` nor `subscribeFilters` around 🤔

projects/js-toolkit/packages/js-api/liferay.d.ts
```diff
@@ -0,0 +1,52 @@
...
+		events: string | string[],
+		callback?: (event?: any) => void
+	): EventHandler;
+
+	namespace State {
```
dsanz: Main goodness of our new solution is that this type is no longer needed here. We reduce State API exposure to  0.

Of course, third party code can still do `Liferay.State` but won't receive any type assistance from this public package.

projects/js-toolkit/packages/js-api/liferay.d.ts
dsanz: Now we don't need this either, do we?

projects/js-toolkit/packages/js-api/tsconfig.json
```diff
@@ -2,10 +2,11 @@
...
 	"compilerOptions": {
 		"composite": true,
+		"jsx": "react",
 		"lib": ["DOM", "ES2020"],
 		"outDir": ".",
 		"rootDir": ".",
 		"target": "ES2020"
 	},
-	"include": ["**/*.ts"]
+	"include": ["**/*.ts", "**/*.tsx"]
```
dsanz: Do we need any `*.tsx` here?

dsanz (commented):
@antonio-ortega pls check comments. Also ,a minor nit, the name in the PR needs to be updated to reflect our last decisions.
Thanks a lot!

izaera (commented):
Why do we need a `client.ts` file? 
Wouldn't the types in `index.ts` suffice?
Won't we have two paths to import the same types (`/` and `/client`) ?
🤔

dsanz (approved):
LGTM ;)

PR #1269, opened 2026-06-16
feat(js-api): Add ckeditor5 types and isolate js-api project from the monorepo yarn workspace

dsanz:
Continuation of https://github.com/liferay/liferay-frontend-projects/pull/1266  where I'm fully detaching the yarn workspace for js-api. Only commonality is the build scripts


js-api/package.json
```diff
@@ -4,12 +4,16 @@
...
 		"type": "git",
 		"url": "https://github.com/liferay/liferay-frontend-projects.git"
 	},
+	"dependencies": {
+		"@ckeditor/ckeditor5-word-count": "*",
+		"typescript": "^5.0.0"
+	},
 	"scripts": {
-		"build": "tsc",
+		"build": "./node_modules/.bin/tsc",
```
izaera: The node_modules reference is not needed (and may fail in Windows)

client-extensions/js-api/.gitignore
```diff
@@ -1,5 +1,7 @@
 *.d.ts
 *.js
+!.eslintrc.js
+!copyright.js
```
izaera, line 5: why do we want to gitignore these files? 🤔  They are commited, right?

izaera (changes requested):
Other than the issues I've signaled it LGTM

js-api/package.json
dsanz: nice one!

client-extensions/js-api/.gitignore
dsanz: they are exceptions from the `*.js` ignores, so they shall be allowed

dsanz:
@izaera please take a look at the new proposal, thank you

client-extensions/js-api/.yarnrc
```diff
@@ -0,0 +1,3 @@
+# Read by liferay-changelog-generator to derive the release tag prefix
```
izaera, line 1: this file is not needed if we are not using yarn

client-extensions/js-api/.npmrc
```diff
@@ -0,0 +1,2 @@
+tag-version-prefix=js-api/v
```
izaera, line 1: not sure if this works like in yarn...

client-extensions/js-api/.npmrc
```diff
@@ -0,0 +1,2 @@
...
+message=chore: prepare @liferay/js-api v%s release
```
izaera, line 2: not sure if npm publish commits anything to git

client-extensions/js-api/package.json
```diff
@@ -0,0 +1,34 @@
...
+	},
+	"devDependencies": {
+		"@liferay/eslint-plugin": "1.9.1",
+		"eslint": "7.23.0",
+		"prettier": "3.2.5",
+		"typescript": "^5.0.0"
+	},
+	"scripts": {
+		"build": "tsc",
+		"changelog": "npx liferay-changelog-generator",
```
izaera, line 34: I'm not sure if we want a changelog... I mean, it's a good thing but it has a maintenance cost and I doubt anyone is using it.

Also, for a project like this where we plan not to have breaking changes and nothing is executable we may get rid of it, probably.

izaera (changes requested):
I've added some more comments. Some of them are simply unknowns to me. I guess trying to publish a version of this with npm will clear them.

dsanz:
I'm merging this as a first attempt to publish the js-api package in its new shape

PR #1275, opened 2026-09-14
feat(js-api): let a Client Extension take over the filtering of a data set

antonio-ortega:
A Client Extension could read and write the search state of a data set through an FDS connection, and nothing else. Now, it filters too: `setFilters()` takes plain OData expressions, which travel through the `connectionFilters` member of the data set state, and `clearFilters()` leaves nothing applied without giving the filtering back. The filters the data set declares stop reaching the request for as long as a connection owns the filtering, so the Client Extension owns the whole expression, and those filters stay out of that state on purpose, since the data set owns them and a connection has no business writing them.

**What a connection takes over is declared when it opens,** through the new `owns` option. That is what lets the data set drop its filters dropdown and its filter chips as part of connecting instead of flashing them into view and back out, and it makes the search-only case expressible: a connection that leaves filters out drives the search alone and the data set keeps filtering as it always has.

Filtering a data set is part of its address, and taking the filtering over used to cost it that, since the expressions a Client Extension applied lived only in the memory of the page. `setFilters()` now takes, alongside them, one value of the Client Extension's choosing. The data set keeps it in the URL for as long as those expressions reach the request and reads nothing in it, and hands it back through the new restore state change callback on the next visit, and again whenever the back or forward button lands on a different address. The expressions themselves never reach the URL, so a link can only ever ask for a filter the Client Extension would have built itself, and the data set cannot check the value either, so validating what comes back from a URL anyone can edit is the Client Extension's obligation. It has to be put back from inside the callback rather than from an effect it schedules, because a data set loading an address that carries a value holds its first request until the connection has handed it over.

A data set has one filtering owner and only one can get it, so that two Client Extensions on one page cannot both filter it with whichever called `setFilters()` last deciding what it shows. The rest are refused and keep the connection they would have had with `owns: ['search']`, settling at the new refused status rather than ready, so a Client Extension that enables its controls once ready leaves them alone without writing any code for the case. Which app owns the filtering is said in the state of the data set, as `filteringOwnerAppId`, since the data set is the side that keeps what a connection remembers in the URL and what it keeps is filed under an app id. That is also why owning the filtering requires an `appId`. One owner does not need the key today. It is there so that the day the filtering can be split between several Client Extensions, a link saved now still reads as a link naming one of them rather than as a format that has to be guessed at, and nothing has to be migrated.

A Client Extension is a custom element and may be taken off the page without disconnecting, which would leave the filtering owned by a connection nothing on the page can reach until the next reload. A connection may now name the element it is drawn in, and a claim whose element has left the document is handed back the way disconnecting would have handed it back.


dsanz (approved):
LGTM ;)

